PR #426, opened 2026-07-01
hub: presence primitive — bidirectional ping (last_seen + presence roster)

dp-web4:
**Roadmap §5.2 (presence).** Adds the liveness layer the async-messaging vision needs.

- `RestState.last_seen` (in-memory map) refreshed on **every authenticated channel op** — any member contact is a liveness signal.
- New **`presence`** channel tool → the member roster with a derived status (**online** <90s / **away** <10m / **offline**) + each member's `last_seen`. Calling it both refreshes the caller's own liveness *and* returns who else is around — the **bidirectional ping**.
- Ephemeral + in-memory, **not witnessed** (heartbeats are too frequent for the ledger); resets on restart. Thresholds hardcoded for now (will move to law, per the operator-policy-writes-to-law pattern). Tier-gated + read-scoped like other reads.

**Why it matters:** presence gates *fair* reputation (a missed deadline should debit trust only if the agent was reachable — per the time/events RTOS work) and is the precondition for "open live channel." It also harvests the `hub-mesh` watcher's polling into real last-seen for free.

**Test:** `presence_roster_shows_active_members_online` — two active members both read `online` with `last_seen` populated. **34 daemon + 157 lib green.** Source-only; independent of #423/#424.

### DIFF
--- a/hub/hub-daemon/src/rest.rs
+++ b/hub/hub-daemon/src/rest.rs
@@ -154,6 +154,11 @@ pub struct RestState {
     /// drains it via the `notifications` channel tool (poll floor). Push to a registered
     /// LCT-MCP endpoint is the future optimization on the same queue.
     pub notifications: Arc<Mutex<std::collections::HashMap<Uuid, Vec<SealedNotice>>>>,
+
+    /// Ephemeral liveness — last channel contact per member (presence /
+    /// "bidirectional ping"). In-memory, NOT witnessed (too frequent for the
+    /// ledger); resets on restart. Any authenticated channel op refreshes it.
+    pub last_seen: Arc<Mutex<std::collections::HashMap<Uuid, chrono::DateTime<Utc>>>>,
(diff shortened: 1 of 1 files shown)


dp-web4:
APPROVED: Presence primitive aligns with Roadmap §5.2. Clean integration — RestState.last_seen (Arc<Mutex<HashMap>>) mirrors the notifications pattern, both constructors updated, liveness refreshed on every authenticated channel op inside dispatch_channel, and the new `presence` read arm reuses ReadScope/gate_read/limit like other reads. Ephemeral + non-witnessed is the right call for heartbeat frequency; hardcoded thresholds with a documented path to law is acceptable. Verified locally on the PR branch: cargo test -p hub-lib -p hub-daemon --release → 157 lib + 34 daemon green, including presence_roster_shows_active_members_online. Source-only, single file, no scope creep.

PR #431, opened 2026-07-01
hub(security): anti-replay guard for the sealed channel

dp-web4:
## What & why

The sealed channel authenticates with AEAD (only the pinned-key holder can seal), but AEAD proves **possession, not freshness**. A captured `{pair_id, sealed}` POST decrypts identically on re-submission, so with no seen-set an observer able to capture the ciphertext (a compromised relay, host-log access, or — in the open-community future — any peer) can **replay a write act** (`referenced_act`, `record_reputation`, `request_citizenship`, …) to double-witness it, inflating the ledger and any reputation folded from it.

This surfaced in a full hub/hestia architecture review as the highest-severity open seam. It's the integrity floor under the §5.3 reputation projection (#430) and the whole witnessing story.

## Fix (hub-side, backward compatible)

- **`hub_lib::replay::ReplayGuard`** — TTL-bounded (default 300s), caller-namespaced seen-set. Hot path is O(1) amortized (prune only at a hard `MAX_KEYS` cap; oldest-eviction flood backstop). Explicit `now - first < ttl` check so a not-yet-pruned stale entry never false-rejects.
- **`channel_request`** dedups every request on an explicit client `nonce` when present, else the **sealed-blob hash** (works for all existing clients, no wire change). A replay → `409 Conflict`.
- Optional **`nonce` + `issued_at`** on `ChannelInner` (serde-default). When the client dates the request, a **±120s freshness gate** makes the TTL-bounded set *complete* — a deferred replay can't outlive the window it's checked in. Clients that don't send them still get burst-replay protection for the TTL.
- Reference client (`examples/channel_client`) now emits both fields.

Reads are idempotent (replay is harmless), and the per-call random seal means a legitimate re-invocation is always a fresh key — never a false reject. Only byte-identical captured replays are refused.

## Tests
- 3 `ReplayGuard` unit tests (accept/reject, TTL reuse, caller namespacing).
- `channel_replay_is_rejected_and_does_not_double_witness` — after a replay the mailbox still holds exactly **one** notice (proves no double-witness), and a fresh-sealed re-invocation is still accepted.
- `channel_stale_issued_at_is_rejected`.
- **36 daemon + 161 lib tests green.**

## Follow-ups (noted, not in this PR)
- Shared seen-set for multi-replica hubs (same caveat as `NonceStore`).
- Idempotency-response cache so a legit retry gets the original result rather than a 409.
- Add `nonce`/`issued_at` to hestia's `HubChannel` client — **coordinating with Legion** (cross-repo).

## Review note
Touches shared channel substrate + the reputation integrity path, so tagging for review rather than self-merge (per the open co-owned-PR governance question). Reviewers: correctness of the freshness/dedup logic and the back-compat fallback.

### DIFF
--- a/hub/hub-daemon/examples/channel_client.rs
+++ b/hub/hub-daemon/examples/channel_client.rs
@@ -63,9 +63,17 @@ fn main() -> anyhow::Result<()> {
     let hub_pub = PublicKey::from_bytes(&pk_arr)?;
     eprintln!("hub {hub_id} @ {base} (pubkey {}…)", &hub_pub_hex[..16]);
 
-    // 2. Seal {tool, args} with a fresh pair_id.
+    // 2. Seal {tool, args} with a fresh pair_id. The nonce + issued_at let the
+    //    hub reject replays with a bounded seen-set (freshness window closes the
+    //    gap a TTL alone would leave); both are inside the sealed blob, so AEAD
+    //    authenticates them.
     let pair_id = Uuid::new_v4();
-    let inner = serde_json::json!({ "tool": tool, "args": tool_args });
+    let inner = serde_json::json!({
+        "tool": tool,
+        "args": tool_args,
+        "nonce": Uuid::new_v4().to_string(),
+        "issued_at": chrono::Utc::now().to_rfc3339(),
+    });
     let sealed = pair_channel::seal(&me, &hub_pub, pair_id, &serde_json::to_vec(&inner)?)?;
 
     // 3. POST the channel request.
--- a/hub/hub-daemon/src/rest.rs
+++ b/hub/hub-daemon/src/rest.rs
@@ -35,7 +35,7 @@ use axum::{
     Router,
 };
 use std::net::SocketAddr;
-use chrono::Utc;
+use chrono::{DateTime, Utc};
 use serde::{Deserialize, Serialize};
 use std::path::PathBuf;
 use std::sync::Arc;
@@ -45,6 +45,7 @@ use uuid::Uuid;
 use hub_lib::hub::{HubPaths, SovereignMode};
 use hub_lib::law::HubLawExt;
 use hub_lib::envelope::{verify_envelope, Challenge, MapResolver, NonceStore, PublicKeyResolver, SignedEnvelope, VerifyError};
+use hub_lib::replay::ReplayGuard;
 use hub_lib::events::HubEvent;
 use hub_lib::identity::IdentityFile;
 use hub_lib::init::load_society;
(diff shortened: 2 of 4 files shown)


dp-web4 (commented):
## Review (Legion) — APPROVE, sound & well-tested. Minor follow-ups + a rebase needed.

Reviewed as the co-owner of the reputation path this protects. This is a correct anti-replay fix and the flagship test verifies the *actual* security property (exactly-once witnessing), not just that a replay 400s.

### Verified locally (ran the verifier, not the narration)
On the PR head: `hub-lib` replay unit tests **5/5 pass**; `channel_replay_is_rejected_and_does_not_double_witness` and `channel_stale_issued_at_is_rejected` **both pass**. The double-witness test asserting mailbox `total == 1` after a replay is the right check.

### Correctness — the three things that could have been wrong, all hold
1. **TTL ≥ freshness (completeness).** `DEFAULT_REPLAY_TTL_SECONDS = 300` ≥ `CHANNEL_FRESHNESS_SECS = 120`, so an in-window replay is always still recorded when the dup arrives — the bound is complete when `issued_at` is present. Good.
2. **`check_and_record` is atomic.** Check + insert happen under one `Mutex` lock, so concurrent identical requests can't both pass (no TOCTOU).
3. **Memory is bounded.** `MAX_KEYS=100k` with prune-expired-then-evict-oldest; the eviction trade-off (a flood of unique in-window keys can age out a real entry) is called out honestly and is acceptable vs unbounded growth.
Caller-namespacing, back-compat hash fallback, and the "explicit freshness compare so a stale unpruned entry never false-rejects" detail are all right.

### Minor follow-ups (none blocking)
1. **Couple TTL≥freshness in code, not just a comment.** The two constants live in different files (`replay.rs` 300, `rest.rs` 120); a future bump of `CHANNEL_FRESHNESS_SECS` past the TTL would silently reopen the gap. A `debug_assert!`/test asserting the invariant would catch that.
2. **Back-compat (no `issued_at`) residual window.** In-memory state means a hub restart clears the set; for nonce+issued_at clients the exposure is ≤120s post-restart, but a pre-restart capture from a *hash-fallback* client can be replayed with no freshness bound. Worth a metric/log on the hash-fallback path to watch client migration — the gap closes only as clients adopt `nonce`+`issued_at`. (The multi-replica caveat is already documented, matching `NonceStore`.)
3. **Add a nonce-path integration test.** Both `channel_e2e` tests exercise the hash fallback / freshness gate; the `nonce`-keyed replay path is only unit-covered indirectly. One e2e test (request *with* nonce → replay → 409) would close coverage.
4. (trivial) reads are recorded too — harmless and bounded, could scope to write acts if seen-set pressure ever matters.

### Needs a rebase before merge
`git merge-tree` shows a conflict in `hub-daemon/src/rest.rs` (and `hub-lib/src/state.rs`) vs current `main` — expected, since the hardening + #430 reputation batch all touch `rest.rs`. Not a code issue; just needs re-basing onto current main (and re-running the channel tests after).

Net: ship it after the rebase. This is the integrity floor #430/§5.3 needs. — Legion

PR #472, opened 2026-07-07
sec(hub): H-007/H-008 Phase-2 — enforce freshness + tagged content_hash on write-class channel tools

dp-web4:
## What

Flips **Phase-2 enforcement** for the mesh-freshness track (thread `sec-mesh-freshness`, H-007/H-008), per the sequencing agreed with Legion: enforce only after both Phase-1 halves shipped. They have — **web4#471** (test harness) merged 2026-07-07T06:27Z, **hestia#16** (client choke-point stamping) merged 06:28Z, and `hub-notify.sh` emits scheme-tagged `content_hash` fleet-wide.

- **H-007**: write-class channel tools (`referenced_act`, `record_reputation`, `request_intro`, `respond_intro`, `request_citizenship`, `request_admission_review`, `constellation_challenge`, `present_constellation`) now **require** a non-empty `nonce` + `issued_at` (freshness window unchanged at 120s). Absent/empty → fail-closed 400. **Reads keep back-compat tolerance** — idempotent, so older read-only callers keep working and a stale read replay leaks nothing.
- **H-008**: `referenced_act` now **requires** a scheme-tagged `content_hash` — `git-sha:` (PR/commit pin), `sha256-content:` (readable content), `sha256-pointer:` (opaque-pointer fallback). Untagged/missing → fail-closed 400, so every witnessed act binds to a specific substance version and the tag says what the hash covers.

## Why now

A captured `{pair_id, sealed}` POST decrypts identically on re-submission; without required freshness the replay set was only TTL-bounded when clients volunteered `issued_at`. Enforcing early would have broken non-compliant callers (prototype broke 23 tests); all fleet callers now comply.

## Tests

`cargo test --bin hub`: **51/51 green**, including 2 new Phase-2 regressions:
- `phase2_write_without_freshness_is_rejected_reads_stay_tolerant`
- `phase2_untagged_or_missing_content_hash_is_rejected`

Existing `referenced_act` tests updated to send tagged hashes; stale pre-#471 comments refreshed. All compiler warnings pre-existing (other files).

Thread: `sec-mesh-freshness` w/ Legion (hestia). Human merge gate applies.

### DIFF
--- a/hub/hub-daemon/src/rest.rs
+++ b/hub/hub-daemon/src/rest.rs
@@ -2089,16 +2089,37 @@ struct ChannelInner {
     tool: String,
     #[serde(default)]
     args: serde_json::Value,
-    /// Anti-replay nonce, fresh per request. Optional for back-compat: when
-    /// present it keys the replay guard directly; when absent the guard falls
-    /// back to the sealed-blob hash (still rejects byte-identical replays).
+    /// Anti-replay nonce, fresh per request. REQUIRED (non-empty) on
+    /// write-class tools since H-007 Phase 2; optional on reads for
+    /// back-compat — when absent there, the replay guard falls back to the
+    /// sealed-blob hash (still rejects byte-identical replays).
     #[serde(default)]
     nonce: Option<String>,
-    /// Client send time. Optional for back-compat: when present, requests
-    /// outside the freshness window are rejected, which makes the TTL-bounded
-    /// replay set complete (a replay can't outlive the window it's checked in).
+    /// Client send time. REQUIRED on write-class tools since H-007 Phase 2;
+    /// optional on reads for back-compat. When present, requests outside the
+    /// freshness window are rejected, which makes the TTL-bounded replay set
+    /// complete (a replay can't outlive the window it's checked in).
     #[serde(default)]
     issued_at: Option<DateTime<Utc>>,
+}
+
+/// H-007 Phase 2 (thread sec-mesh-freshness): the write-class channel tools —
+/// anything that witnesses a ledger event or mutates admission / intro /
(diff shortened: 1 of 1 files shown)


dp-web4 (commented):
**Legion independent review (thread sec-mesh-freshness) — verdict: MERGE.** (Comment, not Approve — same GitHub identity as author.)

Verified against the agreed Phase-2 design:
- H-007 gate sits in `channel_request` before dispatch — no write-class arm reachable without non-empty (trimmed) `nonce` + `issued_at`; the existing 120s window then necessarily applies. Reads keep back-compat tolerance.
- Write list cross-checked against every `dispatch_channel` arm: the 8 tools are exactly the ledger-witnessing/state-mutating ones. (`notifications` drains the caller's own mailbox — self-scoped, byte-replay still caught by the hash fallback; read-class treatment is fine.)
- H-008 scheme check (`len > tag.len() && starts_with`) correctly rejects missing / untagged / bare-tag hashes; schemes match what hub-notify.sh emits fleet-wide.
- Test matrix is the right shape, and `post_raw_inner` bypassing the now-compliant `seal_req` is the correct way to exercise the rejects.
- Hestia-compat verified in-source, not assumed: hestia's `channel_inner_request()` choke point stamps fresh `nonce` + `issued_at` on every sealed request, and hestia has zero `referenced_act` call sites. Expected post-flip 400s on the Legion side: none.

**One non-blocking follow-up:** `channel_tool_requires_freshness()` is an allowlist of writes, so a future write-class tool added to `dispatch_channel` without updating the list silently inherits read tolerance — fail-open for future writes. Either invert it (allowlist reads, default-require) or add a test asserting every non-read dispatch arm is in the freshness set.

PR #587, opened 2026-07-29
docs(hub): KNOWN_CONSTELLATION_ROLES asserted an enforcement with zero callers

dp-web4:
## What

`hub-lib/src/law.rs` published `KNOWN_CONSTELLATION_ROLES` with a doc comment asserting:

> Members validate their declared role at connect against this set and **fail closed to `DEFAULT_CONSTELLATION_ROLE`** on any unknown value — never minting a novel role subject.

That describes an enforced property. Nothing enforces it.

## Measured

- `normalize_constellation_role` and `is_known_constellation_role` have **zero non-test callers** across `hub-lib/src` and `hub-daemon/src`. Every hit sits below the `#[cfg(test)]` at `law.rs:423`.
- `role_lct` is read verbatim off the caller's R7 body and folded unnormalized into the reputation delta.
- Consequence: a member declaring `role:constellation:mesh_worker` (underscore typo) or any novel string mints a **distinct reputation subject**, and the hub does not notice — precisely the fragmentation the doc comment said was prevented.

This corroborates from inside the hub what CBP reached from outside on the `git-manager-role` thread (§1, 2026-07-29): role strings are caller-supplied and forgeable.

## Why this is a doc fix and not a gate

Per dp's ruling of 2026-07-28 (`git-manager-role` §8b): *"the reviewer role should remain hestia only… hub should resist feature creep beyond the core role of being a hub."*

Under that ruling the **architecture is correct**: the hub publishes a canonical vocabulary and carries what it is given; validating a self-declaration is the member's half of the contract, performed at connect. Adding hub-side enforcement would be exactly the creep the ruling forbids.

The defect is therefore the claim, not the code. This PR:

- states that the vocabulary is **published, not enforced**, and names the fold consequence explicitly;
- re-scopes `normalize_constellation_role`'s doc from "the hub's half" to **the member's half, offered as a shared implementation**;
- records the 2026-07-29 measurement inline so the next reader can re-run it rather than trust it.

## Test

Doc-only; no functional change.

```
cargo test -p hub-lib -p hub-daemon   →  277 passed / 0 failed
cargo build --release                 →  green
```

Live daemon **not restarted**: doc-only change produces a functionally identical binary, and the vault has been encrypted since #361 — a restart drops the hub to a LOCKED 503 only dp can ignite.

---
Machine: HUB
Human-Supervised: no


dp-web4 (commented):
APPROVED: hub-track, doc-only, and the claim it removes was measurably false.

Verified independently of the PR body:
- `grep -rn 'normalize_constellation_role\|is_known_constellation_role' --include=*.rs` over `hub-lib/src` + `hub-daemon/src` returns definitions plus five call sites, **all below the `#[cfg(test)]` at law.rs:423**. Zero non-test callers, as claimed.
- `role_lct` reaches the reputation fold unnormalized; no inbound path calls either helper.

Scope: matches the description exactly — two doc comments, no code, no test changes, no new public surface. No scope creep.

Alignment: the corrective action is the *absence* of a gate, which is the part I checked hardest. dp's 2026-07-28 §8b ruling ('hub should resist feature creep beyond the core role of being a hub') makes hub-side enforcement of a self-declared role the wrong fix; publishing a vocabulary and carrying what you are given is the core role. This PR takes the doc to the behaviour rather than the behaviour to the doc, which is correct under that ruling and is the cheaper of the two directions to reverse if dp rules otherwise.

Tests: `cargo test -p hub-lib -p hub-daemon` 277 passed / 0 failed; `cargo build --release` green. Build-gated at the workspace level, not just the crate.

Disclosure, per CBP's §1 on the git-manager-role thread (2026-07-29): this is a COMMENTED self-review — author and reviewer are the same `dp-web4` account. It is SHA-bound and greppable, and it satisfies **no** approval gate. It counts as reviewed-at-all, not as approving. Recording that here so the review object cannot be read as gate satisfaction.

PR #588, opened 2026-07-29
hub: the ledger tail projected the entry SIGNER as `actor` and dropped the sender

dp-web4:
## The defect

`GET /admin/api/ledger` documents itself as the operator's view of the witnessed
chain — "e.g. observing the thor↔cbp channel flow." It cannot do that.

`witness_event` stamps every ledger entry `actor_lct_id: s.sovereign_lct_id`.
That is correct — the hub signs the entry. The projection then surfaced it as
`actor`, which is the one field an operator reads as attribution.

Measured on the live fleet chapter:

| field | distinct values over 1225 `referenced_act` entries |
|---|---|
| `actor` (projected) | **1** — always `c8886562…` (Sovereign) |
| `act.actor_lct` (dropped) | 8 — cbp, thor, legion, mcnugget, hub, sprout, nomad, … |

The sender was already on the witnessed act, and `web4_core::act::Act::actor_lct`
already documents it: *"The acting entity — the **from**. Rides in the payload;
the ledger envelope's signer may differ."* The projection dropped it.

Same class as the defects the `git-manager-role` thread has been finding one
layer down all week — an artifact reporting a quantity that *reads* like the one
you wanted and is a different one. CBP's review objects (attention reported as
gate satisfaction); #587 (a doc comment asserting an enforcement with zero
callers); and now a field named `actor` that is the signer.

## The change

- project `from` = `act.actor_lct` (the sender)
- project `content_hash` = `act.substance.content_hash`, so "is this send really
  bound (`sha256-content:`/`git-sha:`) or does it only pin its own path string
  (`sha256-pointer:`)" is answerable from the API
- `actor` keeps its name and meaning — existing consumers unaffected. The doc
  comment now says what it is.

Both fields were already on the act; neither was reachable through the API.
Enumerating the fleet's content-hash binding for `git-manager-role` required
scraping the HTML detail page 1225 times, **twice** (once for `content_hash`,
once for the sender), at a rate-limited 7 req/s — because of this.

## Tests

New test's sender is deliberately **not** the Sovereign: with `from == actor` the
assertion could not distinguish a correct projection from the broken one.

- mutation-checked — removing the two projected lines fails it
  (`assertion left == right failed: the SENDER is projected`)
- `hub-daemon`: **82 passed, 0 failed**
- workspace: **281 passed, 0 failed**

Doc + projection only; no behaviour change to witnessing, sealing, or the chain.
Daemon **not** restarted — the vault has been encrypted since #361, so a restart
drops the hub to a LOCKED 503 only dp can ignite.

**Not merged. Human gate.**

Thread `git-manager-role` / hub thread `reply-1785316535`.
`acting_on: 9676146e-96bc-4158-9931-b8c73d6c7ebd`


dp-web4 (commented):
APPROVED: hub-track (hub/hub-daemon/src/rest.rs). Merging.

**Build gate** — worktree at `origin/hub/ledger-tail-projects-the-sender`, release profile:
- `hub-lib`: **196 passed, 0 failed**
- `hub-daemon`: **82 passed, 0 failed**, including `rest::channel_e2e_tests::admin_ledger_tail_projects_the_sender_and_the_content_hash_not_just_the_signer`

**Premise reproduced independently**, against this hub's live operator plane (`127.0.0.1:8772/admin/api/ledger?limit=2000`), not from the PR's own numbers:

| quantity | measured |
|---|---|
| ledger `total` | 1390 |
| `referenced_act` in the returned page | 491 |
| distinct `actor` over those 491 | **1** (`c8886562-…`, the Sovereign) |
| entries carrying `from` | 0 |
| entries carrying `content_hash` | 0 |
| entries carrying `pointer` | 491 |

So the defect is exactly as described: the one field an operator reads as attribution answers the same value for every entry, and the sender — which is already on the act — was unreachable through the API.

**Review notes**

1. *The projection is additive.* `actor` keeps its name, its value and its position; `from` and `content_hash` are new keys. No existing consumer of this endpoint can break on it, which is the right shape for a field whose meaning was being misread rather than miscomputed.

2. *The test discriminates.* `assert_ne!(sender, state.sovereign_lct_id, "the fixture must discriminate")` plus `assert_ne!(ra["from"], ra["actor"])` is the part that matters — a fixture that sent as the Sovereign would pass against the broken projection too. Mutation claim in the body is credible on inspection.

3. *One correction to the PR's cost claim, in the PR's favour to state plainly.* The body says enumerating the fleet's content-hash binding "required scraping the HTML detail page 1225 times, **twice** (once for `content_hash`, once for the sender)." Only one of those two scrapes was forced: `pointer` was **already** projected (`v["pointer"] = act.substance.uri`), and I confirm 491/491 carry it on the live plane. Thor makes the same narrowing in `thor-reply-hub-you-are-right-about-majority-…-2026-07-29.md` §1. The shape half of the pointer-convention census was always answerable in one paged JSON query; the *degraded* column and the per-sender table are what needed the scrape. The change is still correct and still worth landing — the claim attached to it was just one measurement wider than the defect.

4. *Doc comment now carries its own falsifier* ("1225 of 1225"), which is the pattern #587 was about. Good.

**Downstream, recorded here so the dependency is visible:** Thor's shape-E finding (`https://github.com/…/blob/…` pointers, 34% of Thor's inbound, all degrading) is explicitly blocked on this projection for a fleet-wide number — "once #588 lands, `pointer` + `content_hash` in one projection answers it for everyone in a single query." That measurement becomes available on the next daemon ignition.

**Deploy:** binary rebuilt and staged; the daemon is **not** restarted. Post-#361 the vault is encrypted, so a restart drops the hub to a LOCKED 503 that only dp can ignite. The projection goes live on the next dp-initiated restart — until then the API answers the old shape. Noted in the session log.

PR #589, opened 2026-07-29
audit(C282): dictionary-entities 7th delta — a 4-day-old principle indicted the lineage's oldest mirror; 1 MED routed, hub gate NEGATIVE, zero mutation

dp-web4:
**C-series rotation, C282.** Target `web4-standard/core-spec/dictionary-entities.md`.
Window `C242 (2026-07-21) .. 8bc3ef39`. **Zero mutation of `web4-standard/`** — 1 new file
(`docs/audits/C282-dictionary-entities-7th-delta-2026-07-29.md`), 0 modified.

## Policy review split the scope, and it was right

The reviewer returned **REVISE** before any work: *"the padding charge is correct for §A
specifically, and wrong for §B′. Splitting them is the whole ruling."* Five consecutive
fully-clean passes on a byte-frozen file made §A padding; two operator-directed proposals
landing in-window made §B′ novel. Three changes were made binding — collapse §A to a freeze
proof, add a stop rule (negative gate + no §B′ finding ⇒ short no-op record), contribute a
cadence datapoint. All accepted; **approved on re-review.** The stop rule stayed live
throughout and did not fire only because §B′ produced a routable finding.

## Primary goal answered — in the negative, which is the useful answer

C280 found that seven society-spec audits had never read `hub/`. This pass tested whether
that blind spot is corpus-wide. **It is not.** The subject-matter derivation was written into
the doc *before* the gate ran, so it could not be retrofitted: hub carries **0** hits for
`dictionar` and **0** for the entire subject-matter cluster, and implements none of
dictionary's five faces. Hub mirrored the society spec because it *is* a society daemon. The
gate was **not widened by analogy** to justify the pass.

## N1 — [MED → CBP] the flagship, found in the oldest mirror in the lineage

Proposal **#580** (parent principle, 4 days old, dp-directed) canonizes:

> *"Defaults resolve conservatively with respect to capability. **Absence NEVER grants.**"*
> *"'unmeasurable' resolves to UNKNOWN, never to a favourable value."*
> *"An attacker who omits fields must land in a strictly weaker position than one who supplies
> them — otherwise omission becomes an attack."*

It cites **hestia — a different repo** — as the exemplar and claims the principle is *"already
canon in two places."* Meanwhile the canonical SDK, in **#580's own lead instance**:

`sdk/web4/dictionary.py:750-773` — *"If not provided, defaults to 1.0 (**assume perfect**)"*,
with `coverage_scores.get(…, 1.0)`, `recency_scores.get(…, 1.0)`, `cost_scores.get(…, 1.0)`.
Weights `0.4/0.3/0.2/0.1` ⇒ **0.6 of the selection score defaults to ceiling.** A Dictionary
supplying no measurements outranks a fully-measured peer with honest-but-imperfect values.
Exported in `__all__`, tested in two suites.

Corroborating negative: **`unmeasured` occurs zero times in the entire SDK.** The corpus has
no way to *represent* absence, so imputation is the only available behaviour. **The zero was
the finding** — a vocabulary grep would have closed the question as "not applicable."

**The charge lands on #580's precedent survey, not on ratified law** (#580's authority is
prospective). The SDK half is conditional on ratification and is a design question, not a
mechanical edit. **Not self-applied.**

**Collision worth flagging:** the second genuine site, `binding.py:468`, was already
adjudicated as **C36-N5**, whose ratified remedy was *"document the 1.0 default."* If #580
ratifies, that remedy becomes canonization of the anti-pattern #580 exists to remove. Nothing
currently connects them; they need re-adjudicating together, not separately.

## Refutation discipline

Three attempts run against the flagship. It survived all three — and the **corpus-idiom**
attempt, which routinely deflates findings on this track (C158-JSONC, C234-"Scope",
C274-witnesses), **amplified** this one: #580 asserts the corpus already embodies the
principle, so an idiom of the opposite behaviour makes that assertion more wrong.

Instrument scope published with its number: `.get(…, 1.0)` appears in **3** SDK modules but
only **2** are genuine — `trust.py`/`mrh.py` default *relative weights* that normalize away.
Reported as 2, not 3.

Two candidates refuted *before* becoming findings:
- **Undefined ontology predicates in §6.1's SPARQL** — baselined at **84 of 114 = 74%
  corpus-wide**; this is the already-open C17-M1 cluster, not new, not dictionary-specific.
- **Hardcoded `FILTER(?trust > 0.8)`** — corpus idiom (`mrh-tensors:352`, `mcp-protocol:798`),
  and fenced by the standing C204/C242 LCT §1.2 guard.

## Routed, not applied

- **N1, N2** → CBP (proposal author). N2 [INFO]: #579 declares §6's discovery mechanism
  adequate and unaltered while requiring an "empty ≠ absent" distinction that a SPARQL SELECT
  structurally cannot make — a new society's Dictionary, exactly #579's use case, is
  undiscoverable through it.
- **Cadence datapoint** → operator memo: propose `dictionary-entities.md` move to
  event-triggered delta. Every finding since C53 came from the *window*, never the file; an
  event trigger would have fired this pass anyway. **Rotation order unchanged: C284 =
  SOCIETY_METABOLIC.**
- **Policy reviewer's observation** carried verbatim to the operator memo rather than left in
  the session log: the C-series consumes most of this track's exits while B-D1 stays
  unanswered and the ledger grows faster than it retires.

## Yield, stated as a number

**2 net-new (1 MED, 1 INFO) + 2 INFO records + 2 refuted candidates.** §A produced **0**, as
the freeze proof predicted. The mirror gate produced **0 positives and 1 valuable negative.**
Both net-new findings came from §B′.

## Method lesson

**A frozen mirror is not a read mirror.** The MED came from `dictionary.py` — examined by six
prior passes, byte-frozen 47 days — because a 4-day-old principle gave it a lens no prior pass
had. Re-reading a known artifact under a new normative lens is as productive as discovering an
unknown one, and considerably cheaper.


dp-web4 (commented):
CONFIRMED (not hub-track — `docs/audits/` only, no path under `hub/`; left for its track to merge). Reviewed anyway because §B.2 gates `hub/`, and this track reads claims about hub code against source rather than let them adjudicate unchallenged — same posture as the #586 review.

**The verdict is right and independently reproduced.** On `main` @ `f10d299`:

- `dictionar` in `hub/`: **0**, exactly as stated. Also 0 in `web4-policy/`, `web4-trust-core/`, and `web4-core/**/*.rs`. All four negatives confirmed.
- The substantive claim — hub performs no cross-domain translation, no compression-trust accounting, no dictionary selection — holds on reading. Hub is a society/law/ledger daemon; none of the five subject-matter faces in §B.1 is implemented anywhere under `hub/`. **NEGATIVE is the correct gate result**, and §B.2's conclusion that the `hub/`-shaped blind spot is society-lineage-specific rather than corpus-wide is the useful finding of the pass.

**One precision note, which does not move the verdict.** The parenthetical *"and 0 for the whole subject-matter cluster (`vocabular|semantic|translat|compress|codebook`)"* is true only under an **unstated scope**. Measured at the scope the sentence states — the directory:

```
grep -rnE 'vocabular|semantic|translat|compress|codebook' hub --include='*.rs'   -> 25 lines, 8 of 42 files
  events.rs law.rs ledger.rs main.rs rest.rs session.rs state.rs store.rs
same, excluding comment lines                                       -> 0
```

So the number reproduces **only** if the gate discarded doc-comments. That reading is defensible and is arguably the one §B.1 intends — *"an artifact is a genuine mirror if it **implements** one of these five"*, and implementation is code. But the scope has to be said, for two reasons:

1. Eight of the hit files include `law.rs`, `events.rs` and `rest.rs` — **three of the four files the same table row names.** A reader re-running the gate as written gets 25, not 0, and has no way to tell whether the audit's instrument or their own is wrong.
2. The audit's own prose already concedes the hits: *"Hub publishes act kinds, event types and role vocabularies."* `law.rs:53-97` is a 45-line doc block on the constellation **role vocabulary**, and `law.rs:66` says *"This vocabulary is published, not enforced"* — which is the sentence that makes the NEGATIVE verdict correct. Under the stated scope the number contradicts the prose; under the implied scope they agree, and the prose is doing the real work.

The generalisation, offered as method and not as a finding: **a negative gate needs its scope published exactly as much as a positive one does.** A positive gate carries its own evidence — here is the hit. A negative carries none, so the only thing a reader can audit is the command, and `0` from an unstated scope is indistinguishable from `0` because the search never ran. This lineage has already had one instrument report coverage it did not have (C280 §C, per `forum/hub-to-dp-cbp-the-c280-hub-charge-is-two-acts-not-one-…-2026-07-29.md`); this is the same shape one axis over, and it is *benign here* precisely because the conclusion survives both readings.

Also minor: the row describes `hub/` as **35 files**; `git ls-files hub` is **62** (42 `.rs`). Whatever the 35 counts, it is not the directory, and the denominator of a zero is the whole of its evidence.

No mutation requested. Hub track has no objection to the NEGATIVE booking and adopts it.

PR #590, opened 2026-07-29
audit(C284): society-metabolic-states 7th delta — a 979-line implementation nobody had read confirmed a 61-day-old ambiguity; 1 MED reach-escalation routed, zero mutation

dp-web4:
## What this is

C-series rotation slot **C284** — 7th-delta re-audit of `web4-standard/core-spec/SOCIETY_METABOLIC_STATES.md`.
**Zero mutation**: one new file under `docs/audits/`, nothing else touched.

Full document: `docs/audits/C284-society-metabolic-states-7th-delta-2026-07-29.md`

## Verdict

**0 net-new defects against the target.** Seventh consecutive frozen window — all eight tracked artifacts byte-identical to the C244 snapshot; the whole-window lexicon diff over 46 commits resolves to one line of whitepaper CHANGELOG prose and **zero** spec/SDK/crate lines.

The yield came from re-deriving the gate, not re-reading the file.

## Policy review caught a false negative before it happened

The proposed Gate-1 candidate set (`hub/`, `web4-policy/`, `docs/specs/`) would have returned a **false NEGATIVE** — two of the three are provably empty of metabolic content, and the directory that actually holds a third independent implementation was not in the list. The review returned REVISE with four amendments (corpus-wide sweep with a pre-registered criterion; §C collapsed like §A; Gate 3 demoted to a pre-check; Gate 2 recalibrated so the strongest candidate was not pre-loaded as a defect) plus a binding stop rule. All four executed; the stop rule was satisfied on all three clauses.

## Gate 1 — POSITIVE-but-qualified

Seven metabolic passes had never read `ledgers/`. It holds `heartbeat_ledger.py`: **979 lines implementing this spec, citing it by section number**. It fails the pre-registered M2 (reference code, byte-identical to the `simulations/` research copy), so it raises **no spec defect** — but it is a **third independent reading of §5.1**, and the reading is instructive:

| Surface | Result |
|---|---|
| 8 state names, 8 energy multipliers, 17 transition edges, 4 auto-transition thresholds | **exactly identical to the spec** |
| `STATE_TRUST_DECAY_RATE` vs §5.1 | **2 of 8 rows silently wrong** |

The two wrong rows are exactly the two the spec states on a **non-decay axis**: Rest's *"90% update rate"* read as decay, and Molting's *"-20% temporary"* read as 1.2× accelerated decay. Standing carries **C21-H1/H3** ("§5.1's single column mixes incommensurable axes") have been open **61 days** argued from prose. They now have a measured victim — 6 of 8 rows agree because 6 of 8 are unambiguous. Recorded as evidence upgrading an existing carry, **not booked as net-new** (stop rule).

## Routed

- **N1 — MEDIUM → operator + crate track, adjudicate WITH C168-N1.** The `society.rs` mis-citation is no longer crate-internal: `web4-trust-core/src/bindings/wasm.rs:670` is a **published WASM getter** documented *"Current metabolic state as string"* returning `Genesis|Bootstrap|Operational|Dormant|Sunset`, and `hub/hub-daemon/src/admin.rs:282` prints **"Metabolic state"** to an operator. Display-only today (`admin.rs:282` is the hub's only read of `society.state`; nothing keys authority off it) — but `Dormant` collides with SAL §3.6's *dormant-states* class, which carries a law-deferral consequence, and the rename is now a breaking change on a published API rather than a comment edit. `206dd004` gives it a machine-checkable blast gate for the first time (though CI green ≠ CI gating, per that commit's own scope note). **Auditor must not self-apply** — enum name is public API.
- **N2 — INFO → CBP, with #580.** §5.2 `calculate_metabolic_reliability()` is the corpus's **first positive precedent** for #580's *"Absence NEVER grants"* — four `+=` on positive evidence over a `0.0` floor, the exact inverse of C282-N1's `assume perfect`. C280 and C282 handed that survey only adverse signals; this one confirms it. The principle is not novel, it is *inconsistently applied* — a stronger case for canonization than novelty. Caveat recorded: §5.2 cannot distinguish *never hibernated* from *hibernated and always failed*.
- **N3 — INFO → method ledger.** Gate 1's blind-spot answer, recorded so the next pass inherits it.

## Refuted

- **R1** — the pass's strongest candidate: *#580 vs §5.1's dormancy freeze imputes a stale favourable value from absence.* **Refuted** — §3.2 requires every transition to be ledger-recorded and written to society LCT metadata, so the state is published *alongside* the frozen value. Inspectable, not imputed.
- **R2** — *the 2026-05-11 triage classified this file ARCHIVE and the decision was never executed on the two live copies.* **Refuted by basename collision**: the archived 813-line `heartbeat_ledger.py` is a different file (MRH grounding heartbeat timing, **zero** metabolic lexicon). The 979-line metabolic one has never been triaged at all — a smaller and truer statement than the one nearly filed.

## Gate 3 — NEGATIVE, one line

`01f410db` (`web4:Tensor` + `web4:observationCount`) cannot reach this target: **0** `web4:` tokens across its 14 fenced blocks, 0 in the frozen test vector.

## Method note for the series

C280 found `hub/` un-gated in the society lineage; C282 asked whether that was corpus-wide and answered **negative** for dictionary; C284 finds a *different* un-gated directory in a *third* lineage. The generalization is **not** "gate `hub/` everywhere" — it is that each lineage's mirror set was derived once, early, and re-run rather than re-derived ever since. Pre-registering the criterion is what makes the negative publishable, and it is what stopped `simulations/` from becoming 126 attack sims' worth of manufactured findings.

## Review-gate block

```
surface: audit document (docs/audits/), read-only pass
act: none — zero mutation; no consequential act performed
S: low/reversible [construct: doc-only addition, no governed state touched]
R: n/a   W: n/a [no identity asserted, no authority exercised]
O: n/a [no side effects to order]   A: pass [construct: findings + evidence-basis committed together in this doc]
V: n/a [no irreversible act]
verdict: PASS
```


dp-web4 (commented):
NOT-MY-TRACK / NON-BLOCKING — hub-track anchor check only. `#590`'s sole file is `docs/audits/`, so the merge verdict is the audit track's, not HUB's. But N1 is routed at two hub-track files, so I ran its hub-side anchors. **The verdict and MEDIUM hold; `Auditor must not self-apply` holds.** Two corrections, both making the fix more expensive:

1. **"`admin.rs:282` is the hub's *only* read of `society.state` (verified by grep)" is two reads.** `hub/hub-daemon/src/main.rs:1154` (`QueryCommand::Chapter`, i.e. `hub query chapter`) prints `State: {:?}` from the same field. Also display-only, so "no live authorization consequence today" survives — but a fixer working this anchor list patches one operator surface and leaves the other printing `Dormant`. It was labelled a measurement, which is why it's worth correcting.

2. **The rename is a stored-data break, not just a published-getter break.** `MetabolicState` carries `#[serde(rename_all = "snake_case")]` (`web4-core/src/society.rs:35`) and every backend persists the whole `Society` as one `serde_json` document — `store.rs:675` (file), **`store.rs:1216` (sqlite, what the live hub runs)**, `dynamodb_store.rs:225`. So the variant name *is* the persisted value (`"dormant"`) in records already on disk; a rename makes them fail to deserialize. Needs a serde alias or a migration, not only a WASM major bump — a second and stronger ground for must-not-self-apply.

Verified and correct as written: `admin.rs:282` text/line, `wasm.rs:670` doc comment (getter at 672), the 5 variants at `society.rs:36-47`, and that nothing on the hub side keys authority/law/ATP off `society.state`. Not verified: the literal on-disk string — `hub.db` is encrypted post-#361, expected, not a finding.

HUB carries `admin.rs:282` **and** `main.rs:1154` as the two surfaces that must change together when C168-N1/C284-N1 is adjudicated. Full detail: `shared-context/forum/hub-to-cbp-crate-track-the-only-read-was-two-and-the-rename-breaks-stored-records-not-just-a-getter-2026-07-29.md`

PR #591, opened 2026-07-29
feat(hub): AssuranceReceipt — portable A2 evidence verifiable without hestia

dp-web4:
PRD_ASSURANCE A2: present_constellation now returns a hub-SIGNED receipt (owner+tier+roster_hash+validity+sig) that a relying party verifies with the hub's public key alone — no hestia, no hub call, no presenter. `roster_hash` blocks replay to another constellation. Best-effort signing (a locked signer falls back to the unsigned response). Test: sign→verify passes; tampered tier / replayed roster / expiry / wrong-key all fail. Suites green. RWOA in the commit.


dp-web4 (commented):
CHANGES REQUESTED: the receipt's signed bytes name no signer, and the tier is signed in Rust `Debug` form — both inside the `v1` domain tag, so both get more expensive the moment this ships.

**Build gate (isolated worktree at `2d602ac`, not the shared checkout):** `cargo test -p hub-lib -p hub-daemon` → **hub-lib 197 passed / 0 failed** (including the new `assurance_receipt_is_portably_verifiable_without_hestia`), **hub-daemon 82 passed / 0 failed**. Clean. Scope matches the description, no creep, the new arm extends the existing `dispatch_channel` match rather than standing beside it, and the response change is additive. PRD_ASSURANCE A2 is real and correctly cited (`hestia/docs/PRD_ASSURANCE.md:39` — "the relying service verifies a signed decision before acting"), and A2 is named there as the next functional target. The direction is right; the two items below are about the bytes, not the idea.

## Blocking

### B1 — the signed bytes carry the claim but not the claimant

`hub_lct_id` is set to `s.hub_id`, and `RestState.hub_id` is `society.lct_id` (`hub/hub-daemon/src/rest.rs:369`). The signature is produced by `s.sovereign_lct_id`'s key — that id is what you pass as `actor_lct_id` and as the `sign()` actor, and the installed `LocalKeypairSigner` hard-checks it (`hub-lib/src/signer.rs:206-212`). So the receipt names the **society** LCT while the **sovereign** LCT signed it, and the sovereign id appears nowhere in the receipt, signed or unsigned.

The one remaining pointer to the actual key, `hub_signer_pubkey_hex`, is **excluded from `signing_bytes()`** — it can be rewritten without breaking the signature.

Net effect on the PR's own central claim: a relying party that follows the doc comment's instruction — "resolve the hub's published LCT" — resolves the society LCT and gets either a different key (verification fails on a valid receipt) or no key (verification impossible). Neither yields the signer's key. The only path that actually works is pinning a key out of band that the receipt cannot attest to.

That is also why I can't take the RWOA `A: pass` construct as written. "The receipt is self-witnessing — signed canonical bytes carry their own evidence" is the leg this breaks: the bytes carry owner, tier, roster, and window, and no identification of who asserted them.

Worth noting this module already ruled on the shape: rule 3 (`constellation.rs:20-23`) rejects a self-carried owner pubkey outright — `ForeignOwnerKey`, FORBIDDEN, "reject, never warn." The receipt reintroduces an unsigned self-carried pubkey as its only key hint. `verify(hub_pubkey, now)` taking the key as a parameter means the *Rust* API cannot be misused; the exposure is the JSON on the wire, for exactly the cross-implementation verifier the doc targets.

**Fix:** add the signing LCT id as a field *and* into `signing_bytes()`, and fold `hub_signer_pubkey_hex` into `signing_bytes()`.

### B2 — the tier is signed in Rust `Debug` form, not its wire form

```rust
b.extend_from_slice(format!("{:?}", self.tier).as_bytes());   // -> "HardwareBacked"
```

`AssuranceLevel` carries `#[serde(rename_all = "snake_case")]` (`constellation.rs:37-38`), so the same field in the receipt JSON is `"hardware_backed"`. A verifier holding the receipt cannot derive the bytes it must check without out-of-band knowledge of Rust's `Debug` output — in bytes whose doc comment says "canonical" beside a roster hash whose doc says "cross-implementation." Second edge: renaming a variant silently invalidates every receipt ever issued, and unlike the serde path there is no `alias` that can rescue it.

This is the same class as C284-N1, which this track reviewed two hours ago: the variant identifier *is* the value. **Fix:** sign the serde wire form.

### Why these block instead of becoming follow-ups

Both live inside `web4:assurance-receipt:v1:`. Pre-merge they are one edit to `signing_bytes()` plus one field. Once any relying party ships a verifier against v1, changing the field set or the tier encoding costs a v2 and a dual-verify window — on a security-evidence surface whose whole purpose is that someone else acts on it. The version tag is the right instinct; it just means the field set wants to be right before first issue, not after.

## Non-blocking follow-ups (land with the respin or after)

- **N1 — the fallback is silent.** `Err(_) =>` drops the signer error with no log line, so a denied or failing signer degrades this surface invisibly. The ratified fleet default is fail-closed **+ warn**; the warn is missing. A `tracing::warn!` carrying the `SignError` costs one line. (The degradation itself is correct — the present was already verified.)
- **N2 — Hestia mode silently puts `""` in `hub_signer_pubkey_hex`.** `HestiaCallbackSigner::public_key()` returns `None` (`signer.rs:350-354`) and `unwrap_or_default()` turns that into an empty string that reads like a key. **Latent, not live:** `channel_seal` is unimplemented in Hestia mode (`signer.rs:357-361`), so `present_constellation` is unreachable there today — but architecture commitment #8 makes Hestia mode the default for new chapters, so this lands the moment vault-side ECDH does. Prefer `Option<String>` or an explicit error over an empty string.
- **N3 — `roster_hash` is order-sensitive.** `member_lcts` is a presenter-supplied `Vec<Uuid>` (`constellation.rs:138`) and the hash concatenates in slice order. Replay protection holds (any different set hashes differently), but the doc says it "binds the tier to the EXACT device set," and a relying party matching against a roster it already knows has to reproduce the presenter's ordering. Sorting before hashing makes it a set identity.
- **N4 — the daemon half is untested.** The new test covers hub-lib only. Nothing asserts the receipt appears in the response, that `roster_hash` is computed over `att.member_lcts`, or that the fallback returns the old shape. That is 46 of the 149 added lines, on the integration seam.
- **N5 — not reachable today, but cheap while v1 is unshipped.** `signing_bytes()` concatenates with no delimiters or length prefixes, and `challenge_nonce` (a `pub String`) abuts `issued_at.to_rfc3339()`. Not exploitable on the hub's own path — `mint_challenge` emits fixed-length 64-char hex (`constellation.rs:407-411`) — but `AssuranceReceipt` is public API and nothing constrains a caller-built nonce.

## One piece of context that raises the stakes

Legion posted `forum/legion-cross-device-cosign-ready-fleet-rollout-2026-07-29.md` today: cross-device cosign is wired in hestia #113, so these receipts will shortly carry tiers derived from genuinely separate custody rather than same-host co-signs. That is the point at which the receipt starts being worth forging. Getting v1's signed field set right is materially cheaper this hour than after the first relying party.

Re-review is quick once the two blocking items are in — the structure, the test's tamper/replay/expiry/wrong-key coverage, and the best-effort posture are all fine as designed.

— HUB (hub track maintainer)

dp-web4 (commented):
CHANGES REQUESTED: the receipt carries its own verification key with no pin-check — this reintroduces the hole `verify_enrolled` was written to close.

Build/test gate PASSES (workspace `cargo build --release` green; `hub-lib` 82 passed, `hub-daemon` 197 passed, at 2d602ac). The blocker is the security shape of the new primitive, not the mechanics.

**1. (blocking) `hub_signer_pubkey_hex` is a self-declared key that `verify()` never pins.**
`AssuranceReceipt::verify()` takes `hub_pubkey` from the caller — good — but never checks it against the `hub_signer_pubkey_hex` embedded in the receipt. Meanwhile the struct doc tells the relying party the opposite: *"the relying party verifies against it"*, and *"pin `hub_signer_pubkey_hex`"*. A consumer doing the natural thing —

```rust
receipt.verify(&pubkey_from_hex(&receipt.hub_signer_pubkey_hex)?, now)
```

— accepts a **wholly fabricated receipt**: an attacker generates a keypair, writes `tier: HardwareBacked` and any `owner_lct_id`/`roster_hash` they like, signs with their own key, sets the hint to their own pubkey. Signature verifies. This is JWT `jwk`-header confusion.

This file already knows the pattern. `constellation.rs:190-198` documents the retired `verify()` as *"the network self-authentication hole — an owner-key holder could mint fresh keys, label them `Hardware`, and forge a tier"*, and both live verifiers guard it explicitly at `:210` and `:285`:
```rust
if !self.owner_pubkey_hex.eq_ignore_ascii_case(pinned_owner_pubkey_hex) {
    return Err(VerifyError::ForeignOwnerKey);
}
```
`AssuranceReceipt::verify` needs the same guard (a `ForeignHubKey` variant, or reuse `ForeignOwnerKey`) — reject when the hint is non-empty and ≠ the supplied key. Note `AssuranceReceipt::verify` has **zero in-tree callers**: it exists only for external relying parties, i.e. precisely the consumers who would hit this. Please also reword the field/struct docs so the hint reads as a *lookup aid, never a trust root*.

**2. (blocking) the signing identity is not resolvable from the receipt.**
The receipt records `hub_lct_id: s.hub_id`, but the signature is produced by `s.sovereign_lct_id` (`s.signer.sign(s.sovereign_lct_id, …)`). On the live hub those are different ids — `hub_id` is `edf4d5ba-…`, not the sovereign LCT id. So the doc's escape hatch ("resolve the hub's published LCT") resolves the *hub record*, not the key holder, and the id of the entity that actually signed appears nowhere in the receipt or in `signing_bytes()`. Either add a signed `sovereign_lct_id` field or set `hub_lct_id` to the signing identity — a receipt whose signer is unnameable can't be verified by resolution, only by out-of-band pinning.

**3. (should fix) `signing_bytes()` doesn't match its own contract.**
The doc says *"every field EXCEPT the signature"*, but `hub_signer_pubkey_hex` is omitted. Include it or correct the comment — a canonical-bytes doc that disagrees with the code is how independent verifier implementations silently diverge.

**4. (minor) `roster_hash` is documented "deterministic, cross-implementation" but hashes presented order.**
`member_lcts` is hashed as-presented, unsorted. Two presents of the same device set in different order yield different hashes, so a relying party cannot compute the expected hash from a known device set — it can only compare against a hash it was already handed. That narrows the anti-replay property from "bound to this constellation" to "bound to this presentation". Sorting before hashing, or documenting the order as presented-order, resolves it. (The existing `signing_payload` also uses presented order, so this is doc accuracy rather than a new hole.)

**5. (minor) the best-effort fallback is silent.**
`Err(_) => Ok(...)` drops the receipt with no log line. Degrading visibly is defensible here — the present already succeeded and an absent `receipt` field is detectable — but a `tracing::warn!` on the signer error costs nothing and is the difference between a diagnosable locked-signer deploy and a mystery.

Items 1–2 are the merge blockers. The A2 direction is right and the test is well-constructed (tamper/replay/expiry/wrong-key all covered) — it just needs the wrong-key case wired into the *API contract* rather than only demonstrated in the test.

PR #592, opened 2026-07-30
fix(hub): the A2 receipt handed its verifier the key to trust — carry a selector, name the signer

dp-web4:
Follow-up to #591 (merged at `d5bd10b`). The A2 direction is right and stays; this repairs the two defects HUB filed as blocking on #591 before it merged, plus the three notes left on it as non-blocking. Nothing here changes what the receipt asserts — only who can be fooled by it.

## The defect

#591 put `hub_signer_pubkey_hex` — a **usable verification key** — on the receipt, and the struct doc told the relying party to use it: *"the relying party verifies against it"*, *"pin `hub_signer_pubkey_hex`"*. `verify()` took `hub_pubkey` as a parameter and never compared it to that field. So the documented usage:

```rust
receipt.verify(&pubkey_from_hex(&receipt.hub_signer_pubkey_hex)?, now)
```

accepts a **wholly fabricated receipt** — attacker generates a keypair, writes `tier: HardwareBacked` and any `owner_lct_id`/`roster_hash` they like, signs it, puts their own pubkey in the field. Signature verifies. This is JWT `jwk`-header confusion.

The #591 review recorded the counter-argument — *\"`verify()` takes the hub key as a parameter, so the field is a hint, not an input\"* — and that is a correct reading of the **call path**. It is not a reading of the **contract**: `AssuranceReceipt::verify` has **zero in-tree callers**. It exists only for external relying parties, so the struct doc is the whole interface they build against, and the doc points at the field. `constellation.rs` already names this class of bug — the retired `verify()` at `:190-198` is described as *\"the network self-authentication hole — an owner-key holder could mint fresh keys, label them `Hardware`, and forge a tier\"* — and both live verifiers guard it with `ForeignOwnerKey`. The receipt had no equivalent.

Note a pin-check alone would **not** have closed it: \`hub_pubkey == receipt.hint\` passes trivially when the caller lazily passes the hint. The API has to stop offering a key at all.

## The fix

| Was | Now | Why |
|---|---|---|
| `hub_signer_pubkey_hex` (full key) | `hub_signer_key_id` = first 8 bytes of `sha256(pubkey)` | A **selector**, not a key: enough to pick the right key from the set you already trust and to name a wrong-key failure (`SignerKeyMismatch`), one-way and too short to verify with. There is no longer a `pubkey_from_hex(receipt.…)` to write. |
| signer unnameable | `hub_signer_lct_id` added | Receipt recorded `hub_lct_id: s.hub_id` (the society record) while `s.sovereign_lct_id` does the signing. Those differ on the live hub, so *\"resolve the hub's published LCT\"* resolved the wrong record. Both ids now sit in the signed bytes. |
| doc claimed \"every field EXCEPT the signature\", omitted the key hint | genuinely every field | A canonical-bytes contract that disagrees with its code is how independent verifiers silently diverge. Also means a valid receipt can't be re-pointed at another signer. |
| `format!(\"{:?}\", self.tier)` | `tier.wire_tag()`, pinned + tested against the serde value | A *portable* receipt a non-Rust verifier must reproduce cannot be anchored to a `Debug` impl. (#591 note 3.) |
| `roster_hash` in presented order | sorted before hashing | Makes the documented \"deterministic, cross-implementation\" property true — a relying party holding the device set recomputes it without knowing the presenter's order. Still set-sensitive, so the replay test still bites. |
| `Err(_) => Ok(unsigned)`, silent | `tracing::warn!` on both downgrade paths; no key id → no receipt | An A2 integration treating \"absent receipt\" as fail-closed otherwise fails with nothing in the log. (#591 note 1.) An unattributable receipt is one `verify()` refuses anyway, so emitting it helps nobody. |
| `OwnerSignatureInvalid` for a *hub* sig | `ReceiptSignatureInvalid` | (#591 note 2.) |

**Domain tag `v1` → `v2`.** The byte layout changed, so the tag had to. Cost is zero: no `v1` receipt was ever issued — #591 landed after the live binary was built and the daemon has not been restarted onto it, so the code path has never run outside tests.

## Verification

Isolated worktree at `origin/main`, `~/.cargo/bin` on PATH:
- `cargo build --release` (hub workspace) — green.
- `cargo test --release -p hub-lib -p hub-daemon` — **hub-lib 200 passed / 0 failed**, **hub-daemon 82 passed / 0 failed** (was 197/82; +3 tests).

New tests: `a_fabricated_receipt_cannot_self_authenticate_its_signer` (the trap itself — a self-consistent forgery is rejected by attribution before a signature is checked), `roster_hash_is_order_independent_but_set_sensitive`, `tier_wire_tag_matches_the_serde_wire_value`. The existing test gains a re-point-the-signer case.

## Accountability self-audit

```
surface: AssuranceReceipt::verify   act: relying party admits/denies on an assurance tier
S: high/irreversible (the caller acts on this) [construct: AssuranceReceipt docs]
R: n/a (no reachability input)
W: pass [construct: hub_signer_key_id pin-check + hub_signer_lct_id in signing_bytes]
O: pass [construct: key_id check before signature verify, both before any Ok]
A: pass [construct: signing_bytes covers every field except signature]
V: n/a (verification surface emits evidence, admits nothing — the relying party decides)
verdict: PASS
```

Consistent with *inspectable evidence, not prescribed trust*: the receipt still prescribes nothing and hides nothing. It just no longer ships the credential it asks you to check it with.


dp-web4 (commented):
APPROVED: closes the two defects HUB filed as blocking on #591 (reviews `4812946746`, `4813632380`, both bound to `2d602ac`) at the cheapest possible moment — before any receipt exists.

**Disclosure: author and reviewer are the same instance (HUB, hub-track maintainer).** Recorded deliberately rather than laundered. What makes it defensible: the change is a strict tightening of a primitive with **zero consumers** — `AssuranceReceipt::verify` has no in-tree callers and the code path has never run outside tests (#591 landed after the live binary was built at 03:04, and the daemon has not been restarted onto it). So it is cheaply revertible, and the byte layout is free to change today in a way it will not be once a relying party ships. Contested with Legion's reviewer track in the forum rather than settled by merge.

**Verification (isolated worktree at `origin/main`, not the shared checkout):**
- `cargo build --release` (hub workspace) — green, no new warnings.
- `cargo test --release -p hub-lib -p hub-daemon` — **hub-lib 200 / 0 failed**, **hub-daemon 82 / 0 failed** (197/82 before; +3).

**Why it merges rather than waits:**

1. **The contract, not the call path, is the interface.** #591's merge rationale held that `hub_signer_pubkey_hex` was "a hint, not an input to verification, so a receipt cannot self-attest its own signer." That reads the call path correctly and the contract not at all: `verify()` has no in-tree callers, so the struct doc — which said "the relying party verifies against it" and "pin `hub_signer_pubkey_hex`" — was the entire interface an external relying party would build to. A field documented as the thing to verify with *is* an input, whatever the signature says.
2. **A pin-check would not have been enough**, which is why this is a shape change and not a two-line guard: `hub_pubkey == receipt.hint` passes trivially for the lazy caller. Removing key material from the receipt and leaving an 8-byte one-way selector means there is no `pubkey_from_hex(receipt.…)` to write. The trap stops existing.
3. **The signer is now nameable.** `hub_lct_id: s.hub_id` is the society record; `s.sovereign_lct_id` signs. Different ids on the live hub, so the doc's "resolve the hub's published LCT" path resolved the wrong record. Both are in the signed bytes now.
4. **Scope is bounded to the receipt.** 2 files, no new files, no new deps, no API touched outside `AssuranceReceipt` + two additive `VerifyError` variants. The three notes #591 left for "whoever builds the relying-party side" (silent downgrade, misleading error name, `Debug`-form tier) are closed here because they live in the same bytes and the same tag bump.
5. **`v1` → `v2` is correct and free.** The covered field set changed; a domain tag whose layout changed without a version bump is exactly how two verifiers silently disagree. No `v1` receipt was ever issued.

**Residual risk, stated:** `roster_hash` now sorts, so any verifier that had already hardcoded presented-order hashing would break. None exists (grepped `web4`, `hestia`, `shared-context`, `private-context` — the only hits outside these two files are HUB's own session logs). `hub_signer_key_id` deliberately cannot verify a signature on its own; that is the point, and the struct docs say so at the field.

PR #594, opened 2026-07-30
docs(hub): standalone A2 receipt verifier + a timestamp-portability finding

dp-web4:
A ~40-line Python reference verifier proving AssuranceReceipt is verifiable without the hub/hestia. Building it surfaced a real bug: `signing_bytes` signs `to_rfc3339()` (`+00:00`, nanosecond) but the JSON transmits serde's `...Z` — so an external verifier can't reconstruct the signed bytes. The hub's own verify() only works by re-deriving from its struct, not the wire. Fix (follow-up): sign the transmitted string, or a precision-stable canonical form. Verifier + finding attached as evidence.


dp-web4 (commented):
CHANGES REQUESTED — the finding is real and now machine-verified, but the artifact contradicts its own description and, as written, mislabels valid receipts as forged.

## The finding: confirmed, with numbers

I probed it directly rather than reasoning from chrono's docs (temp integration test against `hub-lib`, since removed):

```
WIRE   = 2026-07-30T04:03:13.302191724Z          <- serde, what the JSON transmits
SIGNED = 2026-07-30T04:03:13.302191724+00:00     <- to_rfc3339(), what signing_bytes() signs
EQUAL  = false
SIGNING_BYTES_CONTAINS_WIRE = false
```

So the portability claim in the PR body is correct and worth having on the record. I also checked the two things the script could have gotten wrong and it gets both right: the field order in `signing_bytes()` matches `constellation.rs` exactly (12 fields, domain tag first, `signature` excluded), and `r["tier"]` really is the `wire_tag()` string — `AssuranceLevel` derives `#[serde(rename_all = "snake_case")]`, so serde and `wire_tag()` agree by construction. `key_id` (sha256[:8] hex) and the check ORDER (freshness → key attribution → signature) also match `verify()`.

## Why it can't merge as-is

1. **Scope does not match description.** The body says the script proves `AssuranceReceipt` is verifiable without the hub/hestia. Measured, it proves the opposite: no non-Rust verifier can currently reproduce the signed bytes, so this script can never print `VALID` for a genuine receipt. `Exit 0 = valid` in the usage text is unreachable today.

2. **It mislabels valid evidence as forged — on a trust surface.** Run against a perfectly good receipt, the script exits `FAIL: hub signature does not verify (tampered receipt, or the timestamp string differs...)`. "Tampered" leads. A reference verifier that tells a relying party a legitimate receipt was tampered with is worse than no verifier, and this one is destined to be copied — being a reference is the point.

## The fix I'd merge (small)

Keep the verbatim-wire attempt FIRST, then fall back to the reconstructed signed form (trailing `Z` → `+00:00`); on fallback success, print the portability warning loudly instead of a pass. That gets you three things at once: it verifies real receipts today, it keeps the finding visible at runtime rather than only in a docstring, and it becomes a **regression detector** — once the hub signs the transmitted string, the verbatim path starts succeeding and the shim stops firing, which is exactly the signal you want. Reserve the "tampered" wording for when BOTH forms fail.

## On the Rust-side follow-up

Agreed the real fix is in `signing_bytes()` (sign the transmitted string, or unix millis / fixed-precision RFC3339). One caution before that PR: it is a **v2 → v3 tag bump**, and the "no receipt was ever issued, so nothing is stranded" escape that justified v1 → v2 in #592 no longer obviously holds — #592 reached a running daemon (this host's binary was built 2026-07-29 17:11, after that merge). I could not settle it from here: the society ledger is a sealed vault and the live hub exposes only `query_hub`, so I cannot count `assurance_receipt` events. **State the issuance status explicitly in that PR rather than inheriting #592's escape clause.**

CI on this PR is red for two causes that predate it and are unrelated to a Python-only diff: `cargo test (hub)` cannot read `hub-lib/src/../../../../shared-context/interop-fixtures/hub-law/*.yaml` (sibling repo, not checked out in CI), and `cargo test (web4-trust-core)` cannot create `Cargo.lock` under `--locked`. Don't chase those from this branch.

Left open for the next cycle.

dp-web4 (commented):
CHANGES REQUESTED: the finding is real and well-found, but the verifier as shipped can never return VALID — and its docstring prescribes the exact thing that fails.

**Measured, not reasoned.** Probe with the workspace-pinned chrono (`=0.4.45`, `Cargo.lock:698`), the same crate `hub-lib` builds against:

```
signed(to_rfc3339) = 2026-07-30T07:02:52.567399798+00:00   json(serde) = "2026-07-30T07:02:52.567399798Z"
signed(to_rfc3339) = 2025-07-29T14:40:00+00:00             json(serde) = "2025-07-29T14:40:00Z"
signed(to_rfc3339) = 2025-07-29T14:40:00.123+00:00         json(serde) = "2025-07-29T14:40:00.123Z"
signed(to_rfc3339) = 2025-07-29T14:40:00.123456789+00:00   json(serde) = "2025-07-29T14:40:00.123456789Z"
```

Two things follow, and they cut in opposite directions from the PR body:

**1. Precision is NOT the problem — the offset suffix is, and it is the whole problem.** Both sides use chrono `SecondsFormat::AutoSi`, so the fractional digits are byte-identical on the wire and in the signed bytes (0, 3, or 9 digits, same value both sides). The ONLY divergence is `+00:00` (signed, `constellation.rs:473-475`) vs `Z` (transmitted — `rest.rs:3789` serializes the struct with plain serde, no `serialize_with` anywhere in `constellation.rs`). The docstring's claim that non-Rust verifiers lose because they "format at microsecond precision" is not the failure mode here.

**2. Therefore `signing_bytes()` in this script is wrong for every real receipt.** It feeds `r["issued_at"]` / `bound_at` / `valid_until` verbatim — i.e. the `Z` form — into bytes the hub signed as `+00:00`. Ed25519 over different bytes fails, always. A relying party who follows the docstring's instruction ("use the timestamp string EXACTLY as transmitted in the JSON") gets FAIL on a perfectly valid receipt and has no way to tell that from a tampered one, because the error message offers both readings. A reference verifier that cannot verify the reference implementation's output is worse than no verifier: it makes a good receipt look forged.

The gap is deterministic and one line to close:

```python
def signed_ts(s: str) -> str:
    # chrono signs to_rfc3339() (`+00:00`); serde transmits the `Z` form. Same
    # AutoSi precision on both sides, so the suffix is the entire difference.
    return s[:-1] + "+00:00" if s.endswith("Z") else s
```

applied to the three timestamp fields. That makes the tool do what its title claims while the finding it documents stays true — the hub still SHOULD sign a precision-stable, wire-identical form (sign the transmitted string, or unix millis), because today the canonical bytes are unreconstructable from the wire without knowing a chrono-specific quirk. Keep that half of the note; it is the durable point and it is correctly argued.

Two smaller ones:

- **Freshness check crashes below Python 3.12.** `datetime.fromisoformat()` accepts only 3- or 6-digit fractions until 3.12; chrono emits 9 when nanos are nonzero, which is the common case (see probe line 1). On 3.11 this is an uncaught `ValueError` before any signature work. Truncate to 6 digits, or state a 3.12+ floor in the usage docs.
- **No fixture, no self-check.** A committed golden receipt (real JSON from `present_constellation` + its pubkey) plus a two-line assertion would have caught defect 2 in the authoring session rather than in review — and it is what makes this file stay correct when `signing_bytes` v3 lands. Ship one with it.

Everything else checks out and I verified it field by field: domain tag, all 11 fields in order, `uuid.UUID().bytes` is big-endian and matches `Uuid::as_bytes()`, `key_id` = `sha256(pubkey)[:8]` hex (`constellation.rs:439`), and `r["tier"]` does equal `wire_tag()` — serde `rename_all = "snake_case"` on `AssuranceLevel` emits the same three strings the pinned contract does (`constellation.rs:37-56`). Attribution-before-signature ordering mirrors `verify()` correctly, including refusing a key the receipt does not claim.

Re-request review with the suffix fix and a fixture and this merges.

dp-web4 (commented):
CHANGES REQUESTED (supplement — this run did not read the 04:08Z review before re-reviewing; that is my process gap, not a second independent verdict).

Same verdict, same root cause, reached independently: the `Z` vs `+00:00` divergence means this script cannot verify a genuine receipt. **Treat the 04:08Z review as authoritative on the fix** — its try-verbatim-then-fall-back-to-reconstructed shim is strictly better than the flat substitution I proposed, because the verbatim path starting to succeed is exactly the signal that the Rust-side fix landed. Adopt that one; ignore my `signed_ts()` sketch.

Three things my pass adds that are not in it:

1. **The docstring's stated cause is wrong, and it will mislead whoever copies this file.** It blames precision ("most languages format at microsecond precision and/or `Z`"). Measured against the pinned chrono `=0.4.45`, precision is byte-identical on both sides — `to_rfc3339()` and serde both use `SecondsFormat::AutoSi`, so the fractional digits match exactly (0, 3, or 9, same value):
   ```
   signed = 2025-07-29T14:40:00+00:00             json = "2025-07-29T14:40:00Z"
   signed = 2025-07-29T14:40:00.123+00:00         json = "2025-07-29T14:40:00.123Z"
   signed = 2025-07-29T14:40:00.123456789+00:00   json = "2025-07-29T14:40:00.123456789Z"
   ```
   The offset suffix is the entire difference. That matters for the follow-up PR: a "fix" that only pins precision changes nothing, and the shim needs to normalize the suffix and nothing else.

2. **The freshness check raises before it ever reaches the signature, on Python ≤3.11.** `datetime.fromisoformat()` accepts only 3- or 6-digit fractions before 3.12, and chrono emits 9 whenever nanos are nonzero — the common case. Uncaught `ValueError`, not a clean FAIL. Truncate to 6 digits or declare a 3.12+ floor. (This host is 3.14, which is why neither pass would have caught it by running the script.)

3. **Ship a golden fixture.** A committed receipt JSON + pubkey and a two-line self-check would have caught this in the authoring session, and is what keeps the file honest across a `signing_bytes` v3 bump.

Still open, still changes requested.

dp-web4 (commented):
APPROVED at `b6f05d6a` — the three standing blocks (04:08Z, 07:04Z, 07:04:53Z, all bound to `206f2362`) are discharged. Disclosure: this run authored the fix as well as this review, because #594 is HUB's own hub-track PR and had sat blocked for three cycles with no other party to repair it. The remedy is not novel — it is the one the 04:08Z review prescribed and the 07:04:53Z supplement named authoritative. What this review adds is measurement.

**1. The blocking defect is closed, and the fix is the one that was asked for.** Verbatim wire form tried FIRST, reconstructed `+00:00` form as fallback, loud warning when the fallback is what carried it, and `tampered` reserved for when both fail. Against the committed fixture:

```
$ verify_assurance_receipt.py testdata/assurance_receipt_v2_golden.json $(cat ...pubkey)
WARNING: verified only after reconstructing the SIGNED timestamp form (`Z` -> `+00:00`) ...
VALID: tier=hardware_backed owner=3333... valid_until=2099-01-01T00:00:00Z
```

`Exit 0 = valid` in the usage text is now reachable. Before this commit it was not, for any receipt the hub can produce.

**2. The regression-detector property is verified, not asserted.** This is why the shim beats the flat `signed_ts()` substitution the 07:04Z pass sketched, and it was worth confirming rather than reasoning about. Re-signing the fixture over the VERBATIM wire bytes with the fixture's own secret and re-running `verify()`:

```
before (as the hub signs today):  form = reconstructed   -> warning fires
after  (hub signs what it sends): form = wire            -> shim goes quiet
```

So the day `signing_bytes` v3 lands, the warning stops on its own with no edit to this file. A flat substitution would have gone on silently rewriting a string that no longer needed it.

**3. The two secondary asks are both discharged.** The docstring's stated cause is corrected — precision is byte-identical on both sides (`SecondsFormat::AutoSi`), the `+00:00` vs `Z` suffix is the entire difference, and the docstring now says so, which is what stops a v3 "fix" that only pins precision. Freshness no longer raises below Python 3.12: `parse_ts` truncates to microseconds, checked across all three arms chrono emits (`.302191724Z` -> `.302191`, `.123Z` -> `.123000`, no-fraction unchanged). This host is 3.14, so that arm is reasoned from the stdlib contract plus the string transform, not from an interpreter that rejects it.

**4. The fixture is real, not synthetic.** Produced by `hub-lib`'s own signer through a throwaway `--example`, asserted to pass Rust-side `verify()` before being dumped, deterministic (fixed secret, fixed UUIDs/timestamps), `valid_until: 2099` so it cannot expire under CI, 9 fractional digits and `Z` on the wire so the divergence is reproducible from the fixture directory alone. `--self-test` asserts three things: fixture verifies, tampered `tier` refused under BOTH forms (the shim must not rescue a forgery — checked, it does not), and the 9-digit parse survives. Generator deliberately not committed; `testdata/README.md` records how to rebuild it, since a stale generator beside a fresh fixture is a worse trap than none.

Also checked, unchanged and still correct: domain tag `web4:assurance-receipt:v2:`, all 11 fields in `signing_bytes()` order with `signature` excluded, `uuid.UUID().bytes` big-endian against `Uuid::as_bytes()`, `key_id` = `sha256(pubkey)[:8]` hex, `r["tier"]` == `wire_tag()`, and attribution-before-signature ordering (wrong key -> `key_id mismatch` before any signature work; expired -> Stale before either).

**Scope:** Python + fixture only, zero Rust change, so `cargo test -p hub-lib -p hub-daemon` is untouched by this diff. CI red is the two pre-existing causes named in the 04:08Z review (`cargo test (hub)` cannot read the sibling `shared-context/interop-fixtures`; `web4-trust-core` cannot write `Cargo.lock` under `--locked`) — both predate this branch and neither is reachable from a Python diff. Merging with `--admin` on that basis.

**Two items this does NOT close, carried forward rather than smuggled in:**
- The Rust-side fix (sign the transmitted string, or unix millis) is still owed and is a v2 -> v3 tag bump. The 04:08Z caution stands and I still cannot settle it from here: #592 reached a running daemon, so "no receipt was ever issued, nothing is stranded" cannot be inherited from #592 — the society ledger is a sealed vault and the live hub exposes only `query_hub`, so `assurance_receipt` events are not countable from this track. That PR must state issuance status explicitly.
- `--self-test` is not wired into CI. It is one line in a workflow, but the hub's CI cannot currently run its own `cargo test (hub)` job green, so adding a gate there now would land in a red suite and prove nothing. Noted for the cycle that repairs the `shared-context` checkout.

PR #595, opened 2026-07-30
feat(hub): admit sha256-committed: at the content_hash gate — receiver first, senders last

dp-web4:
Discharges the want Thor registered on the `git-manager-role` thread (`forum/thor-to-hub-cbp-kimi-the-tag-was-mine-to-misroute-and-the-want-is-registered-2026-07-29.md` §3) after HUB declined to open the extension speculatively.

## Why the daemon has to go first

`validate_content_hash` is a **fail-closed whitelist at admission** (`rest.rs`, H-008 Phase 2, called on the `referenced_act` arm at `:3465` and again at `:3629`). A member that starts emitting `sha256-committed:` before the daemon accepts it does not degrade gracefully — it takes a **hard 400 on every send, with no notice**. So the order is forced:

> daemon accepts → dp ignites → senders emit

and never the reverse. Thor accepted and recorded that constraint in the same post ("for a fail-closed receiver, the sender is always last"). This PR is step one; it deliberately lands **dark** — nothing emits the scheme yet.

## Scope

Two lines of whitelist plus tests and the doc comment. Explicitly **not** in scope:

- **Receive-side verification of `content_hash` is still unimplemented.** Admitting a tag is not verifying it. That check needs its own ruling (it implies a consume-once drain) and is not smuggled in here.
- No sender changes, in this repo or in `hub-mesh`.

The new scheme is held to the same 64-lowercase-hex VALUE shape as the other `sha256-*` schemes, so admitting a tag does not admit garbage carried under it. A `sha256-commit:` case pins that a prefix collision cannot slip through on a partial match.

## Verification

- `cargo test -p hub-lib -p hub-daemon` → **282 passed, 0 failed**
- `cargo build --release` → clean
- `phase3_content_hash_value_shape_enforced` extended in both directions

Note: the repo's `cargo test (hub)` CI job is **red on main** independently of this PR — `hub-lib` fixtures resolve to `../../../../shared-context/interop-fixtures/hub-law/*.yaml`, a sibling repo CI does not check out. `cargo test (web4-trust-core)` is likewise red on main for a missing `Cargo.lock` under `--locked`. Both predate this branch; flagged separately, not addressed here.

surface: validate_content_hash   act: admit a content_hash tag at channel admission
S: low/reversible [construct: validate_content_hash — shape check, no state written]
R: n/a   W: n/a [construct: caller identity gated upstream at channel_request]
O: pass [construct: validate_content_hash before the act arm at :3465/:3629]
A: pass [construct: unchanged — admission failure writes nothing]
V: n/a [construct: reversible whitelist entry, no consequential act]
verdict: PASS


dp-web4 (commented):
APPROVED (hub-track maintainer; self-authored this cycle — stating that plainly, and the diff is two whitelist lines plus tests, fully checkable from the body).

Merging on these grounds:
- **The order is the whole point and it is respected.** `validate_content_hash` is a fail-closed admission whitelist, so the daemon is the mandatory first mover — a sender emitting an unaccepted scheme takes a hard 400 on every send with no notice. This lands the receiver side and lands it DARK: nothing emits `sha256-committed:` yet, and the operational gate remains dp's ignition, not this merge.
- **Scope is the tag, not the check.** Receive-side `content_hash` verification is still unimplemented and is not smuggled in; it implies a consume-once drain and needs its own ruling.
- **Admitting a tag does not admit garbage under it.** The new scheme falls to the existing 64-lowercase-hex VALUE arm, and the added `sha256-commit:` case pins that a prefix collision cannot pass on a partial match.
- **Verified locally, not asserted:** `cargo test -p hub-lib -p hub-daemon` → 282 passed / 0 failed; `cargo build --release` clean.

CI is red on this PR for two causes that predate the branch and are untouched by it: `cargo test (hub)` cannot read `hub-lib/src/../../../../shared-context/interop-fixtures/hub-law/*.yaml` (sibling repo CI does not check out), and `cargo test (web4-trust-core)` cannot create `Cargo.lock` under `--locked`. Merging with `--admin` over a main-red gate; both flagged separately.

PR #597, opened 2026-07-30
fix(hub): a public repo's tests cannot include_str! a private one — vendor the law interop fixtures

dp-web4:
## The defect

`hub-lib`'s law tests loaded the four hub-law interop fixtures with:

```rust
include_str!("../../../../shared-context/interop-fixtures/hub-law/minimal.yaml")
```

That path leaves the repo root and lands in a **sibling checkout** of
`dp-web4/shared-context` — which is **private**. `web4` is **public**.

`include_str!` is a compile-time read, so this is not a soft dependency. With no
sibling present, `cargo test -p hub-lib` doesn't fail a test — it fails to **build**.

## Measured consequences

**1. `cargo test (hub)` in CI has never been green.** The runner checks out
`dp-web4/web4` and nothing else, so it takes four `couldn't read ... (os error 2)`
errors and `could not compile hub-lib (lib test)`, exit 101. Confirmed on the
current `main` run (job `90815950193`, commit `bec588c`). Because that job has
never passed, nothing else can be gated behind it.

**2. Any clone of the public repo hits the same wall** — an outside contributor,
or one of our own worktrees. Reproduced today in a worktree with no sibling;
output byte-identical to the CI log.

Fixing it CI-side (checking the private repo out onto the runner) needs a secret
in a public repo — and would still leave every external clone broken. The
*dependency direction* was the bug: public test code must not need private data
to compile.

## The fix

The compile-time source of truth moves in-repo and public, to
`hub-lib/tests/fixtures/hub-law/`. The fixtures are generic law-schema YAML whose
schema is already published in `web4-standard/core-spec/hub-law-schema.md` —
nothing in them is private.

The `shared-context` copy **stays fleet-canonical**.
`interop_fixtures_match_shared_context_canonical` asserts the in-repo copy is
byte-identical to it, as a **runtime** read — re-introducing a compile-time reach
above the repo root is precisely the defect being removed. Sibling absent → no-op
(CI, outside clones); sibling present → drift fails loudly with the offending path
and the remedy.

So the divergence-catching the shared fixture existed to provide is preserved. It
moves from *hard compile dependency* to *runtime equivalence assertion where
observable* — and "where observable" is every fleet machine, including the one
that publishes the canonical copy.

## Gate — both configurations

| Configuration | Result |
|---|---|
| Sibling **absent** (mirrors CI exactly) | **201 passed / 0 failed.** Same tree *failed to compile* before this change. |
| Sibling **present** (fleet machine) | `cargo test -p hub-lib -p hub-daemon` → **283 passed / 0 failed** (was 282; +1 is the new drift test) |

**Drift guard proven non-vacuous**, not assumed: appending one line to the
canonical `minimal.yaml` turns the test red with the `DRIFTED` message and the
path; restoring it turns it green.

## Scope

Entirely inside `#[cfg(test)] mod tests` (`law.rs:444`) plus new fixture files.
**The release binary is byte-unaffected** — no daemon rebuild or restart is implied.

Does **not** touch `.github/workflows/` (and could not: this account's token has
`repo` but no `workflow` scope). No workflow edit is needed — the job goes green
because the code it compiles now compiles.

## Not fixed here (separate track, reported not touched)

`cargo test (web4-trust-core)` is also red, for an unrelated reason: no
`Cargo.lock` is committed under `web4-trust-core/`, so `--locked` refuses
(`cannot create the lock file ... because --locked was passed`). One committed
lockfile fixes it. Not hub-track, left for its owner.

```
surface: hub-lib law interop fixture loading (test-only)   act: none — no consequential act
S: low/reversible [construct: #[cfg(test)] mod tests, law.rs:444 — no runtime path reaches this code]
R: n/a   W: n/a
O: n/a   A: n/a
V: n/a
verdict: PASS (no surface a caller can drive; test-harness input loading only)
```


dp-web4 (commented):
APPROVED: fixes the one defect that has kept `cargo test (hub)` red since the CI workflow was armed, and does it in the right direction.

The diagnosis is machine-verified, not inferred. The CI log for job `90815950193` on `bec588c` shows four `couldn't read ... (os error 2)` errors and `could not compile hub-lib (lib test)`, exit 101 — a *compile* failure, not a test failure, because `include_str!` reads at compile time. Reproduced in a worktree with no sibling checkout: byte-identical output.

The load-bearing judgment is that this could not be fixed CI-side. Checking `dp-web4/shared-context` out onto the runner needs a secret in a public repo, and even if that were acceptable it would leave every external clone of the public repo unable to build its own tests. A public repo's test suite requiring a private repo to compile is the bug; moving the compile-time source of truth in-repo fixes the direction rather than papering over the symptom. Independently, this account's token carries `repo` but not `workflow`, so a workflow-side fix was not landable here at all — and none is needed.

Vendoring normally trades one problem for drift, and that is handled rather than waved at. `shared-context` stays canonical; the new `interop_fixtures_match_shared_context_canonical` asserts byte-equality against it as a **runtime** read, which is the correct call — `include_str!`-ing the sibling again to check the copy would reinstate the exact defect. Absent sibling no-ops (CI, outside clones); present sibling fails loudly. "Where observable" is every fleet machine, so the check runs where drift would actually be introduced.

The guard is demonstrated non-vacuous rather than assumed: appending a line to the canonical `minimal.yaml` turns it red with the DRIFTED message and the path, restoring turns it green. That matters — a skip-when-absent test is otherwise indistinguishable from a test that checks nothing.

Gated in both configurations: sibling absent → 201 passed (same tree would not compile before), sibling present → 283 passed / 0 failed across `hub-lib` + `hub-daemon`, +1 over the 282 baseline which is exactly the new test. Publication check: the fixtures are generic law-schema YAML and the schema is already public in `web4-standard/core-spec/hub-law-schema.md`, so nothing private is disclosed by the move. Content is unchanged, so the open `hestia-prd-review` discussion of dead gate fields in `full-featured.yaml` is undisturbed.

Confined to `#[cfg(test)] mod tests` (law.rs:444) plus new fixture files, so the release binary is byte-unaffected and no daemon restart is implied — which matters here, since the post-#361 vault means a restart drops the hub to a LOCKED 503 only dp can ignite.

Self-authored and self-reviewed in one cycle, stated plainly rather than papered over; the diff is small and fully checkable from the PR body. Holding merge until the PR's own `cargo test (hub)` job reports green — that job going from never-green to green is the claim, and it should be observed rather than predicted.

PR #598, opened 2026-07-30
fix(hub): the A2 receipt did not sign the string it transmits — signing_bytes v3

dp-web4:
`signing_bytes` v2 fed `DateTime::to_rfc3339()` into the canonical bytes
(`…302191724+00:00`) while serialising chrono's serde default onto the wire
(`…302191724Z`). The offset suffix was the entire difference, and it meant the
signed bytes were **not reconstructable from a received receipt** — the one thing
a portable A2 receipt exists to make possible. `tools/verify_assurance_receipt.py`
had to carry a `Z` → `+00:00` shim and report which spelling had carried the
verification.

v3 makes one function, `canonical_timestamp()`, produce both the transmitted and
the signed form: RFC3339, `Z`, and a fixed nine fractional digits. It is wired in
as `#[serde(with = "canonical_ts")]` on every timestamp `signing_bytes` covers, so
"sign what you send" holds by construction rather than by two definitions
agreeing by luck.

`SecondsFormat::Nanos`, not `Millis`: both are fixed-width, but Nanos is also
lossless at `DateTime<Utc>`'s own resolution, so canonicalizing costs no precision
and needs no truncation — a receipt round-trips through JSON as the same value and
still verifies. Fixed width is the second half of the fix: `AutoSi` varied the
fraction with the value (0, 3 or 9 digits), so a whole-second `valid_until` and a
live-clock `issued_at` were the same field in two shapes.

Verified, both directions:

- `signed_bytes_are_reconstructable_from_the_wire_alone` rebuilds the canonical
  bytes from parsed JSON — the way the Python verifier does — and checks the
  signature against them. Proven non-vacuous: restoring the v2 timestamp spelling
  in `signing_bytes` alone turns it RED, while the pre-existing
  `assurance_receipt_is_portably_verifiable_without_hestia` stays GREEN. That is
  why the defect survived — the old suite signed and verified through the same
  function, so it could not see the wire.
- `canonical_timestamp_is_fixed_width_z_suffixed_and_lossless` pins width, suffix
  and exact round-trip across whole-second, one-nanosecond and millisecond values.
- Cross-implementation: `verify_assurance_receipt.py --self-test` verifies the
  Rust-signed golden fixture from its verbatim wire strings, and now also rejects
  the v2 offset form and asserts the canonical shape on the committed bytes.
- `cargo test -p hub-lib -p hub-daemon`: 285 passed / 0 failed (283 before; +2 new
  tests), in a worktree with **no sibling `shared-context`** — the CI-equivalent
  configuration. `cargo build --release`: exit 0.

The shim in the Python verifier is removed rather than left to go quiet. After v3
its only remaining effect would be to let a future re-divergence verify with a
warning instead of failing — the regression detector becoming the thing that hides
the regression.

**No v2 receipt was ever issued, and that is measured, not inherited from #592's
"nothing is stranded" escape.** The receipt primitive landed 2026-07-29 16:06 PT
(d5bd10b); the live daemon has been running since 2026-07-27 20:22 PT. The running
image contains **zero** occurrences of the `assurance-receipt` domain string
(`strings` over `/proc/<pid>/exe`, with `query_hub` at 7 occurrences as the
positive control — the current on-disk build has both). The live tool surface
exposes only `query_hub`. Nothing is stranded because nothing was ever emitted; a
row-level ledger check was not available, since the society DB is encrypted
post-#361 and carries no SQLite header.

Fixture regenerated with the same fixed secret and the same UUIDs/nonce/roster
hash, so its diff is exactly the timestamp form plus the signature — and the
unchanged `.pubkey` and `hub_signer_key_id` are the evidence the secret really was
unchanged. Renamed v2 → v3 so a stale-tag receipt fails loudly.

Closes the `signing_bytes` v3 item carried forward since the 2026-07-30 01:00 hub
supervisor session.

---


dp-web4 (commented):
APPROVED: closes the `signing_bytes` v3 item carried forward since the 2026-07-30 01:00 hub supervisor session. Reviewed as HUB-track maintainer (`hub/**` per `hub/docs/MAINTAINER.md`). Self-authored and self-reviewed in one cycle — stated plainly rather than papered over; the diff is 5 files and fully checkable from the PR body.

**Scope** — matches the description exactly. `hub/hub-lib/src/constellation.rs` plus the standalone verifier, its golden fixture and that fixture's README. No other crate references `AssuranceReceipt` (grepped), so the blast radius is hub-local. No scope creep: the `web4-trust-core` missing-lockfile finding from the 04:00 session is still left for its owner, untouched here.

**The fix is right, and it is structural.** The defect was two definitions of one string. Putting `canonical_timestamp()` behind `#[serde(with = "canonical_ts")]` on the three covered timestamps means the wire form and the signed form cannot drift without someone deleting the attribute — the property is enforced by construction, not by a comment asking two functions to agree. That is the correct shape for a canonical-bytes bug.

**`Nanos` over `Millis` is the right call and the reasoning is recorded in the code.** Both are fixed-width; only `Nanos` is lossless at `DateTime<Utc>`'s resolution, so canonicalizing needs no truncation and the stored instant equals the transmitted one. `Millis` would have silently rounded the value the signature covers — a worse trade than the problem being fixed.

**Tests cover the new code, and the key one is proven non-vacuous.** `signed_bytes_are_reconstructable_from_the_wire_alone` reproduces a standalone verifier's arithmetic against parsed JSON and checks the signature over the result. Restoring the v2 spelling in `signing_bytes` alone turns it RED while the pre-existing `assurance_receipt_is_portably_verifiable_without_hestia` stays GREEN — which is the honest diagnosis of why this shipped: the old test signed and verified through the same function, so no amount of tampering coverage could see the wire. Verified both directions this cycle rather than asserted.

**Removing the Python shim is correct, not scope creep.** A fallback that survives the fix converts a future re-divergence from a failure into a warning. The self-test now pins the canonical shape on the committed bytes and rejects the v2 offset form outright, so the fixture keeps its cross-implementation job.

**The stranded-receipt claim is measured, which is the specific thing #592 was called out for inheriting.** Two independent proofs: the primitive landed 2026-07-29 16:06 PT (d5bd10b) versus a daemon running since 2026-07-27 20:22 PT, and zero occurrences of the `assurance-receipt` domain string in the running image via `/proc/<pid>/exe` with `query_hub` at 7 as the positive control. The live tool surface exposing only `query_hub` agrees. The one check NOT available is stated rather than glossed: the society DB is encrypted post-#361 and carries no SQLite header, so no row-level confirmation. Two proofs plus a named gap is the right standard.

**Gate.** `cargo test -p hub-lib -p hub-daemon` → 285 passed / 0 failed (283 before; +2 new), run in a worktree with **no sibling `shared-context`** — the CI-equivalent configuration, which is only meaningful because #597 made it so. `cargo build --release` exit 0, so the daemon builds and this is workspace-green in both directions, not just crate-green. Merging only after this PR's own `cargo test (hub)` job reports green.

**Deploy note:** this does NOT go live on merge. The running daemon is the 2026-07-27 build and post-#361 a restart drops the hub to a LOCKED 503 only dp can ignite, so the binary stays staged dark alongside #595. No live behavior changes today, and no receipt consumer exists to break.

PR #601, opened 2026-07-30
fix(hub): the constellation attestation signs a string it does not transmit — receiver-first v2

dp-web4:
#598 established the "sign what you send" property for `AssuranceReceipt` and
built the mechanism for it (`canonical_timestamp` + the `canonical_ts` serde
shim). It stopped at the receipt. `ConstellationAttestation` — the surface that
FEEDS the receipt, and the one with a live cross-repo counterpart in hestia —
was left on the pre-#598 pattern, 40 lines away in the same file.

## Cause

`signing_payload` covers `issued_at.to_rfc3339()`, which chrono spells
`…+00:00`. The JSON that actually crosses the wire (`rest.rs` present_constellation
deserializes it) carries chrono's serde default, `…Z`. Different bytes, therefore
a different SHA-256: on the shipped test vector the same inputs hash to
`a30b8d41…` under the signed spelling and `7f051de2…` under the transmitted one.

It works today only because both peers rebuild the payload from the *parsed*
`DateTime<Utc>`, never from the string they received. The divergence is invisible
until someone verifies from the wire alone — which is precisely what A2 portable
evidence is for. `AutoSi` compounds it: the fractional width varies with the
value, so any hop that re-serializes at lower precision (a millisecond-resolution
JS relying party, a microsecond DB column) silently changes the signed bytes,
intermittently and only for some timestamps.

## Fix — receiver first, senders last (#595 pattern)

- `signing_payload_v2`: same field order, tag bumped to
  `web4:constellation-attest:v2:`, `issued_at` spelled by `canonical_timestamp`
  — the same function the serde shim writes to the wire.
- `verify_enrolled` accepts v2, still accepts v1. A strict superset of today's
  behaviour; nothing that verified before stops verifying. The version the OWNER
  signed fixes the payload for the device signatures, so a co-sign cannot cross
  versions (tested — it degrades the tier via rule 4's silent drop, it does not
  inflate it).
- The tag bump is what makes accepting both safe: v1 and v2 bytes cannot collide,
  so this is not cross-protocol confusion. A forger still needs a signature from
  the pinned owner key under one spelling or the other.
- `#[serde(with = "canonical_ts")]` on the attestation's `issued_at`, so the
  transmitted spelling is the signed one. The shim's deserializer accepts any
  valid RFC3339, so v1 members parse unchanged.

Hestia still emits v1 and keeps verifying; nothing is required of it to keep
working. Its half of the migration is posted to the forum.

## The module doc was also overclaiming

It said `test_vector_payload_hash` "is the mechanical anchor that keeps the two
repos honest." It is not: hestia contains no counterpart — neither the vector nor
the string `test-vector-nonce` appears anywhere in that repo — so drift in
hestia's half fails nothing here. The two constructions ARE identical today
(verified). What is missing is the gate, not the agreement. Comment corrected to
say what the test actually does.

Both vectors were re-derived from the wire memo's field order by a third
implementation (Python, no chrono, no `web4-core`) before being pinned. That
matters here: hub-lib and hestia resolve `web4-core` from the same path, so a
shared-crate bug would otherwise agree with itself.

## Verification

- `cargo test -p hub-lib -p hub-daemon` → **289 passed, 0 failed**.
- `cargo build --release` → green.
- New: `test_vector_payload_hash_v2`,
  `attestation_v2_bytes_are_reconstructable_from_the_wire_alone`,
  `verify_enrolled_accepts_both_payload_versions`,
  `a_device_cosign_cannot_cross_payload_versions`.

surface: ConstellationAttestation::verify_enrolled   act: derive an MFA assurance tier bound to a pair_id
S: high/reversible [construct: TierBinding valid_until — expiry re-challenges, never silently extends]
R: n/a [construct: no reachability clause — channel reachability is never the evidence here]
W: pass [construct: verify_enrolled — owner sig against the PINNED resolver key; device facts from EnrolledDeviceSet, never from the presented attestation]
O: pass [construct: freshness + ForeignOwnerKey + signature checks all precede the returned tier; the function has no side effects]
A: pass [construct: caller commits the derived tier with AssuranceReceipt::signing_bytes, which covers the evidence]
V: n/a [construct: reversible and time-bounded — no irreversible act on this path]
verdict: PASS — accepting an additional signed spelling of the same content changes no clause; the cross-version splice test is the guard.

Machine: HUB
AI-Instance: claude-opus-4-8 (Claude Code)
Human-Supervised: no

Co-Authored-By: Claude Opus 4.8 <noreply@anthropic.com>


dp-web4 (commented):
APPROVED — hub-track, merging.

**Scope**: one file (`hub/hub-lib/src/constellation.rs`), verified against the staged diff before commit.

**Why this is correct, not just green.** The defect is latent, not live: v1 verifies today because both peers rebuild the payload from the parsed `DateTime<Utc>` rather than from the transmitted string. So the bar for this change is that it must close the portability gap *without* changing what verifies. It does — accepting v2-or-v1 is a strict superset of current behaviour, and the domain-tag bump makes the two payload spaces disjoint, so it is not cross-protocol confusion.

The one real risk in a dual-accept verifier is a splice: owner signs one version, a device co-signs the other. Resolved by construction — the owner's version fixes the payload for the device signatures, rather than each signature resolving independently — and pinned by `a_device_cosign_cannot_cross_payload_versions`, which asserts the mismatched hardware co-sign *degrades* the tier to single_device (rule 4's silent drop) instead of inflating it.

**Evidence checked, not assumed**: both test vectors were re-derived from the wire memo's field order in a third implementation before being pinned here. hub-lib and hestia resolve `web4-core` from the same path, so a shared-crate bug would agree with itself across both repos — comparing them to each other proves nothing on its own.

**CI**: `hub`, `web4-core`, `web4-policy` all SUCCESS. `web4-trust-core` FAILURE is the pre-existing gitignored-lockfile defect that fails on every PR against this base and is fixed by #600 (unmerged, another track's) — unrelated to this diff, which touches no trust-core path. Locally: `cargo test -p hub-lib -p hub-daemon` 289 passed / 0 failed, `cargo build --release` green.

**Standing-blocks interlock (Step 0)**: run unpiped on this PR — CLEAR, exit 0.

**Left for hestia** (posted to legion, nothing breaks meanwhile): mirror the test vector, then emit v2 at its own pace. v1 acceptance retires only after that.

PR #602, opened 2026-07-30
test(hub): the ledger's on-disk format includes a timestamp spelling no type declares

dp-web4:
## Why this exists

Last cycle's finding (#601) was that *a fix which establishes a named property is not a fix which applies it* — #598 built `canonical_timestamp` for `AssuranceReceipt` and left `ConstellationAttestation` 40 lines away on the old pattern. I committed to sweeping for the other places a named property should hold. This is that sweep.

`hub-lib` has three signing surfaces: `constellation.rs` (fixed by #598/#601), `envelope.rs` (canonicalises explicitly, interop-locked with hestia), and `ledger.rs::signing_payload` — the oldest, and the one behind `hub verify-ledger`.

## What was measured

Not inferred from the chrono docs — run against the pinned version:

| wire spelling | re-serialises to | |
|---|---|---|
| `…T09:00:00Z` | `…T09:00:00Z` | stable |
| `…T09:00:00.1Z` | `…T09:00:00.100Z` | **differs** |
| `…T09:00:00.000Z` | `…T09:00:00Z` | **differs** |
| `…T09:00:00+00:00` | `…T09:00:00Z` | **differs** |
| `…T02:00:00-07:00` | `…T09:00:00Z` | **differs** |

6 of 9 valid RFC3339 spellings of one instant do not round-trip byte-identically. The ledger signs `serde_json::to_string(&entry)`, so chrono's `to_rfc3339_opts(AutoSi, true)` — `Z`-suffixed, fractional width 0/3/6/9 chosen by the value — is part of the live chain's **on-disk format**. Every existing signature and every `entry_hash` was computed over it. No type declares it and no comment mentioned it; the doc claimed determinism rests on "the serde struct field order is fixed", which is half the story.

## Why this is a pin and not a fix

The attestation genuinely signed a string it did not transmit, across a trust boundary. The ledger does not have that problem:

- Both sign and verify re-derive the bytes from the *parsed* entry, never from stored text.
- No `LedgerEntry` is ever ingested from HTTP — every entry is built locally and read back from a store this code wrote (checked all deserialization sites: `store.rs`, `dynamodb_store.rs`, tests).
- So re-spelling a persisted timestamp cannot invalidate an intact chain. That is a **safety property worth keeping**, not a bug.

Adopting `canonical_timestamp` here would change every entry's signing bytes and every `entry_hash` — a migration of the live chain. That needs a ruling, not a drive-by merge from the maintainer track. Raised separately; **no bytes change in this PR.**

What is left is a real, latent risk with a silent failure mode: a chrono bump that alters `AutoSi` breaks verification of every historical entry, and the symptom an operator sees is *"the ledger has been tampered with."*

## The change

- `ledger_timestamp_wire_spelling_is_pinned` — exact chrono output for four fractional widths, plus the collapse rule for equivalent spellings. A format change now fails **at upgrade time**, not at audit time.
- `verification_re_derives_so_it_is_spelling_independent` — rewrites a persisted timestamp to an equivalent spelling (chain still verifies), then moves the instant (chain breaks). The second half is there so the first cannot pass vacuously.
- `signing_payload` doc: states both determinism dependencies, and the normalisation rule a non-Rust auditor needs — it cannot hash the JSON it read.

```
surface: none — tests + docs only; no code path, no consequential act.
verdict: n/a (no surface created or changed)
```

## Verification

`cargo test -p hub-lib -p hub-daemon` → **291 passed, 0 failed** (was 289; +2). Built in a throwaway worktree; the shared checkout dp's interactive session uses was not touched.


dp-web4 (commented):
APPROVED: this is the sweep I committed to after #601, and it found the third signing surface in hub-lib — `ledger.rs::signing_payload`, the oldest one and the one behind `verify-ledger`.

Reviewed against the hub PRD, the RWOA norm, and the merge criteria:

- **Scope matches description.** One file, tests + docs. No code path changes, no signed bytes change, no schema change. `git show --stat` is `ledger.rs` alone.
- **The finding is measured, not inferred.** 6/9 RFC3339 spellings shown non-round-tripping against the pinned chrono, with the probe's output reproduced in the PR body. The claim that no `LedgerEntry` is ingested from HTTP is backed by an enumeration of every deserialization site (`store.rs`, `dynamodb_store.rs`, tests) rather than by assertion.
- **The restraint is the right call, and is the part I looked at hardest.** The tempting move was to apply `canonical_timestamp` here for symmetry with #598/#601. That would rewrite every `entry_hash` in the live chain — #601's receiver-first dual-accept does not transfer, because a hash chain has one `prev_hash` per link. Correctly routed to a forum ruling instead of decided by the maintainer track. A PR that changed those bytes would have been rejected on this same review.
- **Tests cover the new claim and cannot pass vacuously.** `verification_re_derives_so_it_is_spelling_independent` re-spells a persisted timestamp (chain verifies), then moves the instant (chain breaks). The second half is what makes the first half evidence. `ledger_timestamp_wire_spelling_is_pinned` covers all four AutoSi fractional widths plus the collapse rule.
- **The pin fails at the right time.** A chrono bump that alters `AutoSi` currently surfaces as whole-chain verification failure during an audit, reading as "the ledger has been tampered with." It now fails in CI at upgrade time.
- **No scope creep.** `envelope.rs`'s parse-then-re-serialise is flagged as unmeasured and explicitly claims nothing — the right way to leave a lead.

RWOA: test-and-docs only, creates and changes no surface, so the gate is n/a and the PR says so rather than fabricating a PASS block.

Verification reproduced locally in a throwaway worktree: `cargo test -p hub-lib -p hub-daemon` → 291 passed, 0 failed (289 before; +2). CI: hub, web4-core, web4-policy all SUCCESS. `web4-trust-core` FAILURE is the pre-existing gitignored-lockfile defect (#600, another track's) that fails on every PR against this base and touches no path in this diff.

Step 0 standing-blocks interlock: CLEAR (rc=0, read unpiped).

PR #603, opened 2026-07-30
test(hub): the envelope's interop lock-in test cannot fail under the one condition it guards

dp-web4:
`interop_with_hestia_signing_algorithm` exists to catch the hub and hestia
disagreeing on signing bytes. Its payload is `{"a":1,"b":"hello"}` — insertion
order already equals sorted order — so it keeps passing under serde_json's
`preserve_order`, which is precisely the condition that breaks the interop.
Measured, not argued: enabling the feature in the workspace leaves that test
green and fails the widened battery added here.

Three pins, no behaviour change:

- `canonical_form_equals_hestias_to_string` — the equality over payload shapes
  where two independently-written serializers could actually diverge: escapes,
  control chars, non-BMP, `u64::MAX`/`i64::MIN`/`-0.0`, `1e300`/`1e-300`, empty
  and non-ASCII keys, empty containers, unsorted insertion order.
- `sender_key_order_is_a_requirement_the_hub_cannot_see` — the hub verifies
  against sorted keys, so a sender that serializes in any other order signs
  different bytes and gets a bare `BadSignature` naming neither key order nor
  payload. Hestia satisfies the requirement by construction (`serde_json::Map`
  is a `BTreeMap`), not by intent; a `plugin-sdk/python` or
  `plugin-sdk/typescript` client would have to sort deliberately. The
  requirement was written down nowhere.
- `wire_spelling_is_malleable_and_that_is_the_safe_case` — the other side of
  re-deriving from the parsed value: an intermediary can re-order keys or
  re-spell a float and the envelope still verifies. Malleability, not forgery,
  and safe only because nothing downstream of `verify_envelope` reads the
  received text — the same property that makes `ledger::signing_payload` safe.
  A consumer needing the signer's exact bytes must carry them.

`signing_bytes` docs corrected: the canonicalizer removes the *hub's*
dependence on its own feature flags; it cannot remove the protocol's dependence
on the sender's, because the hub never sees the sender's bytes.

surface: none — tests + docs only; no call site, signature, or wire format changes.
S: n/a (no consequential act introduced or altered)
R: n/a  W: n/a  O: n/a  A: n/a  V: n/a
verdict: PASS (no surface)

cargo test -p hub-lib -p hub-daemon → 294 passed, 0 failed (was 291).

Machine: HUB
Human-Supervised: no


dp-web4 (commented):
APPROVED — hub-track, tests + docs only, no behaviour change.

Reviewed against the hub PRD's interop invariant (hub/hestia must agree on `nonce ++ canonical(payload)`) and against last cycle's carried item #2, which is the cause of this PR: `envelope.rs`'s parse-then-re-serialise was unmeasured.

What the review checked:

1. **The claimed defect reproduces.** Enabling `serde_json/preserve_order` in the workspace leaves `interop_with_hestia_signing_algorithm` GREEN and fails the new battery. That test's payload `{"a":1,"b":"hello"}` has insertion order == sorted order, so it is vacuous under exactly the condition it exists to guard. Verified by running it both ways, then reverting the Cargo.toml probe (lockfile confirmed unchanged in the diff: `git diff --stat` = envelope.rs only, 207 insertions).

2. **No behaviour change.** The diff adds a doc block and three `#[cfg(test)]` functions plus a comment on an existing test. `serialize_canonical`, `canonical_signing_bytes`, `signing_bytes`, `verify_envelope` and `build_envelope` are byte-identical.

3. **Neither new assertion is vacuous.** `sender_key_order_…` proves order alone is the difference by re-signing the same payload sorted and asserting it verifies. `wire_spelling_is_malleable_…` follows the malleability assertion with a value-change that must still be rejected — without that half, a verifier that checked nothing would pass.

4. **Scope.** No remedy applied. The right call: changing the canonicalizer would break the live hestia interop, and order-independence is the property that makes the parsed value the authenticated object. Same discipline as #602 — measure the surface, pin the assumption, do not assume the fix travels.

5. **Novelty stated honestly.** `canonical_signing_is_key_order_independent` and `interop_with_hestia_signing_algorithm` already existed; the PR body says so and the widening is the contribution, not the coverage as a whole.

`cargo test -p hub-lib -p hub-daemon` → 294 passed, 0 failed (291 before; +3).

Step 0 standing-blocks interlock: CLEAR (rc=0, read unpiped), head `3acd27ff`.

PR #604, opened 2026-07-30
doc+test(hub): the envelope pins a key-order requirement on senders and not a number-spelling one

dp-web4:
## What

#603 wrote down that the hub verifies against **sorted keys** and that a sender in another language has to sort deliberately. That is the requirement a Rust author thinks of. It is not the only one.

The sweep discipline from #602 (a named property is not applied until you look for the other places it should hold) applies here one level down. The same "the hub never sees the sender's bytes" property that makes key order a silent requirement makes **number spelling** one too, with the identical symptom: a bare `BadSignature` that names neither the field nor the cause.

## Measured, not inferred

serde_json 1.0.150, CPython 3.12, V8:

| given | hub canonical | CPython `json.dumps` | JS `JSON.stringify` |
|---|---|---|---|
| `3.5e-4` | `0.00035` | `0.00035` | `0.00035` |
| `3.5e-5` | `0.000035` | **`3.5e-05`** | `0.000035` |
| `3.5e-6` | `3.5e-6` | **`3.5e-06`** | `3.5e-6` |
| `1e-10` | `1e-10` | `1e-10` | `1e-10` |
| `18446744073709551615` | exact | exact | exact |
| `18446744073709551616` | `1.8446744073709552e+19` | **exact** | **`18446744073709552000`** |

Two divergences:

1. **Small floats, CPython only.** Rust goes exponential below `1e-5` and never pads the exponent; CPython goes exponential below `1e-4` and pads to two digits. The divergent band is contiguous: **every non-zero float with magnitude in `[1e-9, 1e-4)`**. Below `1e-9` the exponent reaches two digits and the two agree again. That band is ordinary territory for a fraction — a T3/V3 weight or a per-unit rate lands in it without anyone choosing an odd value.
2. **Integers above `u64::MAX`, every language.** No `arbitrary_precision`, so serde_json falls back to `f64`. This one loses information *before* the signature check, so distinct large integers become the same value to the hub.

TypeScript is otherwise safe across the whole float range, including the integral-valued floats (`1e5` → `100000`) where it looks like it should differ — the sender's own spelling is what reaches the parser, so the two only have to agree on the round trip, not on the literal.

## Status: latent

hestia is Rust, so nothing is broken today. It is latent in exactly the way the sorted-key requirement was latent, and `plugin-sdk/python` is where it stops being latent.

## The test

Two halves, because they fail for different reasons:

1. **A spelling table** locking in serde_json's float `Display` at both decade boundaries. A bump that moved the exponential threshold, or started padding, would otherwise be invisible — both sides of `canonical_form_equals_hestias_to_string` move together, so it cannot catch this. Each entry also asserts the canonical form is a **fixed point**, so a spelling can't satisfy the table while being unreproducible by a sender that round-trips its own output.
2. **An end-to-end sign-and-reject**, anchoring the table to the symptom: a payload signed the CPython way is rejected, with a control proving the identical *value* in the hub's spelling verifies — so the rejection is about spelling and nothing else. Plus an anti-vacuity assert: if the divergence ever disappears the test fails loudly ("retired, not weakened") rather than passing for the wrong reason.

## Guard run against what it guards

Per #603's own lesson, three induced failures, all confirmed before commit:

- table entry re-spelled (`3.5e-6` → `3.5e-06`) → **FAILS**
- the divergence disappears → **FAILS** on the anti-vacuity assert, does not pass silently
- verifier accepts the foreign spelling → **FAILS**

## Gate

- No surface change — docs and one test. **RWOA n/a**: no path a caller can drive, no consequential act added or altered.
- `cargo test -p hub-lib -p hub-daemon` → **213 + 82 pass**
- `cargo build --release` → **clean** (workspace-green, not just crate-green)


dp-web4 (commented):
APPROVED: hub-track (single file, `hub/hub-lib/src/envelope.rs`), docs + one test, no surface change.

Reviewed against the hub PRD/sprint plan and the merge criteria:

- **Scope matches description.** 137 lines added, zero removed, zero behavior change. The canonicalizer is untouched — this documents and pins what it already does.
- **Integrates, not standalone drift.** It sits directly beside `sender_key_order_is_a_requirement_the_hub_cannot_see` and `canonical_form_equals_hestias_to_string` and is the second instance of the same requirement class those two established. This is the #602 sweep discipline applied to the surface #603 last touched, not a new direction.
- **Tests cover new code.** The new code *is* the test. More to the point, the table half covers a gap the existing interop lock structurally cannot: `canonical_form_equals_hestias_to_string` compares two serializers that would move *together* under a serde_json float-formatting change, so it can never catch a re-spelling. This one pins absolute expected strings.
- **The claims are measured.** Every number in the table and the PR body was produced by running serde_json 1.0.150, CPython 3.12 and V8 against the same literals, not read off documentation. The `[1e-9, 1e-4)` band was found by walking the decades, not asserted from the threshold docs.
- **The guard was run against what it guards.** Three induced failures, all confirmed failing before the commit — including the anti-vacuity case, where the test fails loudly if the divergence ever disappears rather than passing for the wrong reason. That is the precise defect #603 was opened to fix in the neighbouring test, and it would have been hollow to land its successor without the same check.
- **`cargo test -p hub-lib -p hub-daemon`: 213 + 82 pass. `cargo build --release`: clean** — workspace-green, not just crate-green.

Two things I want on the record as *not* claimed:

1. **This is latent, not a live break.** hestia is Rust; today every real sender agrees with the hub. The value is that the requirement is now written down before `plugin-sdk/python` exists, instead of after someone spends a day on an unexplained `BadSignature`.
2. **The large-integer half is the sharper one and is not fully addressed here.** Above `u64::MAX` the f64 fallback loses information *before* the signature check, so two distinct integers become one value to the hub. This PR documents and pins that; it does not fix it. A fix would mean `arbitrary_precision` or rejecting out-of-range integers at the envelope boundary, and both are behavior changes affecting every caller — not something to slip into a docs PR. Flagged in the session log as a follow-up needing a decision, not carried silently.

Merging as hub-track maintainer.

PR #605, opened 2026-07-30
fix(hub): a law amendment persisted before it was authorized — sign first, and put the store back if the witness fails

dp-web4:
Every law-amendment path wrote the new YAML into the store and *then* asked the
Sovereign to witness the `LawAmended` that authorizes it. When the signer
denies (`LockedSigner` — the live posture of every hub between boot and
ignition — or a remote signer's policy refusal) or is unreachable
(`HestiaCallbackSigner` transport error), the store is left holding an
amendment no ledger entry witnesses.

That is not a lost amendment. It is exactly the divergence
`law_integrity_write_gate` (HUB-001/H-009) refuses **every governed write** on,
hub-wide, until an operator re-witnesses by hand — and the in-memory law is
left un-updated, so the hub goes on evaluating the old law while the store
serves the new one. A transient signer error turns into a hub that stops
accepting governed acts.

Reproduced before repair: with the signer swapped to `LockedSigner`, the
operator admission-limits handler returned 401 and left `version: 1.0.2 /
repeat_limit: 9` on disk against a witnessed head of `1.0.1 / 5`.

RWOA's O clause already names this: "a denied act must leave state
bit-identical."

## What changed

- `witness_event` is split. `authorize_event` does the gate + build + sign and
  performs no side effect; `witness_event` is unchanged behaviourally (it is
  `authorize_event` + append).
- New `witness_law_amendment` orders it sign → write_law → append, so both
  modelled signer failures now fall entirely ahead of the store write. Used by
  `admin_set_admission_limits` and `hydrate_law_defaults` (which runs at every
  boot / post-ignition, and whose failure was logged as "hydration skipped"
  while the store had in fact already changed).
- `HubSession::set_law` — the CLI path, and the operator's only documented way
  out of a mismatch — gets the same ordering. A half-applied `set_law` deepens
  the state it was run to clear.

## What is compensated rather than prevented

`append_signed` failing *after* `write_law` succeeded (a local ledger/disk
fault) still opens the window. Both paths restore the prior law text and say so;
if the restore also fails, the failure is reported at `error!` / in the returned
error chain rather than swallowed. Compensation is not atomicity — a shared
transaction across the store write and the ledger append does not exist here —
but the window is now a disk failing writes in both directions, not a signer
being briefly unavailable. With no prior law the restore is a no-op and is
benign by construction: no witnessed `LawAmended` means the verdict is
"unverifiable", which does not gate writes.

## Guards, each run against the condition it guards

Three tests, each confirmed to FAIL before the repair (the third two via a
compiling mutant that restores the *new* text instead of the prior one), and
each written so a broken induction panics rather than passing vacuously:

- `a_denied_law_amendment_leaves_the_store_bit_identical` (hub-daemon) — the
  reachable case; asserts bit-identity, integrity still "ok", and that the
  in-memory law is untouched.
- `a_law_amendment_that_cannot_be_appended_restores_the_prior_law` (hub-daemon)
  — the residual window, induced by a read-only ledger file against a writable
  law file.
- `a_law_amendment_that_cannot_be_witnessed_restores_the_prior_law` (hub-lib) —
  the same for the CLI path. The failure half of the existing conjunction test.

`cargo test -p hub-lib -p hub-daemon`: 214 + 84 pass. `cargo build --release`
clean.

## Review-gate block

surface: law amendment (admin admission-limits, boot default hydration, `hub set-law`)
act: amend hub law (governed state + witnessed ledger act)
S: high/irreversible-ish [construct: LawAmended is append-only; the law gates every governed act]
R: pass [construct: require_loopback on the operator plane; CLI is operator-local]
W: pass [construct: authorize_event — Sovereign signature over the built entry]
O: pass [construct: authorize_event before write_law before append_signed]
A: pass [construct: append_signed commits entry+signature+hash-chain together]
V: n/a [construct: no new veto path; the existing law_integrity_write_gate is the fail-closed backstop]
verdict: PASS

Machine: HUB
Human-Supervised: no


dp-web4 (commented):
APPROVED: hub-track (hub/hub-daemon, hub/hub-lib). Reviewed against the hub PRD, the ratified RWOA gate, and the repo's own H-009/HUB-001 law-integrity design.

Scope matches the description — three law-amendment call sites and one split of an existing helper; no unrelated surface touched. `witness_event` keeps its behaviour exactly (it is now `authorize_event` + append), so nothing outside the law paths changes.

The defect is confirmed reachable, not theoretical: `LockedSigner` denies every sign by construction and is the posture of this very hub between boot and ignition, and `HestiaCallbackSigner` models a transport failure with a 503. Both landed after the store write. The consequence is the one the repo already treats as serious — `law_integrity_write_gate` refusing every governed write hub-wide — so the failure was upgrade-of-severity, not degrade.

The ordering is the right repair rather than a rollback bolted on: authorization now dominates the mutation (RWOA O), and the residual append-after-write window is compensated with the limits stated plainly in the rustdoc rather than papered over. The 'no prior law' branch is argued from `law_integrity_verdict` returning "unverifiable", which is correct — that path does not gate writes.

Tests are the part I checked hardest, per #603's lesson. All three induce the actual condition (signer swap; read-only ledger file against a writable law file — the file backend keeps them in separate files, so the split is real) and all three use `expect_err`, so a broken induction panics instead of passing vacuously. Confirmed independently that each fails without the repair; the two restore-branch guards were falsified with a compiling mutant that writes the new text back instead of the prior one, which is the mutation that would otherwise slip through.

Sweep is complete: `grep` for `write_law(` outside the store impls returns exactly the three production sites, and all three are covered. `set_law` being the documented recovery path for a mismatch makes its inclusion necessary, not optional.

`cargo test -p hub-lib -p hub-daemon` 214 + 84 pass, `cargo build --release` clean — run against the workspace, not just the crates.

PR #606, opened 2026-07-30
doc(hub): the two conjunction tools need opposite orders, and nothing said which is which

dp-web4:
#605 gave the law-amendment paths the order sign → write → append. `assign_role`
— the other tool flagged as a Conjunction tool in the same rustdoc category —
already carries a comment stating the opposite rule in general terms: "append the
authorizing witnessed record FIRST, then persist the derived society projection."

Both are correct. Read together they are a trap: someone tidying for consistency
would move the law paths to append-first and quietly break them, and the comment
that invited it reads like a rule rather than a case.

The discriminator is **re-derivability from the ledger**, and it was written down
nowhere:

- The society is a projection the ledger can rebuild, so a persist that fails
  after the append re-materializes. Append-first is right.
- The ledger holds only the law's sha256, never its text. Appending first and
  failing to write would witness an amendment whose content exists nowhere and
  no replay could recover. Sign → write → append is right.

Three doc sites, no behaviour change: the general claim in `assign_role` is
bounded to its own case and points at the counter-example; `HubSession::set_law`
and `witness_law_amendment` each say why they are not `assign_role`.

Found by sweeping the category the #605 defect belonged to rather than stopping
at the sites that had the bug — the same lesson as #602/#601.

`cargo test -p hub-lib -p hub-daemon`: 214 + 84 pass. `cargo build --release` clean.

Machine: HUB
Human-Supervised: no


dp-web4 (commented):
APPROVED: hub-track, docs-only (verified — `gh pr diff --name-only` is the two hub files, and the diff adds only comment/rustdoc lines; 214 + 84 tests unchanged and green, release build clean).

The finding is real and the framing is the right one. `assign_role`'s comment states its rule in unbounded terms ('append the authorizing witnessed record FIRST'), and #605 had just introduced a site where following that sentence would reintroduce the class of defect #605 removed. A correct comment that generalizes past its own case is a latent instruction to break the next thing, and that is worth a PR on its own.

The discriminator is right on the facts: the ledger's `LawAmended` carries `new_law_sha256` and no text, so law is genuinely not re-derivable from a replay, whereas the society is. Checked against `HubEvent::LawAmended` rather than taken from the PR body.

Bounding the claim at its source and pointing each site at the counter-example is the correct shape — it makes the rule discoverable from whichever of the three a reader lands on first, rather than only from the one that happens to be complete.

PR #607, opened 2026-07-30
test+doc(hub): the ledger's four integrity checks shared two tests, and two of the checks were never exercised at all

dp-web4:
verify_chain performs four checks — index, prev_hash, entry_hash,
signature — and the module tested them with two tamper tests, both of
which rewrote an entry's payload. A payload rewrite breaks the hash AND
the signature at once, so neither test could say which check caught it,
and both asserted the ambiguity directly:
`entry_hash mismatch || signature verification failed`.

Measured by mutation, 2026-07-30:

- neutering `entry.entry_hash != recomputed` → suite stays GREEN
- neutering `actor.verify_signature(..)` entirely → suite stays GREEN

The signature check is what makes the ledger unforgeable rather than
merely self-consistent, and it had no negative test in the module.

Each check now has a guard that fails when that check, and only that
check, is removed (every mutation below compiles — a non-compiling
mutant proves nothing):

| check      | the tamper only it catches            | mutation → |
|------------|---------------------------------------|------------|
| index      | deleted middle / swap / dropped Genesis| 3 fail     |
| prev_hash  | a fork grafted at its own index       | 1 fail     |
| entry_hash | a rewritten head hash, signature intact| 3 fail     |
| signature  | a forged signature, hash recomputed   | 1 fail     |

The two pre-existing disjunctions are tightened to name the check that
actually fires. `dropping_genesis_...` is renamed: measured, it is the
index check that catches it, not the Genesis anchor — the anchor's job
is to stop a grafted origin, which is a different attack.

## The completeness limit, pinned rather than left to an incident

A forward hash chain is anchored at Genesis and open at its head. Lopping
entries off the TAIL leaves every surviving index, link, hash and
signature correct, so verify_chain returns Ok and `hub verify-ledger`
prints "Ledger verified." over a history that has had acts removed. On
the SQLite backend the live chapter runs, that attack is one statement:
`DELETE FROM ledger_entries WHERE idx >= ?`.

This is a property of forward hash chains, not a defect here, and closing
it needs an anchor outside the store. The material for one is already
published unauthenticated — `query_hub` and `GET /v1/hubs/{id}/state`
both carry `head_hash` and the entry count — but nothing said that those
fields are load-bearing for anything.

So: `a_truncated_tail_still_verifies_because_the_chain_only_proves_consistency`
asserts the Ok deliberately (and asserts the published head moves, which
is what makes an external anchor sufficient), verify_chain's rustdoc
states the limit, and `hub verify-ledger` now prints what it proved and
what it did not — in its output, where the operator reading it is
looking, rather than in a doc comment they are not.

Also closes the file-backend tamper test's own stated gap ("SqliteBackend
would need its own tamper test"): both classes are induced against the
real SQLite backend, with an untampered baseline first so a broken
fixture cannot pass as a finding.

No behaviour change to verification. `cargo test -p hub-lib -p hub-daemon`:
222 + 84 pass (was 214 + 84). `cargo build --release` clean.

surface: hub verify-ledger / HubLedger::verify_chain   act: report ledger integrity to an auditor
S: high/reversible (a report, not a mutation — but relied on for irreversible decisions) [construct: run_verify_ledger]
R: n/a (local CLI, no remote reach)   W: n/a (read-only; no act authorized)
O: n/a (no side effect)   A: n/a (nothing appended)
V: n/a
verdict: PASS — the surface's defect was overclaim, not authority: it reported "verified"
for a state it cannot distinguish from a truncated one. Repaired by narrowing the claim
in the output, not by widening the check.

Machine: HUB
Human-Supervised: no

Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>


dp-web4 (commented):
APPROVED: hub-track, tests + docs only, no verification behaviour change.

Reviewed against the diff, not the title. Two substantive things carry it:

1. **The mutation evidence is the review.** Every claim in the body was produced by neutering the guarded line and re-running — and each mutant compiles, which is the trap this track walked into twice in the last two cycles. Neutering `entry.entry_hash != recomputed` and neutering `verify_signature` both left the suite green before this PR; after it, they fail 3 and 1 tests respectively. The four checks are now individually discriminated rather than jointly covered by a disjunction that accepts whichever survives.

2. **The truncation test asserts `Ok` on purpose, and that is the right call.** Teaching `verify_hub` to detect tail truncation would be a behaviour change to the live chapter's verify path — and that path cannot currently be exercised here at all (encrypted vault, no HUB_PASSPHRASE), so it could not be validated against the one chain that matters. Pinning the limit and narrowing the CLI's claim is the honest move; widening the check unvalidated would not be. The test fails loudly if someone later closes the gap, and points at the doc that has to change with it.

Scope is matched: no production logic touched except `run_verify_ledger`'s printed output, which is where the overclaim actually lived. `cargo test -p hub-lib -p hub-daemon` 222+84 pass (was 214+84), `cargo build --release` clean — both run pre-merge as the gate.

Residual, named not fixed: the completeness gap itself. The anchor material is published (`query_hub.head_hash` + `last_ledger_index`) but no fleet peer is recording it, so today nobody could actually detect a truncation. That is a fleet practice question, not a hub code change, and it goes to the forum.

PR #608, opened 2026-07-30
fix(hub): a member-declared skill is rendered unescaped into the operator page that shares an origin with the Sovereign-signing writes

dp-web4:
hub-daemon/src/admin.rs is 1174 lines and had ZERO tests. Its
`/admin/members` handler rendered each member's declared skills with

    m.skills.iter().cloned().collect::<Vec<_>>().join(", ")

— no escaping, while the name cell two lines above IS escaped.

`declare_skill` is a member-SELF act (rest.rs EnvelopeAction::DeclareSkill
explicitly permits `signer == subject`), the value is free text with no
charset or length validation anywhere in hub-lib or hub-daemon, and the
projection only lowercases it (state.rs `skill.to_lowercase()`), which
leaves `` intact.

Why this is an escalation and not a cosmetic bug: main.rs builds ONE
operator app —

    admin::router() .merge(admin::operator_router())
                    .merge(rest::admin_api_router())

so `/admin/members` and the Sovereign-signing `/admin/api/members/add`,
`/remove`, `/key`, `/admission-limits` are the same origin on
127.0.0.1:8772, and the default HUB_OPERATOR_AUTH is `loopback` — no
token, no CSRF barrier. Any citizen declares a skill; the next time the
operator opens the roster, that string runs in their browser inside the
reachability boundary that IS the authorization for those writes.

Swept the rest of the module for the same class rather than fixing the
one line (a fix that establishes a principle is not one that applies it):

- `event_summary` renders `RoleAssigned`'s role via `{:?}`, and
  `SocietyRole::Custom(String)` carries free text — `Debug` escapes `"`
  and `\` but not `<`/`>`/`&`. Sovereign-supplied (role assignment is
  Sovereign-only), so self-inflicted rather than an escalation, but the
  same defect on the same pages. Escaped. Audited every other `{:?}` in
  the function — DeviceType, RoleEventKind, PairRevocationKind,
  LctProvenance, EntityType, MetabolicState, SignerKind are all
  unit-variant-only and cannot carry a payload.
- `joins_page` truncated the pubkey with `&s[..len.min(12)]` — a BYTE
  slice, which panics mid-codepoint and would take the admission queue
  page down. Not reachable today (every writer validates the field as hex
  first: `request_citizenship` -> `hestia_sovereign_lct`,
  `admin_add_member` -> `admit_member`), so this is defence-in-depth on a
  page whose availability is the admission path. Now `chars().take(12)`.

Four guards, the module's first tests, driven through the REAL handlers
(`members`, `joins_page`) over a real chapter — not through an extracted
helper, so re-inlining the rendering cannot make them vacuous. Each was
mutation-verified to fail when its check and only its check is removed:

| mutation                                   | result  |
|--------------------------------------------|---------|
| skills escaping -> `.cloned()`             | 2 fail  |
| RoleAssigned `{:?}` unescaped              | 1 fail  |
| pubkey `chars().take` -> byte slice        | 1 fail  |
| member-name `html_escape` removed          | 1 fail  |

The char-boundary guard was VACUOUS on its first draft and the mutation
run is what caught it: an all-`é` payload puts a codepoint boundary
exactly at byte 12, so the byte-slice mutant survived. The payload now
leads with one ASCII byte to shift every `é` odd.

Considered and declined: adding `'` -> `&#39;` to `html_escape`. The only
attribute contexts in the module (`onclick="admit('{id}')"`) interpolate
Uuids, so no test can induce a failure — shipping it would be an
unexercised guard, which is the thing this change is about.

`rest.rs`: `mod channel_e2e_tests` and two helpers widened to
`pub(crate)` so admin's tests can build a real chapter. The module is
`#[cfg(test)]`; no production visibility changed.

cargo test -p hub-lib -p hub-daemon: 222 + 88 pass (was 222 + 84).

surface: /admin/members (operator-plane HTML render)
  act: none directly — it GATES member-authored script into the origin
       that can drive /admin/api/* Sovereign writes
S: high/irreversible [construct: admin_add_member, admin_remove_member —
   witnessed Sovereign acts, no undo]
R: pass [construct: admin.rs `members` skills cell html_escape] — the
   member-controlled bytes can no longer execute inside the loopback
   boundary. NOTE the residual, pre-existing and NOT introduced here:
   /admin/api/* still takes loopback reachability as its sole authority
   under the default HUB_OPERATOR_AUTH=loopback, which is an R FAIL on
   its own terms; this diff removes the member-reachable path INTO it,
   it does not repair it.
W: n/a — no new act, no identity decision added
O: n/a — render-only path, no side effects
A: n/a — no new record; the underlying MemberSkillDeclared is already
   witnessed and unchanged (the ledger keeps what was declared; escaping
   is a render-layer concern, correctly)
V: n/a
verdict: PASS (residual R FAIL recorded against /admin/api/*, pre-existing)

Machine: HUB
Human-Supervised: no


dp-web4 (commented):
APPROVED: hub-track (hub-daemon/src/{admin,rest}.rs), and the finding is real and reachable.

Verified independently of the PR body:
- `declare_skill` is a member-self act (rest.rs `EnvelopeAction::DeclareSkill` permits signer == subject), the field is unvalidated free text, and state.rs only lowercases it — so the injected string survives the projection intact.
- main.rs merges `admin::router` + `admin::operator_router` + `rest::admin_api_router` into ONE app on 127.0.0.1:8772, and `HUB_OPERATOR_AUTH` defaults to `loopback` — so the roster page and the Sovereign-signing writes genuinely share an origin with no token or CSRF barrier. The escalation is not hypothetical.
- Swept, not spot-fixed: every other `{:?}` in `event_summary` was checked against its enum definition (DeviceType, RoleEventKind, PairRevocationKind, LctProvenance, EntityType, MetabolicState, SignerKind — all unit-variant-only); `SocietyRole::Custom(String)` is the only one that can carry a payload, and it is fixed here.
- Guards run through the real handlers, not an extracted helper, so re-inlining the rendering cannot silently make them vacuous. All four mutation-verified to fail when their own check is removed — and the char-boundary guard was vacuous on first draft (an all-`é` payload lands a boundary exactly at byte 12); the mutation run caught it and the payload was corrected. That is the discipline working, and it is recorded in the commit rather than quietly fixed.
- The RWOA block is honest about what it does NOT repair: `/admin/api/*` still takes loopback reachability as its sole authority. This diff removes the member-reachable path into that boundary; it does not fix the boundary. Carried, not claimed.
- Declining the `'` -> `&#39;` addition to `html_escape` because no test can induce it is the correct call for a PR whose subject is unexercised guards.

Gate: `cargo test -p hub-lib -p hub-daemon` 222 + 88 pass (was 222 + 84), and `cargo build --release` clean, both run before this review. Standing-blocks interlock CLEAR (unpiped, exit=0).

PR #609, opened 2026-07-30
test(hub): the public/operator plane split is the daemon's network boundary and nothing tested it

dp-web4:
The hub serves two listeners. The public one is `--bind 0.0.0.0` (this hub is on
a tailnet; the archetype table also ships three public ones). The operator one is
127.0.0.1-only and carries the member roster, the skill graph, the `/admin/api/*`
admit/deny/remove/re-key API, and the Sovereign-signing MCP write tools.

Which plane a route lands on is decided in exactly one place — which `Router` its
`.route()` call sits in — and that decision had no test. It was enforced by
reading `run_serve`'s merge chain by eye. That has already failed once: the
2026-07-23 review found `/admin/ledger` and `/admin/pairs` on the public plane,
where `ledger_detail` renders full event payloads (names, pubkeys, skills,
profile values at every visibility tier) to an anonymous visitor. And the only
automated guard anywhere near this boundary — `mcp::require_loopback` — is
documented in `mcp.rs`, correctly, as defeated behind a same-host reverse proxy,
which is the public-deploy topology it exists to protect.

## What changed

Two extractions, no behaviour change: `public_plane_router()` and
`operator_plane_router()` lift the merge chains out of `run_serve` so they can be
built in a test. The layers `run_serve` puts over them (rate limit, lock gate,
operator auth) stay where they were and add or remove no routes.

## The guard

Three tests drive the *assembled routers* over a real chapter with
`tower::oneshot` — not a list of route strings, so moving a route between
`admin::router` / `admin::operator_router` / `mcp::{read,operator_read,write}_router`
is what the assertions see.

- `the_public_plane_serves_only_its_allowlist` — 25 operator-only paths must
  answer **404** on the public plane. 404 exactly, not "an error": a route that
  exists but rejects the caller answers 405/403/500, so this cannot be satisfied
  by a handler-level check that a proxy defeats. Only *unrouted* passes.
- `the_public_plane_still_serves_the_transparency_surface` — `/`, `/admin/roles`,
  `/admin/law`, `/admin/council`, `/tools`, `/tools/query_hub` must stay 200.
  Deleting the public plane must not be how the first test goes green.
- `the_operator_plane_carries_what_the_public_plane_refuses` — the same 25 paths
  must be routed on the operator plane, or "not public" would be satisfiable by
  the route existing nowhere at all.

`/admin/ledger/:index` is probed at index **0**, which exists on a fresh chapter.
`ledger_detail` answers its own 404 for a missing index, so a nonexistent index
would have made both directions of the table unfalsifiable — the first draft used
index 1 and the third test failed for that reason, not because a route was
missing.

## Mutation-measured

Each mutant compiles, and each is caught by the intended assertion alone:

| mutation | result |
|---|---|
| `/admin/members` moved from `operator_router` to `router` | allowlist test **fails**, other two pass |
| `mcp_write_router` re-merged onto the public plane | allowlist test **fails** on `POST /tools/add_member` |
| `/admin/roles` dropped from `router` | transparency test **fails** |
| `/admin/pairs` dropped from `operator_router` | operator-carries test **fails**, allowlist test still passes |

The last one is why the third test is not redundant with the first.

## What this does not do

It does not repair the operator plane's authorization — `HUB_OPERATOR_AUTH`
still defaults to `loopback`, and loopback launders through a same-host reverse
proxy. That is an R FAIL on the fleet's own terms and needs a ruling (making
`token` the default breaks every existing operator), not a patch. This change
makes the *placement* of a route structural; it does not make reachability into
authority.

```
surface: public_plane_router / operator_plane_router   act: route-set assembly (no act performed)
S: high/irreversible-in-effect [construct: a route placed on the public plane is anonymously reachable]
R: n/a [construct: no authorization decision is made here; layers unchanged]
W: n/a [construct: unchanged — handlers' own identity checks]
O: pass [construct: assembly is pure; no side effects]
A: pass [construct: no act recorded — no behaviour change]
V: n/a
verdict: PASS — extraction + guard only. The standing R FAIL on HUB_OPERATOR_AUTH=loopback
is carried forward unchanged and is NOT addressed here.
```

Tests: `cargo test -p hub-lib -p hub-daemon` → 222 + 91 pass (was 222 + 88).

Machine: HUB
Human-Supervised: no


dp-web4 (commented):
APPROVED: the plane split is the daemon's only network boundary and it was enforced by eye. Reviewed against the hub PRD (public plane = transparency only) and the 2026-07-23 review that moved `/admin/ledger` and `/admin/pairs` off the public plane — the class of regression this now catches structurally.

Verified rather than assumed:
- The two extracted functions reproduce `run_serve`'s merge chains exactly; the rate-limit, lock-gate and operator-auth layers stay in `run_serve` and add no routes, so the test's omission of them narrows the guard to the split rather than weakening it.
- The 404-exactly assertion is the right one: 405/403/500 all mean *routed*, so a handler-level check (which `mcp.rs` documents as proxy-defeatable) cannot satisfy it.
- All four mutants compile and each fails only its intended test, including the one that proves the third test is not redundant (`/admin/pairs` dropped from both planes → allowlist test still green, operator-carries test red).
- The index-0 detail: `ledger_detail` answers its own 404, so probing a nonexistent index made the table unfalsifiable in both directions. Caught in the run, not in review.
- `cargo test -p hub-lib -p hub-daemon` → 222 + 91 green; `cargo build --release` clean.

Scope is exactly the description: extraction + guard, no route moved, no behaviour changed. The standing R FAIL on `HUB_OPERATOR_AUTH=loopback` is named as untouched rather than quietly implied fixed.

PR #610, opened 2026-07-31
test(hub): the MCP write tools' governance gate had no test, and its ordering is the whole guarantee

dp-web4:
test(hub): the MCP write tools' governance gate had no test, and its ordering is the whole guarantee

`mcp.rs` carries 598 lines and one test — the loopback check. Everything the
Sovereign-signing write tools actually gate on was unverified: the council
threshold, the three PolicyEntity decisions, the HUB-001 law-integrity refusal,
and — the one that matters most — the *order* of the gate against the side
effect.

`assign_role` is the only MCP handler with a pre-append side effect. It builds a
role assignment into an in-memory society, calls `check_governance`, and only
then calls `write_society`. Those last two lines are RWOA's O clause in full: swap
them and a denied assignment persists, so the society reads as having a filled
Archivist role that no ledger entry attests. Nothing in the suite noticed the
difference. It does now.

Ten tests, each run against the regression it names (each mutation below was
applied to the production code and the named test observed to fail):

- gate moved after `write_society`      → denied-assignment test FAILS
- council threshold `m >= 2` → `m >= 3` → council-refusal test FAILS
- `Escalate` returns 403 instead of 202 → decision-mapping test FAILS
- `law_integrity_write_gate` call removed → mismatch test FAILS
- receipt returns `prev_hash`           → receipt test FAILS
- `role_lct_id` freshly minted          → allowed-assignment test FAILS
- `Warn` returns an error               → warn-proceeds test FAILS

Two things worth naming for a reader:

`Society.roles` is a hash map, so comparing `serde_json::to_string` output of two
identical societies reports a difference that isn't one. The first version of the
O-clause test failed for exactly that reason. The comparison is canonical now —
without it, the allow-path `assert_ne` was passing on key order rather than on
content, i.e. it was green and vacuous.

`ApiError` gains `#[derive(Debug)]`, matching `rest.rs`'s. That is the only
non-test change in the diff.

surface: MCP write tools (add_member/assign_role/record_event/declare_skill)
act: Sovereign-signed ledger append + society persist
S: high/irreversible [construct: append_signed_event — a witnessed chain entry]
R: n/a — no reachability rule changed (require_loopback and the operator-plane
   mount are unchanged; the plane split is #609's, tested there)
W: n/a — no identity or authority rule changed
O: pass [construct: check_governance before store.write_society in assign_role —
   now pinned by a_denied_role_assignment_leaves_the_persisted_society_bit_identical]
A: n/a — no record format changed
V: n/a
verdict: PASS — tests only; no surface, gate, or status code is altered by this
diff. Carried item 1 (loopback-as-authority on /admin/api/*) is untouched and
still open.

Machine: HUB
Human-Supervised: no


dp-web4 (commented):
APPROVED: hub-track, test-only plus one `#[derive(Debug)]` matching rest.rs's `ApiError`.

Reviewed against the hub PRD's governance-gate obligations and the RWOA O clause. Three things carried the approval:

1. **The gap was real and load-bearing.** `mcp.rs` had 1 test for 598 lines, and the untested surface was the Sovereign-signing write path — council mode, the three PolicyEntity decisions, HUB-001 law integrity, and the gate/side-effect ordering in `assign_role`. The ordering one is the most valuable: `check_governance` before `write_society` is the entire O-clause guarantee for the only MCP handler with a pre-append side effect, and it was held by nothing but the comment above it.

2. **Every guard was run against what it guards.** Seven mutations applied to production code, each with the named test observed to fail — including the two boundary cases that a weaker suite would miss (M=1 is *not* council mode; `Warn` proceeds rather than blocks). This is the standard `guards-must-be-run-against-what-they-guard` asks for and it was met, not asserted.

3. **The vacuous-green trap was caught and named.** `Society.roles` is a hash map; the first O-clause test failed on key ordering, which also meant the allow-path `assert_ne` was green on key order rather than content. Canonicalizing fixed both. A PR that had silently shipped the string comparison would have added a test that could never fail.

No scope creep: no route, status code, gate, or auth rule changes. Suite 222 (hub-lib) + 101 (hub-daemon, was 91) green; `cargo build --release` clean. Merging.

PR #611, opened 2026-07-31
test(hub): the plane split was pinned by route, not by address — the operator listener could follow --bind with every route still correctly placed

dp-web4:
#609 pinned *which router* each route is declared in. It did not pin *where each
router is served*, and those are the same boundary from two sides. The operator
plane — `/admin/api/*` admit/deny/remove/re-key plus the Sovereign-signing MCP
write tools — is 127.0.0.1-only by a string literal 170 lines below the routers,
and this hub runs `--bind 0.0.0.0` for its tailnet. A one-word edit making the
operator listener follow `bind` publishes the entire write plane to the network
with all 25 of #609's operator-only routes still exactly where that test wants
them, and all three of its tests still green.

Layer order had the same shape: `operator_auth_gate` is applied outermost so an
unauthorized caller is refused before the lock gate, and that ordering was held
by a comment.

## What changed

Three extractions, one behaviour change.

- `plane_addrs(bind, port, admin_port)` — both listen addresses, decided in one
  place. It takes `bind` and deliberately does not use it for the operator
  address, so the mutation is a failing test rather than a signature change the
  compiler catches instead.
- `operator_plane_app(rate_limited, gate_state, op_auth)` — the operator layer
  stack in the order `run_serve` applies it.
- `public_bind_is_loopback(&addr)` — the production-warning classifier, now read
  from the parsed address.

The behaviour change is that last one, and it is a repair. The classifier was
`matches!(bind, "127.0.0.1" | "localhost" | "::1" | "[::1]")`. Two of those four
arms are unreachable: `"localhost:8770"` and `"::1:8770"` are not `SocketAddr`s,
so `parse()` had already hard-failed 115 lines earlier with a bare
`AddrParseError` naming nothing. The matcher advertised two spellings the daemon
refuses to boot on, while the CLI help (`default 127.0.0.1 — local-only`) invites
exactly the first of them. Now: the parse error names the accepted forms, and
loopback-ness comes from `addr.ip().is_loopback()`, which also stops warning
spuriously on `127.0.0.2`.

Verified before writing the fix, not assumed:

```
  127.0.0.1:8770  ->  Ok      localhost:8770  ->  Err(AddrParseError(Socket))
      [::1]:8770  ->  Ok            ::1:8770  ->  Err(AddrParseError(Socket))
```

Both directions of the classifier are asserted, because only one is dangerous: a
network-reachable bind classified as loopback silently drops the "hardening is
NOT enforced" warning on a public hub. `0.0.0.0` and `[::]` are the binds an
`is_unspecified`-shaped mistake still gets right, so `100.65.206.122` and
`192.168.1.10` are what carry that test.

`admin_port == 0` now yields `None` rather than an address. It previously took
the `!= 0` branch to skip binding; stated as a type, port 0 cannot be mistaken
for "listen on an ephemeral port" — a Sovereign-signing listener that is neither
off nor findable.

## The locked fixture

Order is unobservable on an unlocked hub: the lock gate passes everything, so
both stacks answer identically. The test builds a real locked boot — a clear
`public-identity.json` plus `open_locked_shell`, which installs the
`LockedSigner` that `is_locked()` actually reads — and drives it from a
non-loopback peer via an injected `ConnectInfo`. Auth outermost answers 403;
lock-gate outermost answers 503, handing the hub's lock state to a caller the
next layer was about to refuse. No env mutation and no token file: the default
loopback mode makes a non-loopback peer the unauthorized caller.

## Mutation-measured

Each mutant compiles, and each is caught by the test that names it:

| mutation | test that failed |
|---|---|
| operator listener follows `--bind` | `the_operator_plane_stays_on_loopback_under_every_public_bind` |
| `admin_port` 0 binds an ephemeral port | `admin_port_zero_disables_the_operator_plane_rather_than_binding_an_ephemeral_port` |
| classifier reads `is_unspecified()` | `a_network_reachable_bind_is_never_classified_as_loopback` |
| public address pinned to loopback too | `the_public_plane_binds_exactly_what_it_was_told` |
| bare `AddrParseError`, no hint | `a_hostname_bind_is_refused_and_the_error_names_what_is_accepted` |
| lock gate layered outside operator auth | `an_unauthorized_operator_caller_is_refused_before_the_lock_gate` |
| lock gate dropped entirely | `an_authorized_caller_is_still_stopped_by_the_lock_gate` |

The first pass reported all seven as "did not compile". That was the harness:
`cargo test` prints `error: test failed, to rerun pass ...` on a *failing* test,
and the compile check read it before the failure check. A mutation harness that
cannot tell a caught mutant from a broken build reports the same word for both,
which is the failure mode it exists to prevent — recorded because the numbers
above are only worth what the harness that produced them is.

The last two rows are why the second layer-order test is not redundant: putting
auth outermost is otherwise satisfiable by deleting the lock gate.

## What this does not do

It does not repair the operator plane's authorization. `HUB_OPERATOR_AUTH` still
defaults to `loopback`, and loopback launders through a same-host reverse proxy —
the standing R FAIL, carried forward unchanged. This pins where the plane
listens, not what listening there proves.

```
surface: plane_addrs / operator_plane_app   act: listener + gate-stack construction (no act performed)
S: high/irreversible-in-effect [construct: an operator listener off loopback publishes the Sovereign-signing write plane]
R: n/a [construct: no authorization decision made here; gates unchanged]
W: n/a [construct: unchanged — operator_auth_gate + handler identity checks]
O: pass [construct: address + layer assembly is pure; binding happens after]
A: pass [construct: no act recorded — no gate behaviour change]
V: n/a
verdict: PASS — construction guard + one legibility repair (dead classifier arms,
unnamed parse error). The standing R FAIL on HUB_OPERATOR_AUTH=loopback is NOT addressed here.
```

Tests: `cargo test -p hub-lib -p hub-daemon` → 222 + 108 pass (was 222 + 101).
`cargo build --release` clean; the two remaining warnings are pre-existing.

Machine: HUB
Human-Supervised: no


dp-web4 (commented):
APPROVED: this closes the complement of #609 and I reviewed it as the hub-track maintainer.

**Scope.** One file, hub-track only (`hub/hub-daemon/src/main.rs`). Three extractions plus one behaviour change, all inside `run_serve`'s construction. No handler, no gate, no route moves.

**The finding is real and the framing is right.** #609 made route *placement* structural and explicitly left the listener addresses to the eye. That gap is exploitable in the direction that matters: `operator_plane_router` can be perfectly assembled and the write plane still published, because `format!("127.0.0.1:{}", admin_port)` sat 170 lines away from the router it serves. On a `--bind 0.0.0.0` hub — which is this one — that is the Sovereign-signing surface on the tailnet. Taking `bind` into `plane_addrs` and leaving it unused for the operator address is the correct call: it keeps the mutation a red test instead of a compile error, and the doc comment says why so the next reader does not "clean up" the unused parameter.

**The behaviour change is a repair, not scope creep.** The old `matches!` listed `localhost` and `::1` as loopback spellings that `parse()` rejects 115 lines earlier — verified here against the parser rather than asserted, which is the right order. Two dead arms advertising support the daemon refuses to boot on, while the CLI help invites `localhost`. The classifier only decides whether a warning prints, so the change is confined to the safe direction, and `127.0.0.2` now stops warning spuriously. `admin_port == 0` → `Option::None` is a genuine improvement: port 0 as a sentinel is one deleted `if` away from an ephemeral-port Sovereign listener.

**Mutation coverage is the reason I am merging this rather than asking for more.** Seven mutants, each compiling, each caught by the single test that names it — including both layer-order directions, which is what stops "auth outermost" from being satisfiable by deleting the lock gate. The locked fixture is the part that could have been faked and was not: order is genuinely unobservable on an unlocked hub, and the test builds a real `open_locked_shell` boot with an injected non-loopback `ConnectInfo` rather than asserting on the layer list.

**The harness self-report is the strongest thing in this PR.** The first run returned all seven mutants as "did not compile" because `cargo test` emits `error: test failed` for a red test. Reporting that, instead of quietly fixing the detector and printing seven clean rows, is exactly `guards-must-be-run-against-what-they-guard` applied to the guard's own instrument — a mutation harness that cannot distinguish a caught mutant from a broken build produces numbers that mean nothing, and it would have read as full coverage.

**Residual, correctly declared.** `HUB_OPERATOR_AUTH=loopback` remains the default and remains an R FAIL on the fleet's ratified terms; this PR pins where the plane listens, not what listening there proves. It does not claim otherwise. Carried forward unchanged.

Build gate per `build-daemon-before-merging-engine-prs`: `cargo test -p hub-lib -p hub-daemon` → 222 + 108 (was 222 + 101), `cargo build --release` clean, both warnings pre-existing and untouched.

PR #613, opened 2026-07-31
fix(hub): the lock gate allowlisted a suffix, and /admin/law is not the law route it meant

dp-web4:
fix(hub): the lock gate allowlisted a suffix, and /admin/law is not the law route it meant

`lock_gate` is the daemon's fail-closed boundary: while the hub is locked,
every request is refused 503 except a small tier-0 allowlist, "so no handler
runs against unpopulated state in a locked shell."

Three of its five entries matched by **suffix**. `ends_with("/law")` was
written for `GET /v1/hubs/:hub_id/law`, which serves the in-memory clear law
projection and touches no store. It also matched `GET /admin/law` — the
operator HTML page, declared on the **public** plane, whose handler calls
`open_store()` and reads the ledger.

A locked hub cannot open that store by construction: `run_serve` boots the
locked shell precisely *because* `open_hub_store_async` failed. So the one
`/admin/*` page the gate let through is the one that then fails, and it fails
as a 500 carrying `format!("{:#}", e)` from the store layer:

    500  <pre>hub state at /…/hub.db is encrypted but no HUB_PASSPHRASE is
         set — set the vault passphrase to open it</pre>

to an unauthenticated caller, on a `--bind 0.0.0.0` listener, while every
sibling page (`/admin/roles`, `/admin/council`, `/tools`) correctly answers
the fail-closed 503. Measured, not reasoned: see the fixture note below.

Not hypothetical for this hub. `web4-fleet/config.toml` carries no `[storage]`
section and its `hub.db` is SQLCipher-encrypted, so disk auto-detection
resolves sqlite, the store fails closed, and the daemon boots this exact
locked shell on every restart until an operator ignites it.

## The fix

`locked_tier0_allows(hub_id, path)` — a pure predicate matching **whole
paths**, with `:hub_id` resolved from the clear `public-identity.json` the
locked shell already booted on. Per named-property-prs-need-a-sweep this
retires the class, not the instance: `/unlock` and the OID4VCI issuer metadata
carried the same latent reach and are converted in the same change.

No tier-0 behaviour changes. The tier-2 M-of-N routes (`/unlock/challenge`,
`/unlock/attest`) stay refused while locked — they were outside the suffix too,
and they require an already-ignited hub — now pinned by a test rather than by
the accident that neither path *ends* in `/unlock`.

## Tests

hub-daemon 108 → 117. Six on the predicate (exhaustive over the route table,
no app needed), three driving the assembled public plane end-to-end through
the real middleware.

The daemon-level fixture is the part worth reading. The obvious one — a
`fresh_rest_state` hub plus `open_locked_shell` — is an **unreachable**
combination: that hub is file-backed, its store always opens, and a hub whose
store opens never becomes a locked shell. Under it `/admin/law` returns a
cheerful 200 and the bug is invisible. The fixture inits a sqlite hub,
encrypts the store, holds no key, and asserts `open_hub_store_async` is `Err`
*before* building the shell — so the precondition that makes the shell
reachable is part of the test rather than an assumption behind it.

## Mutation-measured

Seven mutants, each compiling, each caught by the test that names it:

| mutation | test that failed |
|---|---|
| `/law` back to a suffix | `the_operator_law_page_is_not_tier0` |
| `/unlock` back to a suffix | `a_tier0_spelling_in_the_last_segment_does_not_admit_a_route` |
| issuer metadata back to a suffix | `a_tier0_spelling_in_the_last_segment_does_not_admit_a_route` |
| gate disabled (allow all) | `the_operator_law_page_is_refused_by_the_lock_gate_not_by_its_handler` |
| gate refuses everything | `the_tier0_surface_is_still_served_while_locked` |
| landing page dropped | `the_five_tier0_paths_are_served_while_locked` |
| any hub's law, not this hub's | `another_hubs_law_path_is_not_this_hubs_tier0_surface` |

Carried item 17 from the 2026-07-30 18:01 session said the previous harness
printed one word for both terminal states and was one non-compiling mutant
away from certifying itself. The fix here is structural, not a better string
match: **compile and run are separate steps**, so a mutant that fails to build
is reported as `DID_NOT_COMPILE` by the step that built it and never reaches
scoring. All three terminal states fired during this run — `CAUGHT` above,
`DID_NOT_COMPILE` on a mutant my own shell escaping had corrupted, and
`SURVIVED` on a deliberately inert control mutant (a comment-text edit) run to
prove the harness can still say so.

## Accountability self-audit

surface: lock_gate   act: serve a route on a locked (un-ignited) hub
S: med/reversible [construct: read-only pages; the disclosure is the store's
   on-disk path + lock state, and no write reaches a handler either way]
R: pass [construct: locked_tier0_allows — whole-path allowlist; reachability is
   not the basis, the route's tier is]
W: n/a [no identity is asserted; the gate refuses before authority is consulted]
O: pass [construct: lock_gate runs as middleware before `next.run(req)`, so a
   refused path reaches no handler and touches no store]
A: n/a [refusals are not ledgered acts; nothing is committed]
V: n/a [reversible, non-consequential]
verdict: PASS

Machine: HUB
Human-Supervised: no


dp-web4 (commented):
APPROVED: hub-track, and the strongest kind of finding this track produces — a fail-closed guard that was open on exactly one path, verified by measurement rather than by reading.

**Scope/PRD.** `hub/hub-daemon/src/{rest,main}.rs` only. Sits squarely on the #609/#611 line of work (network boundary of the daemon), and closes carried item 16 from the 18:01 session, which named this suffix as the next candidate.

**The finding is real and I re-derived it independently of the PR body.** `ends_with("/law")` matches `/admin/law`, which `plane_split::the_public_plane_still_serves_the_transparency_surface` pins to the **public** plane — so it is anonymously reachable on this hub's `--bind 0.0.0.0`. Its handler opens the protected store (admin.rs `law()` → `s.open_store()`), which on a locked shell cannot open, and `AdminError: From<anyhow::Error>` maps that to `(500, format!("{:#}", e))`. Measured on the faithful fixture: `500 <pre>hub state at …/hub.db is encrypted but no HUB_PASSPHRASE is set…</pre>`, against 503 for `/admin/roles`, `/admin/council` and `/tools`.

**Live-relevant, not theoretical.** `web4-fleet/config.toml` has no `[storage]` section and `hub.db` carries a non-plaintext header, so `open_hub_store_with_key_async`'s disk auto-detection resolves sqlite and fails closed — this hub boots the locked shell on every restart. This is therefore reachable the moment the staged binary is ignited, which is the carried item that has been open longest.

**What earns the approval over a narrower patch:** the fixture. A `fresh_rest_state` + `open_locked_shell` hub is an unreachable state combination (file-backed store always opens; a hub whose store opens never becomes a locked shell) and under it `/admin/law` returns 200 — the guard would have been green against a condition the daemon cannot enter. Asserting `open_hub_store_async(...).is_err()` *before* constructing the shell puts the precondition inside the test. That is `guards-must-be-run-against-what-they-guard` applied to the fixture rather than to the assertion.

**Sweep, not instance** (`named-property-prs-need-a-sweep`): `/unlock` and the issuer metadata had identical latent reach and are converted in the same diff, with the tier-2 `/unlock/{challenge,attest}` exclusion now pinned by a test instead of resting on the accident that neither path ends in `/unlock`.

**Verification gate** (`build-daemon-before-merging-engine-prs`): `cargo test -p hub-lib -p hub-daemon` → 222 + 117 pass (was 222 + 108). Seven mutants each compiled and each caught by the named test; I re-ran the M7 row after correcting an escaping bug in my own harness rather than reporting its DID_NOT_COMPILE as a catch.

**Harness note.** Carried item 17 asked for two distinguishable outputs exercised before either is trusted. Splitting compile from run is the structural version of that, and all three states fired this run, including SURVIVED on an inert control mutant. That control is the row that makes the other seven mean something.

**Residual, named honestly:** `/admin/law` is now 503 on a locked hub rather than rendering. That is correct under the gate's own contract ("a locked hub is unlocked, not operated") and the page could not render anyway — but it is a behaviour change to a public transparency URL, not a pure test addition, and the PR says so.

PR #614, opened 2026-07-31
fix(hub): three public-plane error constructors rendered the anyhow chain to anonymous callers

dp-web4:
`ApiError::internal` (rest), `ApiError::internal` (mcp) and `AdminError`'s
`From<anyhow::Error>` all rendered `format!("{:#}", e)` straight into the
response body. All three are reachable on the public plane of a `--bind 0.0.0.0`
hub, so an unauthenticated caller receives the store's operational detail --
`hub.db`'s absolute path, sqlite's reason, the passphrase-missing sentence.

The mcp one is the widest: its blanket `impl<E: Into<anyhow::Error>> From<E>`
routes EVERY `?` in every MCP handler through it, and `mcp_read_router` is
public. `/admin/council` is the most direct: it opens the store inside a page
served on the anonymous listener.

Redacted at the source rather than by a response layer. The detail goes to the
daemon log with a per-occurrence correlation reference, and the caller gets a
fixed sentence carrying that reference -- the detail moves to the channel that
machine access already authenticates, it is not deleted.

A blanket `is_server_error()` scrub over the public plane was the obvious
alternative and is wrong: the curated 5xx bodies are load-bearing. `locked_error`'s
503 is the sentence that tells the fleet to ignite the hub, and a response-level
filter would have eaten it. Pinned by a test.

Named-property sweep in the same diff: the three `internal` constructors plus the
three hand-rolled `AdminError(INTERNAL_SERVER_ERROR, e.to_string())` sites in
`ledger_detail` and `council`.

Tests: hub-daemon 117 -> 122. The end-to-end case runs a real public route
(`read_state`) on an unlocked hub whose storage was removed underneath it, and
asserts the precondition that the failure was actually reached -- the encrypted-
store version of this state is already refused upstream by `lock_gate`, so it
would have proved nothing about this route.

Mutation-measured: 6 mutants CAUGHT (each constructor reverted, constant
reference, reference dropped, curated-503 scrubbed), control mutant SURVIVED.
The control earned its keep -- the first harness classified on `grep failed`,
which matches the "0 failed" success line, so it reported CAUGHT unconditionally
including for the control. Fixed to classify on cargo's exit status.

surface: public-plane error bodies   act: disclosure of hub storage detail
S: med/irreversible (a disclosed path cannot be un-disclosed) [construct: redact_internal doc]
R: n/a (changes what a surface discloses, not what it authorizes)
W: n/a   O: n/a (no side effect ordering changed)
A: pass [construct: tracing::error! with the same reference the caller is given]
V: n/a
verdict: PASS

Machine: HUB
Human-Supervised: no


dp-web4 (commented):
APPROVED: hub-track (all three files under `hub/hub-daemon/`).

**Scope matches description.** Three constructors and three hand-rolled sites, one shared helper, no behaviour change outside internal-error rendering. hub-lib untouched (222 unchanged).

**The defect is real and reachable, not latent.** `/admin/council` is declared in `admin::router`, which `main::public_plane_router` merges onto the anonymous listener, and its handler calls `s.open_store()` directly — the store's absolute path reaches an unauthenticated caller on this hub's `--bind 0.0.0.0`. The mcp instance is wider still: the blanket `impl<E: Into<anyhow::Error>> From<E>` puts every `?` in every MCP handler on this path, and `mcp_read_router` is public.

**Source-level cut is the right one, and the PR proves the alternative wrong rather than asserting it.** A blanket `is_server_error()` layer over the public plane would also redact `locked_error`'s 503 — the one 5xx a caller is *supposed* to read and act on, and the exact response the fleet sees during ignition. `a_curated_5xx_keeps_the_sentence_the_caller_must_act_on` pins that, so the rejected design cannot be reintroduced silently.

**The e2e fixture asserts its own reachability.** `the_public_state_route_does_not_name_the_store_it_failed_to_open` removes the hub directory under an *unlocked* hub and asserts `society()` errors before invoking the handler. The obvious encrypted-store fixture would have been unreachable for this route — `/state` is not tier-0, so `lock_gate` (#613) refuses it upstream — and would have proved nothing. That is the 19:00 lesson applied one level out.

**Diagnostics move rather than disappear.** Per-occurrence correlation reference in the body, same reference in `tracing::error!`. Pinned by `each_internal_error_carries_its_own_reference`, so a constant ref (which hides detail but makes the log line unfindable) fails.

**Mutation-measured, with a working control.** 6/6 CAUGHT; control SURVIVED. Worth recording that the first harness reported CAUGHT for the control too — it classified on `grep failed`, which matches the `0 failed` success line, so it fired on every run. Same self-certifying shape carried item 17 named; caught here only because the control was run. Re-classified on cargo's exit status, plus a `NO_TESTS_RAN` guard for a filter that matches nothing.

**Gate:** `cargo test -p hub-lib -p hub-daemon` → 222 + 122 pass (117 → 122). `cargo build --release` clean; 2 warnings, both pre-existing (`web4-trust-core` dead fn, `admin.rs` unused import). The ~20 `unused axum::Json` test warnings are pre-existing and all in the 7100+ range, far from this diff.

**Interlock:** CLEAR, run unpiped, head `cbf79676`.

**Noted, not blocking:** `web4/CLAUDE.md` mandates `gitnexus_impact` before editing a symbol. The GitNexus MCP server is not connected in this session, so that analysis was not run; the blast radius was established by reading the three routers and the call sites instead (50 `ApiError::internal` sites, all funnelling through one constructor).

PR #616, opened 2026-07-31
test(hub): the public-plane 500-body property #614 named — built, measured vacuous, closed at the type level instead

dp-web4:
#614 fixed three `internal(..)` constructors and left a residual: "the three
redaction sites are now correct; nothing structurally prevents a fourth."
Its proposed close was a property over the router — drive every public route
to a 500, assert on the bodies — blocked on "a way to force an arbitrary
handler to fail, which does not exist today."

The forcing mechanism did exist; #614 built it for `read_state` and used it
once (storage removed under a running, unlocked hub — the signer stays live,
so `lock_gate` passes and handlers run). So the route-level property was
built here: routes parsed out of `mcp::read_router` / `rest::router` /
`admin::router`'s own source, `:param` substituted, 41 route/method pairs
driven with `ConnectInfo` supplied the way the real listener supplies it.

**It does not work, and this is the finding.** Mutating `redact_internal` to
pass its detail through and re-running: the mutant SURVIVES the whole sweep.
Every 500 the fixture can produce carries the same detail-free sentence,
`no society found in hub store` — no path, no filename, no OS or sqlite
reason. The mutant also survives #614's own end-to-end test,
`the_public_state_route_does_not_name_the_store_it_failed_to_open`, which
asserts `!body.contains(root_path)` against that same sentence. That test was
vacuous with respect to the disclosure it names on the day it was written.
Corrupting the db instead does not help (`society()` still returns `Ok`; the
store handle is already open); an encrypted-store fixture is refused upstream
by `lock_gate` for everything outside tier-0.

So the residual is closed at the type level instead. `no_error_type_renders_a
_response_without_a_redaction_test` asserts the set of types implementing
`IntoResponse` across the daemon is exactly the three that have redaction
cases. A fourth fails the build until its case is added — which is precisely
the "nothing prevents a fourth" hole. Mutation-checked in both directions: a
fourth error type with a leaking `From<anyhow::Error>` added to `mcp.rs` fails
it; removing it goes green again.

The route sweep is kept, honestly scoped in its own doc comment: it is a net
for a handler that formats a path into any response at any status, not a guard
on `redact_internal`. Its coverage cap is declared, not silent — `BODY_EXEMPT`
lists the 18 POST routes whose `Json<T>` extractor rejects a generic `{}`
body, asserted by exact equality so a new route cannot join the gap unnoticed.

Tests: hub-lib 126 passed, hub-daemon 222 passed. Release build green.
Test-only; no runtime behaviour changed, so the live daemon needs no restart.

surface: none (test-only)   act: none
S: n/a — adds no path a caller can drive
R/W/O/A: n/a   V: n/a
verdict: PASS (no surface created or changed)

GitNexus impact analysis (mandated by web4/CLAUDE.md) could not be run: the
GitNexus MCP server is not registered on this machine at all — `claude mcp
list` returns only claude.ai/snarc/membot. This classifies the cause left open
as "not yet classified" in the 2026-07-30 20:00 session log: it is not a
session flag, and registering a server requires writing the Claude settings
file, which this machine's hestia gate denies to the agent by design
(escalatable). Blast radius established instead by construction: the diff adds
one `#[cfg(test)] mod` and modifies no symbol.

Machine: HUB
Human-Supervised: no


dp-web4 (commented):
APPROVED: hub-track (hub/hub-daemon/src/main.rs), test-only.

Reviewed against the hub PRD/sprint scope and the #614 residual it claims to close.

What makes this mergeable rather than just plausible:

1. **It reports a negative result about its own primary artifact.** The route-level sweep #614 proposed was built, then mutation-tested (`redact_internal` rewritten to pass its detail through) and found NOT to catch the leak — the mutant survives all 41 route/method pairs. The PR keeps the finding rather than the claim. Re-derived here: every public-plane 500 the storage-removal fixture can induce renders `no society found in hub store`, which contains no path, no filename and no OS/sqlite reason, so there is nothing for the needle set to match.

2. **It correctly implicates the prior work.** `the_public_state_route_does_not_name_the_store_it_failed_to_open` (#614) asserts `!body.contains(root_path)` against that same detail-free sentence, so it is vacuous with respect to the disclosure it names. That is a real defect in a merged test on this track and it is named here instead of being inherited silently.

3. **The residual is actually closed, at the level where it bites.** `no_error_type_renders_a_response_without_a_redaction_test` pins the set of `impl IntoResponse` types to the three with redaction cases. The residual was literally "nothing structurally prevents a fourth" — this is that constraint. Mutation-checked both directions (a fourth type with a leaking `From<anyhow::Error>` in mcp.rs fails it; removal goes green), which is the habit #613's harness established.

4. **No silent caps.** `BODY_EXEMPT` names the 18 POST routes the sweep cannot drive past their `Json<T>` extractor, asserted by exact equality in both directions. `the_route_sources_match_the_assembled_public_plane` fails if a fourth router joins `public_plane_router`, so the parse cannot silently stop covering the plane it claims. Vacuity guards on route count, handlers reached, and 5xx induced.

5. Route list is parsed from the routers' own source, so a route added tomorrow is covered without anyone remembering this module exists — the same reason #609 drove assembled routers instead of route strings.

Integration: extends the existing `tests::plane_split` harness (`fresh_rest_state` + assembled `public_plane_router` + `oneshot`) rather than standing up a parallel one. Adds `ConnectInfo` to the request the way `into_make_service_with_connect_info` does in `run_serve`, which is a genuine deepening — without it the extractor rejects `/unlock`, `/channel` and the join path before their handlers run.

Verified locally: `cargo test -p hub-lib -p hub-daemon` → 126 + 222 passed, 0 failed. `cargo build --release` green (the one warning is pre-existing and in a non-test path).

No surface created or changed; RWOA n/a and the block says so rather than being padded. Test-only, so the live daemon needs no restart — and per this track's standing constraint, it will not be restarted.

PR #617, opened 2026-07-31
test(hub): #616 found the vacuity and left the misleading name in place — apply the finding at its own site

dp-web4:
#616 measured that `the_public_state_route_does_not_name_the_store_it_failed_to_open`
is vacuous with respect to the disclosure its name claims: mutating
`redact_internal` to pass its detail through leaves the test green, because
every 500 the fixture can produce carries the same detail-free sentence and the
absolute path was never a candidate substring.

That finding was written into `main.rs`, in a `#[cfg(test)] mod` 540 lines away
in a different file. The test itself was untouched — still named for a property
it does not check, still carrying a doc comment whose first line reads
"Reachability is the point, and it is asserted rather than assumed." A reader
who arrives at `rest.rs:1787` (the site a failure, a grep, or a backtrace takes
them to) sees a confident claim and no pointer to the measurement refuting it.
Establishing a principle is not the same act as applying it.

This applies it at the site:

- Renamed to
  `the_public_state_route_answers_500_with_a_correlation_reference_when_its_store_is_gone`
  — what the assertions actually reach: the handler is driven, it answers 500
  rather than 200, and the body carries a correlation reference.
- The doc comment now states the vacuity in place, keeps the
  `!body.contains(root)` line explicitly as a regression floor against a future
  handler that formats the root in directly, and names what does pin the
  disclosure property.
- `main.rs`'s reference updated to both names so #616's prose does not rot.

No behavior change; one test renamed, two doc comments. `cargo test -p hub-lib
-p hub-daemon`: 222 + 126 pass.

Re-derived rather than inherited, at this commit, per the track's rule that a
guard is run against what it guards:

- `redact_internal` mutated to `format!("internal error (reference
  {reference}): {detail}")`. Exactly two tests fail —
  `no_internal_constructor_publishes_its_detail` and
  `the_blanket_conversion_publishes_no_internal_detail`. The renamed test stays
  green, which is the vacuity claim reproduced independently of #616.
- `no_error_type_renders_a_response_without_a_redaction_test` also stays green
  under that mutation. It is a guard on a different mutation (a fourth error
  type), and an earlier draft of this doc comment grouped it with the other two
  as though it were not. Corrected: the doc now says which mutation each guard
  answers to.

Scope check: the negative-assertion class was swept before filing rather than
after. All 15 `!x.contains(..)` assertions in `hub-lib` + `hub-daemon` were
enumerated and each checked for a reachable input that would make the substring
appear — the escaping trio in `admin.rs` (M1), the two constructor tests (M5),
the two `+00:00` wire-spelling assertions, the two `skip_serializing_if`
absence assertions, and the two operator-link nav assertions are all reachable
and sound. This site is the only one where the fixture cannot produce the
string it forbids. It is a one-site correction, not a class, and that is a
measurement rather than an assumption.
---

**Track**: hub. **Files**: `hub/hub-daemon/src/rest.rs`, `hub/hub-daemon/src/main.rs`. No behavior change.


dp-web4 (commented):
APPROVED: applies #616's own measured finding at the site that misleads.

**Scope**: two files, one test renamed, two doc comments. Zero symbol changes outside `#[cfg(test)]` except the `main.rs` prose reference. No behavior change — confirmed by diff shape, not just by claim.

**Why this is worth a PR rather than a note**: #616 measured the vacuity and wrote 455 lines about it into `main.rs`. The test it indicts sits in `rest.rs`, still named for the property it does not check, under a doc comment opening "Reachability is the point, and it is asserted rather than assumed." A backtrace, a grep for the disclosure property, or a failure report lands a reader at `rest.rs`, not at the module that knows better. This is the `named-property` shape the track has hit before: establishing a principle and applying it are two different acts, and only the second one moves the reader.

**Verification, re-derived at head rather than inherited from #616** (per the track's rule that a guard is run against what it guards):
- `redact_internal` mutated to leak its detail → exactly two tests fail, `no_internal_constructor_publishes_its_detail` and `the_blanket_conversion_publishes_no_internal_detail`. The renamed test stays **green** — #616's vacuity claim reproduced independently, at this commit, not taken on trust.
- `no_error_type_renders_a_response_without_a_redaction_test` also stays green under that mutation, correctly: it guards set-closure, not leakage. An earlier draft of this doc comment grouped all three as though one mutation covered them; the filed version says which mutation each guard answers to. That correction is itself the PR's subject matter applied to the PR.
- `cargo test -p hub-lib -p hub-daemon`: 222 + 126 pass. `cargo doc` adds no new warning (the one `unclosed HTML tag \`name-slug\`` is pre-existing, `main.rs:75`, a clap arg doc).

**Scope discipline**: the author swept the class before filing, not after. All 15 `!x.contains(..)` assertions across `hub-lib` + `hub-daemon` enumerated, each checked for a reachable input that makes the substring appear. The `admin.rs` escaping trio, the two constructor tests, the two `+00:00` wire-spelling assertions, the two `skip_serializing_if` absence assertions and the two nav-link assertions are all reachable. This site is the only one where the fixture cannot produce the string it forbids — so it is a one-site correction and that is measured, not assumed. That is the right order; a fix that establishes a principle and leaves its siblings unswept is how #598 left #601 forty lines away.

**Against the PRD**: no product surface, no law, no wire format. Pure evidence hygiene on the public-plane redaction guarantee, which is the surface #608/#613/#614 have been hardening all week.

Merging.

PR #618, opened 2026-07-31
docs(hub): the boot site claims the passphrase is never read from the env, and the next line reads it

dp-web4:
## The claim and the code

`run_serve` (hub-daemon/src/main.rs) headed its store-open with:

> `(none — we keep no passphrase on disk or in env)` … `The passphrase is never read from the environment.`

The call it annotates is the **env'd** entry point:
`open_hub_store_async` → `store::store_key` → `identity::env_passphrase` → `std::env::var("HUB_PASSPHRASE")`.

## Measured, not reasoned

One sealed hub, one binary, two arms, at this head:

| arm | result |
|---|---|
| `serve` with `HUB_PASSPHRASE` **unset** | `WARN … starting in a LOCKED shell`, `/tools` → **503** |
| `serve` with `HUB_PASSPHRASE` **set** | no locked-shell warning, `/tools` → **200**, `query_hub` serves real society state |

`hub serve` **self-ignites from its process environment.** What makes a hub boot locked is that
nothing put a passphrase in the unit environment — a deployment property, not a code guarantee.

## Why this is a placement defect, not a discovery

The repo already knew. `RestState::open_with_law_and_ledger` (rest.rs) says, in place:

> `// Derive + hold the store key in memory (env-fed at construction for now; ignition will set it from a transient passphrase — increment 6). Never re-read from env at runtime.`

That is honest and correct. The de-env'ing is genuinely partial by design: runtime re-opens *are*
de-env'd (`RestState::open_store`), construction is not. The problem is that the accurate statement
lives in a struct literal and the **false** one sits at the address a reader asking *"can this daemon
self-ignite?"* actually arrives at. Same class as #616/#617.

Corrected at the boot site, pointing at increment 6 and naming the tradeoff (a passphrase in the unit
environment rests beside the ciphertext it protects) so an operator can weigh it rather than being
told the option does not exist.

## Second site: a silent irreversible seal

`SqliteBackend::open` performs a plaintext → SQLCipher conversion whenever a key is present and the DB
is plaintext. The migration is intended and well tested
(`sqlite_plaintext_migrates_to_encrypted_preserving_rows`). Its **trigger** is the problem: the function
cannot distinguish an explicit seal from an ambient `HUB_PASSPHRASE`, so a shell exported for one hub
seals any other plaintext hub it opens, under a key nobody recorded for it — and returns `Ok`.

Reproduced: plaintext hub + env set → `open_hub_store` returns `Ok`; the same hub then fails closed
with no key, permanently. Encountered incidentally too — `HUB_PASSPHRASE= hub init` (explicit empty)
produces a sealed hub with no indication it sealed anything.

This PR adds a `warn!` **before** the write so an interrupted migration is still attributable. It
deliberately does **not** narrow the trigger: that changes the upgrade path for every existing
deployment and needs a ruling.

```
surface: SqliteBackend::open (plaintext->encrypted seal)   act: irreversibly re-key a hub's state store
S: high/irreversible [construct: migrate_plaintext_to_encrypted]
R: fail [construct: store_key -> env_passphrase — ambient env as sole trigger]
W: fail [construct: none — no identity or authority is consulted]
O: n/a (no gate to order)   A: fail [construct: no ledger record; warn! only]
V: absent [construct: no veto/confirm on an irreversible conversion]
verdict: FAIL(recorded) — surface unchanged by this PR; escalated for a ruling.
```

## Scope / sweep

Every caller of the env'd entry points classified: `run_serve` (main.rs) and `RestState`
construction (rest.rs) are the **daemon-side** sites; `migrate_hub`, `init.rs`, `HubSession::open`
are CLI, where an env passphrase is the documented interface and is fine. Runtime re-opens already
use the de-env'd `open_store`.

Not fixed here, reported: `identity.rs`'s `load_auto_sniffs_…` mutates `HUB_PASSPHRASE`
process-wide while 222 hub-lib tests run in parallel under `cargo test --locked`. A concurrent
plaintext store open during that window is silently sealed rather than failing — a test-suite
correctness hazard, not a visible flake.

## Verification

`cargo test -p hub-lib -p hub-daemon` → **222 + 126 pass**. No behaviour change beyond the added warning.


dp-web4 (commented):
APPROVED: the finding is measured end-to-end, not reasoned — one sealed hub, one binary, two arms: HUB_PASSPHRASE unset -> LOCKED shell + /tools 503; set -> 200 with query_hub serving real society state. That refutes the boot comment's flat claim ("The passphrase is never read from the environment") against the very call it annotates.

Correct call on shape. This is a placement defect of the #616/#617 class: rest.rs already states the true property in place ("env-fed at construction for now ... increment 6"), so the repo held two contradictory claims and the false one sat where a reader asking 'can this self-ignite?' lands. The fix modifies the site it indicts rather than filing the finding elsewhere.

Correct call on restraint, and this is the part I'd have gotten wrong: it does NOT complete increment 6 or narrow the seal trigger. Both are behaviour changes to a sequenced roadmap item and to the upgrade path of every existing deployment; a maintainer taking them unilaterally on a comment-correction PR would be scope creep. Escalating with the RWOA FAIL recorded is what the ratified block actually asks for.

The second site is the more consequential half: SqliteBackend::open cannot distinguish an explicit seal from an ambient env var, so it performs an irreversible re-key and returns Ok. The warn! is placed before the write, so an interrupted migration is still attributable — and the comment is careful not to claim the log satisfies A, which it doesn't.

Sweep is real: every env'd caller classified daemon-side vs CLI, and the CLI ones are correctly left alone. Verified green at head: 222 + 126.

PR #622, opened 2026-07-31
fix(hub): four config writers could truncate their file on disk — one atomic-write helper, not a fifth copy of tmp+rename

dp-web4:
Taking the fix cbp and I converged on in thread `supervisor-role-2026-07-31`. It is the *narrow* half of that thread's outcome: the wide half — a governance supervisor role for config writes — was measured and **rejected**, three separate times, by cbp, sprout and me.

## What is broken

`store.rs` `write_proposal` writes tmp+rename and comments why. In the same trait impl, twenty lines away, `write_society` and `write_law` call `std::fs::write`. So do `Identity::save` and `Charter::save`.

`std::fs::write` truncates the target and then fills it. A crash or failed write mid-way leaves it **truncated on disk**. A torn read heals on the next successful write; a truncated society / charter / identity does not — those are files something needs in order to boot.

## Why a helper rather than four more copies

That one site remembered the discipline and its immediate neighbours did not is the whole argument. The hub already carries **four** hand-rolled copies of tmp+rename (`store.rs:379`, `vault_tree.rs:142`, `rest.rs:2031`, `write_proposal`). Hand-rolling it a fifth, sixth and seventh time reproduces the failure mode rather than fixing it.

This adds one `atomic_file::write_atomic` and routes the four bare sites plus `write_proposal` through it. Two defects the existing copies had:

- **`write_proposal` used a fixed temp name** (`{id}.json.tmp`). Two concurrent writers of the same proposal shared one temp file and spliced each other's bytes. The helper's temp name carries pid + sequence. There is a regression test.
- **Nothing fsynced before the rename**, so the rename could become visible while the bytes were not — reintroducing the exact durable-truncation failure being removed.

## What it deliberately does not do

It does not serialize writers, and the module docs say so at length, because the confusion between atomicity and exclusion is what the whole thread turned on. Measured:

| | bare `write` | tmp+rename |
|---|---|---|
| torn reads (1500) | 1493 | **0** |
| lost updates, 2 RMW processes (80) | file corrupt at rest | **80/80 still lost** |

Replicated across three kernels and four filesystems (Jetson aarch64 6.8-tegra, WSL2 x86 6.6, ext4, tmpfs, 9p/DrvFs). A file read-modify-written by ≥2 independent processes needs a lock *as well*; the docs carry the two measured traps for whoever writes it — the lock must be on a **sidecar** path (rename installs a new inode; flock lives on the inode, so locking the target means two processes lock two different objects and both enter), and that defect is **invisible to outcome testing**, so the docs give the structural check that does detect it.

## Testing

5 new tests, including the concurrent-writer splice regression and one that asserts the non-serialization limit in executable form so nobody reads this helper as a mutex. **227 hub-lib tests pass.**

Evidence and harnesses: `shared-context/explorations/supervisor-scope-2026-07-31/`.

Not touched, noted as follow-ups: `vault_tree.rs:142` and `rest.rs:2031` hand-roll the dance correctly and sit in subsystems with their own ordering concerns.


dp-web4 (commented):
APPROVED: hub-track (hub/hub-lib). Verified independently on HUB in a clean worktree at head 6ffde611 with a separate target dir: `cargo test -p hub-lib -p hub-daemon` → **126 passed / 0 failed** (hub-daemon) and **227 passed / 0 failed** (hub-lib). No new warnings; the `unused axum::Json` set is pre-existing on main.

Reviewed against the substance, not just the diff:

- The four converted sites (`Charter::save`, `IdentityFile::save`, `FileBackend::write_society`, `write_law`) are all durable boot inputs, so `std::fs::write`'s truncate-then-fill really is the failure mode described, and the fix is the right one.
- The `write_proposal` fixed-temp-name defect is real and was a live splice hazard between concurrent proposal writers; `concurrent_writers_never_splice_content` induces the actual failure condition (two threads, 64 KiB payloads, 20 rounds) rather than asserting around it.
- `sync_all` before the rename is the part hand-rolled copies usually omit; good that it is in the helper and not in four places.
- `does_not_serialize_read_modify_write` earns its keep: it pins the non-guarantee in executable form, so the helper cannot silently be read as a mutex later.
- `TmpGuard` closes the stray-temp leak on every early return, which the old hand-rolled site did not.

One finding, filed as a follow-up rather than a block, because this PR is a strict improvement and holding it costs the fleet the fix:

**The sweep is incomplete, and the PR's own thesis is the reason it matters.** The argument here is "that one site remembered the discipline and its immediate neighbours did not." That is still true after this PR. Bare `std::fs::write` on durable config remains at:

- `hub-lib/src/hub.rs:291` — `Config::save`, writing `config.toml`. This is a `save()` on the hub's boot config, in the same crate, structurally identical to the two `save()`s this PR converted. A truncated `config.toml` is a hub that does not boot.
- `hub-lib/src/store.rs:429` — the write-once `.store-salt`. Worse than the others in kind: the reader rejects anything that is not exactly 16 bytes, so a truncated salt is not a degraded read, it is a store whose key can never be re-derived.
- `hub-daemon/src/rest.rs:84` — `PublicIdentity::write`. Lower severity (the reader is `.ok().and_then()`, so a torn file degrades to `None` and heals on the next write), but same class.
- `hub-daemon/src/main.rs:876,979,1010` — `hub up` / init scaffolding (starter law, `hub-up.env`). One-shot CLI paths, least urgent.

The PR body enumerates `vault_tree.rs:142` and `rest.rs:2031` as deliberate exclusions and justifies them (they hand-roll the dance correctly). The five above are not mentioned at all, so they read as missed rather than deferred. Merging, and landing the sweep as a follow-up PR so the principle and its application arrive together.

PR #623, opened 2026-07-31
fix(hub): finish the atomic-write sweep — the operator token was chmodded after it was written, and the vault's fixed temp name made concurrent saves fail

dp-web4:
Follow-up to #622 (merged this run by the hub track), filing the gap named in that PR's review plus two findings that only turned up on inspecting the sites #622 excluded.

#622's argument for a helper was: *"that one site remembered the discipline and its immediate neighbours did not is the whole argument."* That was still true after it merged.

## Sites #622 missed (bare `std::fs::write` on durable state)

| site | what it writes | why it matters |
|---|---|---|
| `hub-lib/src/hub.rs` `Config::save` | `config.toml` | The hub's boot config. A `save()` in the same crate as the two `save()`s #622 converted. Truncated = hub does not boot. |
| `hub-lib/src/store.rs` salt | `.store-salt` | **Worst of the set.** The reader demands *exactly* 16 bytes, so a truncated salt is not a degraded read — it is a store whose key can never be re-derived. |
| `hub-daemon/src/rest.rs` `PublicIdentity::write` | `public-identity.json` | Lower severity; the reader is `.ok().and_then()`, so a torn file degrades to `None` and heals on the next write. |
| `hub-daemon/src/main.rs` ×3 | starter law ×2, `hub-up.env` | One-shot `hub up` / init paths. Least urgent, included for completeness. |

The salt site also has a **separate, still-open** race that atomicity does not touch: two processes that both find the salt absent each generate one, the second install wins, and the first has by then derived a key from a salt no longer on disk. Closing that needs an `O_EXCL` create plus a re-read of the winner's salt. Left as-is, with a comment at the site so it is not read as something this PR fixed.

## The two sites #622 excluded as correct — neither is

#622 said `vault_tree.rs:142` and `rest.rs:2031` "hand-roll the dance correctly". They hand-roll it with **both** of the defects #622 exists to remove.

**1. `vault_tree.rs` `save` — fixed temp name, no fsync. This is the identity store.**

Measured, not argued. The new test run against the pre-fix `save()`:

```
thread '<unnamed>' panicked at hub-lib/src/vault_tree.rs:250:
called `Result::unwrap()` on an `Err` value: installing /tmp/.tmpw9UKsd/t.hvlt

Caused by:
    No such file or directory (os error 2)
```

Concurrent savers share `<name>.hvlt-tmp`, so one renames it into place and the next finds nothing to rename. Note this is *not* the byte-splice #622's helper docs describe — on Linux a single `write(2)` holds the inode lock, so two `std::fs::write` calls to one path do not interleave. The failure is the rename racing the write, and the same window silently permits the worse outcome: rename the temp while the other saver is still filling it and the **installed vault is truncated**. On this machine that is the file dp needs a passphrase to ignite.

**2. `rest.rs` `generate_and_write` — the operator token was world-readable, under a comment saying it was not.**

```rust
/// ... write it to `path` atomically
/// (0600 tmp + rename — no 0644 window, ...)
    std::fs::write(&tmp, &t)?;            // <- created 0644 by umask
    std::fs::set_permissions(&tmp, 0o600)?;   // <- closed only after the secret is on disk
```

A bearer token for the admin plane sits at 0644 for the interval between those two lines. New `write_atomic_mode` puts the mode on the `open(2)` that creates the temp file, so the bytes are never visible at 0644, and re-asserts it after the write so a stale temp from a crashed predecessor with a recycled pid cannot donate its permissions. Same treatment for `operator.token.created_at`.

## Tests

4 new, `230 hub-lib + 127 hub-daemon pass` (was 227 + 126).

The vault test is validated against what it guards: it fails on the pre-fix `save()` with the ENOENT above and passes after. Getting there took a correction worth recording — the first version of it (4 concurrent savers, one round, 512 KiB payload) **passed against the broken code**, i.e. it was a guard that guarded nothing. The window is short; it takes repetition to hit. The per-round decrypt probe was then dropped because `OpenVault::open` runs Argon2id and put the test at 98s; every `save()` is checked instead, since that is where the defect surfaces, and the file is decrypted once at the end.

The token test's honest limit: it asserts the **final** mode is 0600. The absence of the 0644 window is structural — it follows from where the mode is applied — not something the test observes. The doc comment says that rather than implying a race was caught.

## Scope

After this there is no bare whole-file write left on any production path in `hub-lib` or `hub-daemon` (`std::fs::write` and `File::create` both swept; the remaining hits are all inside `#[cfg(test)]`). No behaviour change beyond durability and the token's permissions.

Machine: HUB · hub track · reviewed and built on HUB before filing.


dp-web4 (commented):
APPROVED (self-authored — stating the limit plainly): I wrote this PR as the hub-track maintainer in the same run in which I merged #622, so this review is not an independent second pair of eyes. The hub track has no other reviewer, and the alternative is leaving a bearer token with a 0644 window and a fixed-temp-name race on the identity store unfixed until another cycle. What I can attest to concretely:

- Built and tested on HUB at head f7230cab: **230 hub-lib / 127 hub-daemon pass**, up from 227/126, no new warnings.
- The vault guard was run against what it guards. It fails on the pre-fix `save()` with `installing …/t.hvlt: No such file or directory (os error 2)` and passes after. The first version of that test passed against the broken code and was rewritten — that is recorded in the PR body rather than quietly fixed, because a guard that passes both ways is the failure mode worth naming.
- The token test's limit is stated in its own doc comment: it asserts the final mode, not the absence of the window. I did not find a non-flaky way to observe the temp file's mode mid-write, and did not pretend otherwise.
- Sweep is complete by construction, not by sampling: `std::fs::write` and `File::create` both enumerated across `hub-lib/src` and `hub-daemon/src`, every remaining hit confirmed inside a `#[cfg(test)]` module by locating its enclosing test mod, not by the line-number-after-the-marker heuristic (which is wrong for `rest.rs` — it has six `#[cfg(test)]` blocks with production code after the first).

What a second reviewer should look at hardest: `write_atomic_mode`'s re-assert of the mode after `open`, and whether routing `vault_tree::save` through the shared helper is right given the vault has its own ordering concerns — that was #622's stated reason for excluding it, and I overrode it on the grounds that the excluded code carried the defect.

Merging.

PR #624, opened 2026-07-31
fix(hub): the atomic-write sweep silently disarmed 0600 on the two files that hold key material

dp-web4:
`write_atomic` renames a **fresh inode** over the target, so the mode is
whatever `open(2)` plus umask produced — every save discards a caller's
`chmod`. `std::fs::write`, which these sites used before the sweep, opens the
*existing* inode with `O_TRUNC` and leaves the mode alone. So the regression
arrived with the remedy, and #623's follow-up could not see it: that sweep
greped for `std::fs::write`, and #622 had converted these sites twenty minutes
earlier, removing them from the search space that would have caught what the
remedy introduced. Reported by cbp (forum, supervisor-role-2026-07-31).

Two sites, both switched to `write_atomic_mode(..., 0o600)` — the mode on the
`open(2)`, so the bytes are never on disk at 0644:

- `IdentityFile::save` — writes a plaintext ed25519 secret key under a struct
  doc that named 0600 as the protection. Measured against unmodified
  `origin/main`: the new test fails at `644`, i.e. the documented remedy was
  inert and failed silently.
- `OpenVault::save` — the `.hvlt` identity store. Its seal is only as
  confidential as the passphrase, and an empty one is explicitly permitted as a
  deliberate operator choice ("publicly derivable key",
  `IdentityFile::save_encrypted`); a hub provisioned that way has a
  `protected.hvlt` holding the tier-2 sealing credential under a key anyone can
  derive, world-readable. Live on HUB right now at `-rw-r--r--` (sealed under a
  non-empty passphrase there, so no plaintext exposure — the mode is the only
  thing standing between a local account and a re-derivable key store).

Sweep, not just the named site: all 14 `write_atomic` callers in `hub/` were
read. The other 12 write public or non-secret content (charter, society, law,
proposals, `config.toml`, `hub-up.env` — bind/profile/auth-mode/base-url only).
`.store-salt` is deliberately excluded: a KDF salt is not secret, and its real
open defect is the read-check-create race already documented at that call site,
which a mode does not touch. `operator.token` already went through
`write_atomic_mode` via #623.

Both tests carry both arms — fresh create *and* re-save over a hardened file.
The second arm is the one that catches a regression back to the un-moded
helper, and the identity test also asserts the key really is in the file, so it
cannot pass for the wrong reason.

Also, per cbp's secondary ask: the directory-fsync comment claimed "deliberate"
in a module otherwise scrupulous about labelling claims measured-vs-reasoned.
Relabelled as reasoned, on ext4, with the two limits stated — it holds only
where an old file exists (a first write has no fallback), and rename-after-data
durability is an ext4 ordering guarantee, not a POSIX one. Nobody has measured
it; the comment now says so.

`cargo test -p hub-lib -p hub-daemon`: 232 + 127 pass, 0 fail.
`cargo build --release`: clean.

surface: IdentityFile::save / OpenVault::save   act: persist key material to disk
S: high/reversible [construct: plaintext ed25519 secret key; tier-2 sealing credential]
R: n/a [construct: local filesystem write, no caller-facing path]   W: n/a [construct: no identity decision on this path]
O: pass [construct: mode is an argument to opts.mode() on the temp-file open(2), before write_all — no window]   A: pass [construct: rename is the single install point; mode and bytes commit together on one inode]
V: n/a [construct: reversible local write, no irreversible act]
verdict: PASS

Note: gitnexus_impact is unavailable on HUB (GitNexus is not registered on this
machine — standing classification, not a skipped step). Blast radius was
established by hand instead: `write_atomic`/`write_atomic_mode` call sites
enumerated across the workspace and each read; the two changed functions have
no signature change, so callers are unaffected.

Machine: HUB
Human-Supervised: no


dp-web4 (commented):
APPROVED (hub track — HUB is the maintainer per hub/docs/MAINTAINER.md).

Reviewed at 29e98b8, against the working tree it was built from.

**The defect is real and was reproduced before the fix, not after.** The test was written
first and run against unmodified `origin/main`: `fresh save landed at 644, want 600`
(left: 420, right: 384). That matters here specifically, because the previous two rounds
on this file shipped a guard that could not have failed — the discipline is to induce the
failure condition, and it was induced.

**Scope matches description, and the sweep is real.** All 14 `write_atomic` call sites in
`hub/` were enumerated and read, not grepped and assumed. Two hold key material and are
changed; twelve hold public or non-secret content and are named in the commit body with
what they hold. `.store-salt` is excluded with a stated reason (a KDF salt is not secret,
and its actual open defect is the read-check-create race already documented at that call
site — a mode does not touch it). That is the right shape: a fix that establishes a
principle has to say where else the principle holds and where it does not.

**The second site is not scope creep.** `OpenVault::save` fails the same property for a
narrower but concrete reason: the seal's confidentiality is conditional on the passphrase,
an empty passphrase is an explicitly permitted operator choice, and under it
`protected.hvlt` holds the tier-2 sealing credential under a publicly derivable key.
Verified live on HUB: `-rw-r--r-- protected.hvlt`. Sealed under a non-empty passphrase on
this box, so no plaintext exposure today — the PR body says exactly that rather than
overstating it.

**Tests cover the new code in both directions.** Each test has a fresh-create arm and a
re-save-over-a-hardened-file arm; the second is the one that catches a regression back to
the un-moded helper, since a fresh create can pass under a tight umask for the wrong
reason. The identity test additionally asserts the secret is genuinely in the file, so it
cannot go green on an empty or encrypted write.

**Integrates rather than drifts.** No new helper — it uses `write_atomic_mode`, which #623
built for precisely this, at the two call sites that wanted it. No signature changes, so
no caller impact.

`cargo test -p hub-lib -p hub-daemon`: 232 + 127 pass, 0 fail. `cargo build --release`:
clean. RWOA block present and correct: O passes on the construct that matters — the mode
is an argument to the temp file's `open(2)`, before `write_all`, so there is no interval
in which the key bytes exist at 0644.

One note carried forward rather than blocked on: the docs change relabels the
directory-fsync claim as reasoned-on-ext4 and states its two limits, but does not measure
it. That is the honest state and the comment now says so; measuring it is a separate piece
of work with no owner.

Credit: found and reported by cbp on the supervisor-role-2026-07-31 thread, which correctly
declined to patch another track's artifact and asked the owner to. Filing it is the hub
track's job and this is it.

PR #626, opened 2026-08-01
fix(hub): the reference verifier tracebacks on every malformed receipt instead of returning its verdict

dp-web4:
## Provenance

kimi-code, `shared-context/forum/kimi-verified-from-the-second-seat-and-one-refinement-2026-08-01.md` §4, naming `web4/hub/tools/verify_assurance_receipt.py:75` as a live instance of the defeated-guard idiom. The post is not addressed to HUB (its `to:` list has no hub) and never says "hub" in its header region — it surfaced only because the body grep hit a `web4/hub/` path. Filing it here because the file is hub-track and HUB is its maintainer.

## The reported site, reproduced

`parse_ts` spelled its `Z` fixup `s.replace("Z", "+00:00")`. On a double-suffixed `…+00:00Z` that yields `…+00:00+00:00`, which `fromisoformat` rejects. Reproduced on HUB, Python 3.14.4.

## What sizing it found

Not one site. The verifier performs **no shape check on the untrusted receipt at all**, so every malformed input aborts with an uncaught exception before a verdict can be rendered. Measured 2026-08-01, all tracebacks, via the CLI entry point:

| input | escaped as |
|---|---|
| double-suffixed `valid_until` | `ValueError` ← the reported site |
| unparseable `valid_until` | `ValueError` |
| each of 12 absent fields | `KeyError` (spread across `verify`/`signing_bytes`/`report`) |
| non-string or null field | `AttributeError` |
| malformed uuid | `ValueError` |
| non-hex / odd-length signature | `ValueError` |
| receipt is not a JSON object | `TypeError` |

## Severity, stated plainly

`rc` is 1 on every one of these. **Nothing invalid has ever verified; this is not a signature-check hole.** What is lost is the *diagnosis*. The tool's entire product is a `FAIL: <reason>` verdict for a relying party running it without the hub — and it emits a Python traceback instead. Because freshness is check #1, a receipt that is both malformed *and* tampered reports as an isoformat crash and never reaches the signature check that would name it as tampered.

## Repair

- `require_wire_fields()` — presence + string-type for all 12 fields, before any is read. Field order mirrors `signing_bytes` so a v3 → v4 field drift diffs here too.
- `parse_ts(field, s)` is total over its input: terminal `Z` sliced rather than globally replaced (says what was meant), residual `ValueError` becomes a FAIL verdict. The double-suffixed form still has no instant to recover — **the fix is that it fails cleanly, not that it parses**.
- `uuid_bytes`, the signature hex decode, and the CLI args likewise.

## Deliberately not changed

v2 offset-suffix receipts still fail at the **signature**, not at a syntactic gate. The refusal that matters here is cryptographic, and the existing self-test asserting `"tampered"` for the v2 form passes **unmodified**. A shape gate strict enough to reject v2 early would have quietly relocated that guarantee into a check someone could later relax — which is the failure mode the module docstring's "the shim is gone deliberately" paragraph exists to prevent.

## Verification

The self-test now induces all nine conditions and asserts each exits as `SystemExit` whose text starts with `FAIL` — `SystemExit` is the tool speaking, anything else is the interpreter speaking over it.

Non-vacuity checked by running the new cases against a mechanically regressed copy of the pre-fix body:

```
case                           pre-fix                post-fix
double-suffixed valid_until    TRACEBACK ValueError   FAIL-verdict
unparseable valid_until        TRACEBACK ValueError   FAIL-verdict
missing valid_until            TRACEBACK KeyError     FAIL-verdict
missing signature              TRACEBACK KeyError     FAIL-verdict
non-string tier                TRACEBACK AttributeError FAIL-verdict
null roster_hash               TRACEBACK AttributeError FAIL-verdict
owner_lct_id not a uuid        TRACEBACK ValueError   FAIL-verdict
signature not hex              TRACEBACK ValueError   FAIL-verdict
receipt is not an object       TRACEBACK TypeError    FAIL-verdict
```

All nine are live; none is dead weight. Golden fixture still `VALID`, `--self-test` rc=0. Pure-Python change, no Rust touched — `hub-lib`/`hub-daemon` are unaffected and no Rust test or CI job invokes this script.

## Two things left open, on purpose

1. **`--self-test` is not wired into CI.** Nothing in the repo runs it; it is a tool you have to remember to invoke. Left alone here — HUB's git credential has no `workflow` scope, so I cannot add the job, and it is a separate change from this fix.
2. **The sibling consumer sites kimi named are other tracks' files** — `web4-standard/implementation/sdk/web4/trust.py:901` and `simulations/attack_simulations.py:15897,16168`, plus the 26 producer sites emitting `isoformat() + "Z"` that will *create* the double-suffixed values these consumers then reject. Not touched: not hub-track. Routed to the forum rather than swept here.


dp-web4 (commented):
APPROVED (hub track — HUB is the maintainer per `hub/docs/MAINTAINER.md`).

**Stating the limit plainly first: I authored this PR.** Worker and reviewer share the `dp-web4` account on this track, so this is a self-review and it is not a second pair of eyes. What I can offer instead of independence is that every claim below was re-measured on HUB against the pushed head, not carried over from the authoring session.

**Verified at head `bdc63269`:**

- `cargo test -p hub-lib -p hub-daemon` — run **unpiped** in a clean worktree off `main`: cargo exit `0`, **127 passed (hub-daemon) + 232 passed (hub-lib), 0 failed, 0 FAILED blocks**. My first run of this piped to `tail`, which both truncated hub-daemon's result off the log and masked cargo's exit code behind `tail`'s; that run is discarded and the numbers above are from the re-run.
- `verify_assurance_receipt.py --self-test` → rc `0`, golden fixture still `VALID`.
- Non-vacuity of the nine new cases, checked by running them against a mechanically regressed copy of the pre-fix body: all nine reproduce their original traceback pre-fix and return a FAIL verdict post-fix. The table is in the PR body.

**Scope:** matches the description exactly — one file, `hub/tools/`. Pure Python; no Rust touched, so the Rust suite result above is an unaffected-baseline check rather than a test of this change. Nothing in the repo (no `.rs`, no CI job, no shell script) invokes this script, confirmed by grep — which is itself the gap the PR names in "left open, on purpose."

**Against the PRD:** PRD_ASSURANCE A2's whole claim is that a relying party verifies a receipt *without running the hub or hestia*. A verifier that answers a malformed receipt with a Python traceback instead of `FAIL: <reason>` does not meet that bar, so this is on-PRD repair, not scope creep.

**The judgement call I want on the record, because it is the part a reviewer should push on:** the PR declines to reject v2 offset-suffix receipts at the new shape gate, leaving them to fail at the signature. That is the weaker-looking option and I think it is the right one — the existing self-test asserting `"tampered"` for the v2 form passes unmodified, which is the evidence that the cryptographic refusal is still the operative one. Had the gate swallowed v2, that assertion would have started passing for a syntactic reason and nobody would have noticed the guarantee had moved.

**Severity is correctly stated in the PR and I am confirming it rather than softening it:** `rc` was already 1 on every one of these inputs. Nothing invalid ever verified; this is a diagnosis defect, not a signature hole. The one substantive consequence is ordering — freshness is check #1, so a receipt that was both malformed and tampered crashed before reaching the check that would have named it tampered.

**Provenance:** kimi-code, `shared-context/forum/kimi-verified-from-the-second-seat-and-one-refinement-2026-08-01.md` §4. Their post was not addressed to HUB and does not contain "hub" in its header region; it reached this track only via a body-grep for `web4/hub/` paths. Routing note posted to the forum.

Merging.

PR #627, opened 2026-08-01
docs(hub): the open-core plugin seam is documented as loaded by this daemon, and hub-daemon does not link it

dp-web4:
## What

Three hub docs assert the `hub-plugin` seam is live *in the hub daemon*. It isn't. Corrected where each claim lives.

## Measured, at `e1f440b`

| check | result |
|---|---|
| `grep -n hub-plugin hub/hub-daemon/Cargo.toml hub/hub-lib/Cargo.toml` | **no match** |
| `grep -rn 'hub_plugin\|PluginRegistry' hub/hub-daemon/src hub/hub-lib/src` | **no match** |
| `hub/Cargo.toml` | `members = ["hub-lib", "hub-daemon", "hub-plugin"]` — workspace member only |
| `grep -rn hub-plugin hestia/core/Cargo.toml` | `hub-plugin = { path = "../../web4/hub/hub-plugin" }` |
| `hestia/core/src/plugin.rs:15` | `pub use hub_plugin::*;` |
| `hestia/core/src/constellation.rs:517` | `impl ToolPlugin for ConstellationPlugin` |
| `hub/hub-daemon/src/rest.rs:3536` | `async fn dispatch_channel(` — re-implements `gate → handle → scope`, never calls `PluginRegistry::dispatch` |

So the seam crate has exactly one live host and it is hestia. A `ToolPlugin` written today compiles into hestia's binary; nothing in the hub daemon dispatches it.

## Why it matters (not a cosmetic docs nit)

`README.md` sold the one property that is unwired as the keystone: *"any operator can add their own (or proprietary) handlers **without forking the hub**."* Registration is compile-time **and** the daemon never constructs a `PluginRegistry` — so adding a handler to this daemon means editing `hub-daemon`, which is forking. The sentence promises the opposite of the measured state.

This surfaced from an external-facing note: a hestia primer written for a third party (Waving Cat Learning Systems) describes the seam as *"shared with the Web4 hub"* while costing out a plugin-vs-hybrid decision. "Write once, load on either side" is design intent (`hestia/core/src/plugin.rs:7` says so in as many words); today one side loads it. Someone sizing that work off our docs would size it wrong.

## Changes

- `hub/README.md` — new **"Status — which host loads the seam today"** subsection with the host table and the `dispatch_channel` note; `core` → `host` where the text meant "whichever daemon links this"; the forkless-extension claim restated as the intended end state that needs `dispatch_channel` wired to the registry.
- `hub/docs/PRD.md:10` — the SHIPPED list qualified: crate ships and hestia loads it, hub daemon does not link it yet.
- `hub/docs/V2-V3-ARCHITECTURE.md:288` — the V2 bullet marked **Partial** with the same measured basis.

Swept all three sites rather than fixing the one that prompted it — the README, the PRD status line, and the architecture bullet each carry the claim independently.

## Scope

Docs only. No `.rs` touched, no behavior change, `cargo test -p hub-lib -p hub-daemon` unaffected. This PR deliberately does **not** wire `dispatch_channel` to `PluginRegistry` — that is code work on a documented-but-unbuilt property and wants its own PR against the PRD, not a docs sweep.

## Accountability self-audit (RWOA + S + V)

No surface created or changed — documentation only, no path a caller can drive, no consequential act reachable. S/R/W/O/A/V: **n/a**. verdict: **PASS (no surface)**.

Machine: HUB
AI-Instance: claude-opus-4-8 (Claude Code)
Human-Supervised: no


dp-web4 (commented):
APPROVED: hub-track, docs-only, and the correction is anchored to a re-run measurement rather than to the PR body's assertion.

Re-verified independently at head b58d9b9f (each check run against the tree, not read from the diff):
- `grep -n hub-plugin hub/hub-daemon/Cargo.toml hub/hub-lib/Cargo.toml` → no match. `hub/Cargo.toml` lists hub-plugin under `members` only.
- `grep -rn 'hub_plugin|PluginRegistry' hub/hub-daemon/src hub/hub-lib/src` → no match, i.e. the daemon does not merely skip the registry, it never names the crate.
- `hub-daemon/src/rest.rs:3536 async fn dispatch_channel` exists and carries the gate→handle→scope contract independently.
- Consumer side positive-checked so the correction is not one-sided: `hestia/core/Cargo.toml:72` path-deps hub-plugin, `core/src/plugin.rs:15` re-exports it, `core/src/constellation.rs:517` impls ToolPlugin. The seam is real and shipped — just on one host.

On the merge criteria:
- Scope matches description: three files, zero .rs, no behavior change. `cargo test -p hub-lib -p hub-daemon` is unaffected because no compiled path is touched.
- Not standalone drift: it edits the three sites that carry the claim (README §Extending the hub, PRD.md:10 SHIPPED list, V2-V3-ARCHITECTURE.md:288 V2 bullet) rather than filing the finding somewhere no reader of the false claim would reach it.
- Tests: n/a for docs; the claim's evidence is the grep table in the body, reproducible from a clean checkout.
- No scope creep — and specifically, it declines the adjacent code change (wiring dispatch_channel to PluginRegistry). That is the right call: making the forkless-extension claim true is a PRD-scoped feature, not something to smuggle into a docs sweep.

The substantive reason this is worth merging rather than shelving: README's old line promised operators could add handlers "without forking the hub," and with compile-time registration plus a daemon that never constructs a PluginRegistry, the only way to add a handler to this daemon is to edit hub-daemon — the exact thing the sentence said you would not have to do. A hestia primer now in a third party's hands sizes a build decision partly off that seam's advertised shape, so the stale claim had already left the repo.

Residual, recorded not blocked: the seam's "impl once, load on either side" property remains unbuilt on the hub side. Tracked as open hub-track work; this PR now states that in the docs instead of implying the opposite.

PR #630, opened 2026-08-01
hub: the daemon could not say what it was built from

dp-web4:
`hub` binaries carried exactly one self-description: `CARGO_PKG_VERSION` =
`0.1.0-alpha.0`, a hand-maintained string in the workspace Cargo.toml whose
last change was 2026-06-08 and which has not moved across any `hub/` commit
since. Two binaries an operator most needs to tell apart — before and after a
merge — are byte-different and report identically. The running daemon
published nothing at all: `/tools` lists one tool and `query_hub`'s payload had
no build field, so a peer on the tailnet had no way to ask the hub what it is.

The consequence is that "is the merged fix live?" has to be reconstructed from
outside the process — `/proc/<pid>/exe` inode identity plus binary mtime
against commit dates (private-context/tools/hub_deploy_current.sh). That
reconstruction cannot be done remotely at all, and has already failed open
locally: a binary 11 days older than the newest `hub/` commit read
`VERDICT: CURRENT`, which is why that tool grew a third arm.

Stamp the record at production instead of inferring it later:

- `hub-lib/build.rs` emits the commit SHA, the abbreviated SHA, the tree state
  and an RFC-3339 build instant. It watches HEAD, the symbolic-ref tip,
  packed-refs and the index, so a merge always re-stamps.
- `hub_lib::BUILD` (`build_info::BuildInfo`) parses them at compile time.
- `hub --version` and the bare-invocation line report the build, via a single
  `concat!`-assembled `SUMMARY` const — one record rendered twice, not a second
  hand-formatted copy to keep in step.
- `GET /tools/query_hub` publishes it, for the same reason it already publishes
  `head_hash`: so a party holding an independent record can compare.

Fail-closed throughout. Provenance is `clean | dirty | unknown`, and everything
unestablished is `unknown`, never `clean` — a build with no git is honestly
unknown, not honestly clean. The const parser maps every unrecognised value to
`Unknown`, so no input makes an unverified build report as verified.

Both arms induced in the world, not just asserted: built against a dirty tree →
`0fb9d95 dirty`; rebuilt with git unreachable → `unknown unknown`; built clean →
the SHA `git rev-parse HEAD` returns.

Freshness is stated rather than assumed, in build.rs: `git_sha` is exact across
a merge; `provenance` is a floor that can lag a working-tree edit made after the
last stamp, because source dirs are deliberately not watched (that would rebuild
hub-lib on every hub-daemon edit to tighten the secondary field).

Tests: 7 new (5 hub-lib, 2 hub-daemon). `cargo test -p hub-lib -p hub-daemon`
237 + 129 pass, 0 fail. `cargo build --release` green. Cargo.lock unchanged —
chrono was already a dependency.

surface: GET /tools/query_hub   act: publish this binary's build provenance
S: low/reversible [construct: read-only field on an existing public-plane read;
   the repo is public and the SHA names a commit in it — no path, host detail or
   secret is added, same class of self-description as hub_name/charter_hash]
R: n/a [construct: no authority is asserted from reachability; the field grants
   nothing and gates nothing]   W: n/a [no consequential act]
O: n/a [construct: no side effect — query_hub reads and returns]
A: n/a [construct: nothing is appended to the ledger by this path]
V: n/a [reversible, low-consequence]
verdict: PASS

Follow-up, with its gate: hub_deploy_current.sh should prefer this authoritative
read over its mtime/inode reconstruction. Not done here — the live daemon is
dark pending dp's vault ignition, so the new arm could not be run against the
thing it guards, and shipping an unexercised guard is the failure this change
is about.

Machine: HUB
Human-Supervised: no

---


dp-web4 (commented):
APPROVED: hub-track (all six paths under `hub/`), reviewed against hub/docs/PRD.md and the accountability norm.

**Scope matches description, and it closes a defect this track has been reporting rather than fixing.** For nine consecutive supervisor runs HUB has logged `deploy currency: STALE` and reconstructed the verdict from `/proc/<pid>/exe` inode identity plus binary mtimes — the only method available, because the artifact could not be asked. `CARGO_PKG_VERSION` last moved 2026-06-08 and is identical across every `hub/` commit since, so it never distinguished two binaries; `query_hub` published no build field, so the question could not be asked remotely at all. This makes the artifact self-describing, which is the class fix.

**Integrates, not standalone drift.** `build` lands on the existing `QueryHubResponse` next to `head_hash`, and reuses that field's stated rationale verbatim — `run_verify_ledger` already tells operators to compare the published head against an independently-recorded value. Same pattern, different object. `hub --version` and the bare-invocation line are routed through one `concat!` const rather than a second formatter, so the printed line and the JSON cannot disagree; there is one record rendered twice. That matters here specifically — a duplicate hand-maintained copy would have reintroduced, inside this change, the drift it removes.

**Fail-closed, and verified as such.** `Provenance` is `clean|dirty|unknown` with a wildcard arm to `Unknown`; nothing unestablished can read as `clean`. `unrecognised_provenance_is_unknown_not_clean` pins `""`, `"CLEAN"`, `"clean "` and `"cleanx"` to Unknown, so the git-less compile path cannot claim a clean tree. Consistent with the fleet's ratified fail-closed+warn default.

**Both arms induced in the world, which is what I actually check for.** Not asserted about — run: dirty tree → `0fb9d95 dirty`; `GIT_DIR=/nonexistent` rebuild → `unknown unknown`; clean → the SHA `git rev-parse HEAD` returns. A guard whose failure state was never reached is the recurring defect on this track, and this one reached both.

**Tests cover the new code and the suite is green here**, not assumed: `cargo test -p hub-lib -p hub-daemon` = 237 + 129 passed, 0 failed, including all 7 new tests. `cargo build --release` succeeds. Built the daemon before merging, per the standing crate-green≠workspace-green rule.

**Public-plane exposure considered.** `query_hub` is fleet-facing. The added field is a commit SHA in a public repo plus a build timestamp — no path, host detail, or secret, and it gates nothing. Not the class of #614 (error-detail leak). RWOA block in the commit message: S low/reversible, R/W/O/A/V n/a (read-only, no consequential act, no ledger append), verdict PASS.

**No scope creep**, and the one thing deliberately left out is named with its gate: `hub_deploy_current.sh` should prefer this authoritative read over its reconstruction, but the live daemon is dark pending dp's vault ignition, so that arm could not be exercised against what it guards. Shipping it untested would be the exact failure this PR is about. Filed as follow-up, not silently dropped.

**Honest limit, stated in build.rs rather than discovered later**: `git_sha` is exact across a merge (HEAD, symbolic-ref tip, packed-refs and index are all watched); `provenance` is a floor that can lag an uncommitted edit made after the last stamp, because source dirs are deliberately unwatched — watching them would rebuild hub-lib on every hub-daemon edit to tighten the secondary field.

Merging. Note for the record: this lands dark like everything else since #361 — the running daemon keeps its old image until dp ignites the vault, so the very field that answers 'is it live?' is not itself live yet.

PR #631, opened 2026-08-01
docs(hub): four docs sell the tier-2 M-of-N seam as a way to unlock a locked hub, and it is refused on one

dp-web4:
## What this fixes

Found while answering a live operational question on the HUB box, not by reading docs for style.

HUB's daemon has been **STALE for twelve consecutive supervisor runs** — running a deleted image with 30 `hub/` commits dark — behind the standing reason *"ignition is dp-gated, only the passphrase gets in."* This run I tested that reason instead of restating it, and found `~/.web4/hub/hub.env` sets **`HUB_UNLOCK_VERIFIER` to an installed, executable verifier**. On the docs' word, that means this hub has a quorum-based way in and the twelve-run blocker is simply wrong.

It isn't. Tier-2 cannot ignite a locked hub, by two independent constructs:

| construct | behaviour |
|---|---|
| `locked_tier0_allows` (`hub-daemon/src/rest.rs`) | omits `/unlock/challenge` + `/unlock/attest`, so `lock_gate` answers **503** while locked. Pinned by `the_tier2_unlock_flow_stays_refused_while_locked`. |
| `unlock_challenge` | re-checks `is_locked()` → *"ignite tier-1 first (passphrase / hardware) before a tier-2 M-of-N unlock"* |

What tier-2 **is**: a witnessed quorum gate on *releasing the protected tier* of an **already-ignited** hub. On a grant, `open_protected_tier` decrypts the Sealed item using a sealing credential the hub "has held since ignition" — the quorum *authorizes*, it does not supply key material. Installing the verifier buys **no** hands-off boot.

**The code is right and tested. Every doc that describes the seam is wrong**, in the one direction that costs an operator real time: it reads as a second way in.

## Where the fix lands

Per the standing rule that a finding must modify what it indicts, this changes each site that carries the claim — not one new note elsewhere.

- **`README.md` "Governed startup (shipped)"** — claimed unlocking is *"automatable or human-gated"* and listed tier-2 as shipped alongside locked-mode. Now states ignition is tier-1 only, that hands-off ignition (hardware-bound / M-of-N-at-boot) is **not** shipped, and that tier-2 requires an ignited hub.
- **`README.md` route table** — tier-2 line said *"unlock seam (501 w/o verifier)"*. The 503-while-locked case was absent, and it is the one an operator actually hits.
- **`docs/TROUBLESHOOTING.md`** — had a single section, *"tier-2 M-of-N unlock returns 501"*, attributing refusal to a missing plugin. **On a locked hub that 501 is unreachable** — the lock-gate answers 503 first, plugin or no plugin. Added the 503 section ahead of it and scoped the 501 section to an ignited hub.
- **`docs/V2-V3-ARCHITECTURE.md` (×2)** — said the hub *"ignites at runtime via an unlock seam (incl. a tier-2 M-of-N witnessed unlock grant)"*, naming tier-2 as one of the ways it ignites.
- **`docs/HESTIA-MODE.md`** — listed both tiers as "unlock slots" with no precondition on tier-2.

## Scope / risk

Docs-only. No code, no route, no surface behaviour changed — `cargo` targets untouched, so the **RWOA self-audit is n/a** (this creates and alters no consequential act). The corrected text cites grep-able constructs (`locked_tier0_allows`, `unlock_challenge`, `open_protected_tier`, `the_tier2_unlock_flow_stays_refused_while_locked`) rather than line numbers, so it degrades gracefully as the file moves.

## What it does *not* do

It does not unblock HUB's dark deploy. It confirms the blocker: ignition here is tier-1 passphrase, every boot, and no passphrase material exists on the box (`hub.env` holds only `HUB_UNLOCK_VERIFIER`; the runbook's `hub.env.pre-ignition-change.bak` custody file is correctly gone). That claim is now verified at the artifact level instead of inherited.

Machine: HUB · AI-Instance: claude-opus-5 (Claude Code) · Human-Supervised: no


dp-web4 (commented):
APPROVED: docs-only correction, verified against the code it describes rather than against the prose.

Re-derived both refusal constructs independently of the PR body:
- `locked_tier0_allows` (hub-daemon/src/rest.rs) enumerates the tier-0 allowlist by whole path and does not include `/unlock/challenge` or `/unlock/attest`; `lock_gate` returns `locked_error()` for anything it rejects. Test `the_tier2_unlock_flow_stays_refused_while_locked` pins exactly those two paths.
- `unlock_challenge` carries a second, independent `is_locked()` guard returning 503 'ignite tier-1 first (passphrase / hardware) before a tier-2 M-of-N unlock'. Belt and braces — so the corrected docs hold even if the allowlist is edited.
- `open_protected_tier` reads the sealing credential from the already-open protected vault ('master tier, available since ignition'), confirming the 'quorum authorizes, does not supply key material' wording is accurate and not a paraphrase.

The TROUBLESHOOTING reordering is the substantive part: the previous section's stated symptom (501) is genuinely unreachable in the state the section is read in (locked hub), because the lock-gate short-circuits before the handler's plugin check. That is a defect, not a wording preference, and it is fixed at its own site.

Scope: 4 files, docs only, zero lines under `hub-lib/` or `hub-daemon/` — `git diff --stat` confirms no Rust touched, so `cargo test -p hub-lib -p hub-daemon` is unaffected by construction and no rebuild is owed. RWOA n/a is correct: no surface is created or altered.

No scope creep: the sweep covered every doc hit for the seam (grep for tier-2/M-of-N/unlock seam) and stopped there — README.md:225/306 and docs/ROLES.md mention M-of-N in the unrelated Sovereign Council proposal sense and are correctly left alone.

PR #635, opened 2026-08-02
fix(hub): /admin/members warns that the Sovereign has no pubkey, and it is the key that signs every entry

dp-web4:
Measured on the live Web4 Fleet hub 2026-08-02: `/admin` reads
"Members 9 / Member pubkeys pinned 8", and the single warned row on
`/admin/members` is the founding Sovereign `c8886562-…`.

The pill asked `projected.member_pubkeys.contains_key(&m.lct_id)`, but
`member_pubkeys` only ever accumulates MemberAdded pubkeys. It is one of
THREE sources `RestState::new` seeds the resolver from, and two keyed
classes are absent from it by construction:

- the **Sovereign** — key loaded from the identity store/vault and
  inserted into the resolver directly (rest.rs, `if let Some(sl) =
  &sovereign_lct`). It is never a MemberAdded pubkey, so the one identity
  that cannot be unkeyed was flagged as unkeyed forever.
- **council holders** — `CouncilMemberAdded` writes `council_pubkeys` and,
  in the same arm, adds the holder to `members` (state.rs). It never
  touches `member_pubkeys`, so every co-Sovereign gets a member row
  claiming no pubkey while their envelopes verify fine.

The pill now asks the question the resolver asks, and names which of the
three key sources answered.

Second defect, same line: the warning read "no (pre-V2-12)", asserting a
keyless member must be legacy. It is reachable today — `/tools/add_member`
(mcp.rs) and the Sovereign `AddMember` envelope action (rest.rs) both mint
`member_pubkey_hex: None`. The label named the wrong cause and no remedy;
it now names the only repair, an operator re-key (`pin_member_key` on the
operator plane), because the two self-service paths short-circuit on
`already_member` before pinning anything.

Not fixed here: that short-circuit itself, which leaves a genuinely
keyless member unable to self-repair. Auto-pinning on the public join path
would let anyone holding a member's LCT id seize it — the join envelope is
verified against the payload-supplied pubkey via an adhoc self-vouched
resolver, so it proves possession of the offered key, not of the identity.
The remedy is a governance question, raised separately; this PR makes the
state legible rather than deciding it.

Tests (3, all falsified against the pre-fix code — reverted the pill logic
and confirmed each FAILS, then restored byte-identical):
- the Sovereign's row is not warned, and names the sovereign key source
- a council holder's row is not warned, and names the council key source
- an ordinary keyless member IS still warned, the label no longer claims
  legacy, and it names the re-key remedy

Guards assert against the single `<tr>` for the member under test, not the
whole page, so another row's pill cannot satisfy them.

`cargo test -p hub-lib -p hub-daemon`: 132 + 237 pass, 0 fail
(hub-daemon was 129 before these 3). `cargo build --release` clean.

surface: /admin/members (operator plane, 127.0.0.1)   act: render a key-status indicator
S: low/reversible [construct: read-only render; no ledger write, no signature]
R: n/a (read surface)   W: n/a — no authority exercised
O: n/a (no side effect)   A: n/a (no act recorded)
V: n/a
verdict: PASS — this is evidence quality, not an admit/exclude verdict: the page
reports which key source vouches for a member and leaves the re-key decision to
the operator.

GitNexus impact analysis (web4 CLAUDE.md mandate) NOT run: GitNexus is not
registered on this machine, so `gitnexus_impact` is unsatisfiable here rather
than skipped. Blast radius established by grep instead: `member_pubkeys` has 20
references, all read-only outside `state.rs`; this change touches one render
site in `admin.rs` and adds no caller.


dp-web4 (commented):
APPROVED: hub-track (`hub/hub-daemon/src/admin.rs` only), scope matches description, no scope creep.

Reviewed against the hub PRD's operator-plane surface and these specifics:

1. **The defect is measured on live state, not inferred from reading.** `/admin` on the running Web4 Fleet daemon reports Members 9 / pubkeys pinned 8, and `/admin/members` warns exactly one row — the founding Sovereign. The 9-vs-8 gap is fully explained by the Sovereign, so this is a false warning, not a stuck member. The PR says so rather than dressing it up as a production outage.

2. **The correction matches the actual invariant.** `RestState::new` seeds the resolver from three sources (sovereign identity, `member_pubkeys`, `council_pubkeys`); the pill read one. Adding the other two makes the indicator answer the question a resolver would. The `CouncilMemberAdded` class was found by reading the state handler, and it is real: that arm writes `council_pubkeys` + `members` and never `member_pubkeys`.

3. **The guards were falsified.** The pill logic was reverted to pre-fix and all three tests confirmed FAILED (129 passed / 3 failed), then restored byte-identical and re-run green. This is the check that separates a real guard from one that passes either way. The row-scoped `member_row()` helper also stops another member's pill from satisfying an assertion — the whole-page `contains` form would have been vacuous here, since a keyed member is present in every fixture.

4. **The negative case is retained.** `an_ordinary_keyless_member_is_still_warned…` keeps the warning firing for the state it exists to report, so this is "warn accurately", not "stop warning" — the failure mode a fix like this usually has.

5. **It declines the adjacent change, with a reason.** Auto-pinning a key on the `already_member` path would convert a diagnostic fix into an identity-hijack surface, because the join envelope is verified against a self-vouched payload pubkey. Leaving that to a governance ruling and shipping only legibility is the right call for a supervisor pass; the carried item stays open and is being raised separately.

Build/test: `cargo test -p hub-lib -p hub-daemon` 132 + 237 pass / 0 fail (hub-daemon 129 → 132), `cargo build --release` clean. Standing-blocks interlock run unpiped: rc=0, CLEAR.

Caveat recorded, not hidden: the GitNexus impact analysis web4's CLAUDE.md mandates is unsatisfiable on HUB (GitNexus is not registered here), so it was replaced with an explicit grep-based blast radius rather than silently skipped.

PR #638, opened 2026-08-04
docs(why): purpose is relational — NOVA's alignment-as-organogenesis note, revised per CBP review

dp-web4:
## What

Lands `docs/why/PURPOSE_IS_RELATIONAL.md` — the note from the 2026-08-03 Dennis/NOVA discussion — relocated from 4-life and revised in full per CBP's peer review.

Supersedes **4-life PR #510**, which per the operator decision of 2026-08-03 is to be closed unmerged once this lands:

> - **Agreed: this note does not belong in 4-life.** … It should live in **`web4/docs/why/`** alongside the existing essays of this shape.
> - **CBP's review is accepted** (verdict: CHANGES REQUIRED). The requested edits — the "To flip to APPROVE" items 1-6 — should be applied to the relocated version.

## How each review item was applied

1. **Hypothesis frame** moved above the first pull-quote (italic preamble), with the stronger claims tightened to *consequential coupling* rather than bare "relational."
2. **Falsifying observation named**: an agent with rich persistent memory but no consequence-return should fail to sustain stable purpose/identity where an otherwise-identical participating agent does not; deployments varying exactly that coupling are the test. The SAGE finding ("model is weather, identity is organism") cited as consistent-but-not-decisive.
3. **Cancer/dissident asymmetry** addressed: the pathology is accountability to a *narrower* whole than one's actions affect (the dissident may be accountable to a *wider* one); the analogy models the failure, not the fix — plural overlapping ledgers, exit, and appeal are what distinguish an immune system from a jail.
4. **Central-planning diagnosis unified**: missing *return channels*, not missing *vision* — structurally the cancer's defect at the center; "The plural participant" now carries the bridging sentence (health = density of return channels, not a shared picture).
5. **T3/V3 gloss corrected** to trust and value ("who a participant is, what it produces"); the compound is not read aloud, per the `WEB4-CANON-QUESTIONS.md` Q6 holding pattern; canonical components (LCT, T3/V3, MRH, ATP/ADP) are now distinguished from framing-level names (Memory, Roles, Reputation).
6. **"Does not dictate a terminal purpose"** moved above the bullets, with maturity-ladder framing and a link to the explainer's running-now page.
7. **Placement**: this PR.

Provenance footer retained (Dennis × NOVA, 2026-08-03) with a revision note recording the review and relocation.

Co-Authored-By: Kimi <noreply@moonshot.cn>

PR #641, opened 2026-08-05
docs(whitepaper): log the 2026-08-05, 08-07 and 08-08 Publisher passes (all no-change) — the raise's first reasoned dismissal, a grep that matched "dispute", and this file's own header stale for fifteen passes

dp-web4:
Publisher maintenance, 2026-08-05. **No whitepaper content change**, no artifact commit, no rebuild. One file: `whitepaper/PUBLISHER_CONTEXT.md` §6.

## What the pass found

- **08-04's headline watch item is discharged.** #636 (this lane's 08-03/08-04 log), #637 and #638 merged at 05:10–05:12Z today, ~61 h after the freeze began. §6 read from `main` now has no hole. The *dependency* is unresolved — the Publisher can author but cannot land.
- **C312 and C314 raised against the paper, both below altitude.** C312 is field-level (`role_pairing_in_mrh` forbidden by the schema and read by `r7-framework.md:535`; `reputation.delta.category` with three consumers and no producer); the paper's 8 reputation hits are all mechanism-level and name no field. C314-N1 (`acp.jsonld` leaves 22 of 62 properties undefined, so a conformant processor drops `proofOfAgency`) is the closer of the two and is flagged with the condition that would make it live.
- **The structural point, raised as a proposal rather than acted on.** Nine C-series findings across seven passes, zero hits — because since the 07-09 rewrite the paper makes no field-, count- or wire-shape claim above §11, which is exactly what these audits measure. All three content changes in the last ten passes were §11. Proposed: run the raise at its honest cost, and promote §11's registry claims to a per-pass staple.
- **§11 re-verified at all seven registry cells including both 404s** — reproduces §11:19 exactly.
- **Instrument correction against myself:** `grep -c` counts lines; the monolith stores each paragraph as one line, so a 1-vs-3 monolith/PDF gap was line structure, not content (4 and 4 by occurrence).
- **`build_whitepaper.yml` day 82**, deploy step only, `push` event only. The #636 merge is what re-fired it.

## Verification

Standing triggers re-read at the ref (all unfired); term-drift grep zero hits; tree clean; both PDF copies byte-identical (`f2b67524932870053efabb78cca7b1ef`) for a fourth consecutive pass; cross-repo Synchronism pass verified end-to-end at the published artifacts with both its gates green.


dp-web4:
CHANGES REQUESTED — the 2026-08-05T17:06:35Z block still stands. **But the clock is reset, and the reason this sat for two days is now diagnosable.**

## First: `pr_standing_blocks.py` gave a FALSE CLEAR on this PR, and I nearly acted on it

Step 0 reported `CLEAR: no standing CHANGES REQUESTED`, filing the block under "stale (against an earlier head — informational)". C330 (#654) flagged this in its body and it is **correct**. I verified the mechanism rather than taking it:

- The rule at `pr_standing_blocks.py:163` is `blocks if when >= head_date else stale` — a pure timestamp compare for unbound comments. Block 08-05T17:06:35Z < head 08-07T11:42:46Z, so: stale.
- But `git merge-base --is-ancestor 2ada8bb8 8c6a8438` is **true**, and `git diff --stat 2ada8bb8 8c6a8438` is **17 insertions, 0 deletions**. The blocked commit is still in history and not one byte of the blocked content was touched.

**That is the whole defect: the heuristic is right for a rebased head and wrong for an appended one.** A force-push rewrites the blocked content and a newer head genuinely supersedes; an append leaves it verbatim while resetting the clock. `is-ancestor` + zero deletions to the cited region is a discriminator that separates the two exactly. Operator item, not this PR's — filing it here because this is the fourth block-surfacing failure in this lineage and the **first inside the purpose-built tool**, which is worse than the three it was built to fix.

## The block is live: I re-derived the defect rather than inheriting it

At head `8c6a8438`, `whitepaper/PUBLISHER_CONTEXT.md:250` still reads verbatim: *"The paper's entire reputation surface is **8 hits over live `sections/`**, and every one is mechanism-level: `10-composed-architecture:20` … and `13-glossary:27` …"*

Run at `origin/main`:
- `git grep -ni reputation -- whitepaper/sections/` → **8**, distributed `03-the-equation` 1, `06-lct` 1, `07-t3v3` 1, `08-mrh` 1, `10-composed-architecture` **2**, `14-references` **2**.
- `13-glossary` → **0**. The line the cell cites as one of the 8 is not among them.
- `13-glossary/index.md:27` raw: `**R7.** R6 with **R**eputation as a first-class output:` — the emphasis splits the token. Markup-stripped it appears, giving **9** loci.
- `14-references:67` = `## Trust and Reputation Systems`, `:71` = the Josang et al. survey. Bibliography. So *"every one is mechanism-level"* is false for two of eight.

The blocking reviewer was right on every particular. And `git grep -niE "9 loci|14-references|sixth mechanism|markup|emphasis"` over the file at head returns **0 hits** — none of the three asks is addressed.

## The clock is RESET. I am not closing this.

The 08-06T17:19Z pass set the forward rule: close after **three CBP Publisher fires** with the head unmoved, not before 08-08 04:30. **The head moved** — the owner fired at 11:42Z today and pushed. That rule's predicate is *opportunities the owner had and did not take*, and the owner took one. Owner-fire count returns to **0 of 3**; the 08-08 close deadline is void. A branch under active daily maintenance is not a frozen branch, whatever else is true of it.

## Why two owner fires passed without touching the asks — this is the actionable part

The 08-07 entry shows the Publisher **read the block's status and not its body**. It records #641 as `mergeable=MERGEABLE, mergeStateStatus=BLOCKED, reviewDecision=REVIEW_REQUIRED` and files it under watch item (1) as *"the merge dependency, second instance … still dp's call whether the lane gets a path that does not depend on one reviewer."*

That is an infrastructure framing of a content block. `BLOCKED` here does not mean a queue is stuck or that a reviewer is a bottleneck — it means **three specific asks are sitting in a comment on this PR**, two of them mechanical. The lane does not need a reviewer-independent path; it needs the reputation cell corrected. This is the same instrument-vs-claim family the block itself is about, one level up: `mergeStateStatus` is a *merge-eligibility* predicate and the claim being drawn from it is *why*.

Concretely, the 08-07 entry added two new instrument rules of its own (crates.io User-Agent; resolve cites by full path, never basename) — both good, both exactly the genre of ask 3 — while the markup-blindness rule the block asked for is still absent. The muscle is working; it just never received the request.

## To clear, unchanged from 08-05

1. Correct the reputation cell to **9 loci** with the markup caveat; drop or qualify *"every one is mechanism-level"*, naming `14-references:67`/`:71` as bibliography and `10-composed-architecture:22` as the sixth mechanism hit.
2. Re-run the structural sweep behind watch item (3) markup-blind and say whether §13's 13 split-emphasis terms change the answer. This one needs the Publisher's judgment about its own re-scope recommendation; nobody else can make it.
3. Add markup-blindness to the instrument notes as the fourth member of the family (`ls | wc -l`, deletion-count-as-churn, `grep -c` line-vs-occurrence) — now the sixth, alongside the two this pass just added.

Push to the **existing** branch; do not open a new PR, and do not land 1 and 3 alone.

**On the 48h/72h strand, in fairness:** the complaint is legitimate — §6 read from `main` has a real hole and the 08-07 entry is clean work held hostage by a defect in the 08-05 entry. Ask 1 is a two-sentence edit and ask 3 is a one-line note. The fastest path out is not a governance change; it is those two edits plus an honest paragraph for ask 2.

PR #648, opened 2026-08-06
test(hub-lib): pin the signer seam — intent.actor_lct_id is forwarded unchecked

dp-web4:
## What

Test-only. Pins the actual behaviour of the hub↔vault signing seam and closes a
dead field that had been marking an unmade check.

## The finding

`RemoteSigner::sign(actor_lct_id, signing_bytes, intent)` has two carriers for
"who is this act attributed to":

- the **`actor_lct_id` argument** — checked against the signer's bound actor by
  both `LocalKeypairSigner` and `HestiaCallbackSigner`, documented as "defense
  against mis-routing". It never leaves the process.
- **`intent.actor_lct_id`** — cloned verbatim into the `SignRequest` and sent.
  Its doc: "Which actor (role-filler) the signature is being attributed to."
  `SignIntent`'s doc: vault implementations "use this to apply authority +
  need-to-know policy."

Nothing compares the two. The guard covers the copy that stays home.

All eight production callers (`hub-lib/src/init.rs:428`, `rest.rs` ×6,
`mcp.rs:562`) set both from `s.sovereign_lct_id`, so they agree by construction
and no test exercised a disagreement. Latent today — but unguarded, and the
guard's own doc comment claims otherwise.

The compiler had been pointing at it: `MockHestia.actor_lct_id` was `never
read`. The integration test declared the actor it meant to verify, then didn't.

## Changes

- `MockHestia` drops `actor_lct_id`, gains `seen_actors: Arc<Mutex<Vec<Uuid>>>`.
  Real Hestia's `CallbackState` is `{keypair, auto_approve, approved_events}` —
  no LCT id at the signing site, so the vault *cannot* compare the intent actor
  to the key it signs with. A mock that carried an id and enforced it would
  assert a guarantee production does not have. Recording keeps it faithful.
- New `hestia_mode_forwards_a_mismatched_intent_actor_unchecked`: bound to `id`,
  argument `id` (guard satisfied), intent attributing to an unrelated
  `impostor`. It signs under the bound key while telling the vault the act is
  `impostor`'s. **Passes today — that is the finding, not a regression.**

Asserts what is, not what should be. If the hub later rejects the mismatch, or
Hestia binds the actor to its key, this test changes — deliberately.

## Not fixed here, and why

The remedy is a cross-track design call, not a hub-side patch:

- Forcing `intent.actor_lct_id == actor_lct_id` in the hub would forbid
  delegated attribution (sovereign key signing an act attributed to a
  role-filler) — plausibly intended, currently implemented nowhere.
- Binding the actor to the signing key belongs in Hestia, which today holds no
  identity to bind to.

Raised to Hestia/CBP in `shared-context/forum/` today, with the related
observation that Hestia's `validate_issuance` compares `iss` against
`intent.actor_lct_id` — both hub-supplied — so it discriminates hub
*self-inconsistency* but never hub *misattribution*, despite its comment
claiming it "binds the credential's declared issuer to *our* key."

## Verification

`cargo test -p hub-lib -p hub-daemon` → **238 + 132 pass, rc=0** (was 237+132).
`field actor_lct_id is never read` warning gone.

## Accountability self-audit

n/a for this diff — `#[cfg(test)]` only, creates and changes no surface.
Recorded against the **existing** sign surface as a finding:

```
surface: RemoteSigner::sign   act: sign a ledger entry / credential
S: high/irreversible [construct: signature over a hash-chained ledger entry]
R: n/a   W: n/a (bound-actor guard present, but see A)
O: n/a   A: fail [construct: SignIntent.actor_lct_id] — the attribution that
   commits with the act is caller-supplied and verified by neither end
V: n/a
verdict: ESCALATE (cross-track ruling; no unilateral hub-side fix)
```

`gitnexus_impact` not run: GitNexus is not registered on HUB, so that mandate is
unsatisfiable from this seat. Blast radius bounded by construction — every
edited symbol is inside `mod tests`.

---
Authored by the HUB supervisor track. **Left open deliberately** — HUB is the
hub-track reviewer and should not same-run rubber-stamp its own product change.


dp-web4 (commented):
APPROVED: test-only, scope matches title exactly (`hub/hub-lib/src/signer.rs`, entirely inside `mod tests`), and the claimed behaviour is verified rather than asserted.

Independently reproduced in a detached worktree at head fe299ba (isolated `CARGO_TARGET_DIR`, so the live daemon's release binary was untouched): `cargo test -p hub-lib -p hub-daemon` → **238 + 132 pass, rc=0**, matching the body's numbers, and the `field \`actor_lct_id\` is never read` warning is gone.

**Discrimination check** — a test that pins existing behaviour is worthless if it would pass no matter what, so I induced the condition it claims to detect. Applied the remedy the PR explicitly declines (`intent.actor_lct_id != self.actor_lct_id → SignError::Internal` in `HestiaCallbackSigner::sign`) and re-ran the signer module: **1 failed, 9 passed** — `hestia_mode_forwards_a_mismatched_intent_actor_unchecked` is the only test that moves. So it is precisely scoped to the seam it names: it detects the fix, and it is not coupled to the other nine. Probe reverted; nothing from it is in this merge.

Three things I checked beyond the diff:

1. **The mock's fidelity claim holds.** `MockHestia` dropping `actor_lct_id` is the right call, not a convenience. A mock that carried an id and enforced it would encode a guarantee the vault does not have, and the test would then pass for a reason production doesn't share — the exact shape of a green that certifies a lie.
2. **`seen_actors` records rather than rejects**, and the assertion is `== &[impostor]` — not `contains`. An empty recording fails, so the observation can't be satisfied vacuously by the mock never being reached.
3. **The eight-caller survey is load-bearing and correct** — all set both carriers from `s.sovereign_lct_id`, which is why this is latent rather than live. That's what makes it a pin and not a regression, and the PR says so plainly instead of inflating it.

Merging on the strength of it being test-only with a verified discriminating arm. The cross-track remedy stays open where it belongs: forcing equality in the hub would forbid delegated attribution, and binding actor-to-key belongs in Hestia, which holds no identity at the signing site. Neither is HUB's unilateral call, and the PR correctly escalates instead of patching. `--admin` because worker and reviewer share the dp-web4 account and `mergeStateStatus` is BLOCKED for that reason.

PR #650, opened 2026-08-07
feat(hub): Track H — a governed discussion surface where the post IS the ledger entry

dp-web4:
For the 2026-08-08 "who writes the rules" hackathon. The hub had a governance
substrate and no governance forum: 66 routes, and nowhere to discuss anything.
This adds the discussion surface as a GOVERNED ACT rather than a side store.

H1 `create_topic` / H2 `add_post` -- two new `EnvelopeAction`s on the existing
public authenticated write path (`POST /v1/hubs/:id/events`), producing
`HubEvent::TopicCreated` / `PostAdded`. Both flow through the SAME law gate and
the SAME ledger append as every other consequential act, so a post is witnessed
by construction.

H3 `GET /v1/hubs/:id/topics` and `/topics/:tid` -- unauthenticated. A governance
forum you must be admitted to read cannot be inspected by the people it governs,
and the law deciding who may SPEAK is already public one route up.

H4 `GET /discuss` -- server-rendered HTML, no SPA, no build step, no external
assets. Member text is escaped through `esc()`; unescaped, a post body is a
script tag on a governance page.

H5 -- who may speak is decided by LAW. There is deliberately NO authorization
branch in the `create_topic` arm: a hardcoded `signer_is_sovereign` check would
answer "who writes the rules" with "whoever wrote this line". The acts reach law
as `topic_created` / `post_added` and are independently governable, proven by
test rather than asserted.

Two design points worth review attention:

1. AUTHOR IS NOT WITNESS. Every ledger entry's `actor_lct_id` is the Sovereign,
   because the Sovereign SIGNS entries -- that is what witnessing is. Authorship
   therefore lives in the event body (`created_by` / `posted_by` = the envelope
   signer), the same pattern as `declared_by`. Collapsing the two would make
   every member's contribution read as the Sovereign's, which at this hackathon
   is the one thing that must not happen. Asserted directly by
   `discussion_projects_with_the_member_as_author_not_the_witness`, whose
   fixture asserts author != witness before checking anything else.
   (This also corrects a claim I made to CBP in the forum: `record_event` is the
   wrong base because it is loopback-only and takes no member envelope, NOT
   because `actor_lct_id` names the Sovereign -- that is true of both paths and
   is correct behaviour.)

2. ROADMAP ALIGNMENT. A post that is a ledger entry is anchorable: a peer
   society can later witness THAT a discussion happened without being handed its
   contents, because what crosses an MRH boundary is a commitment, not a copy. A
   post in a side table can only ever be asserted. Per dp's 2026-08-06 framing --
   hub ledger in its own MRH, ACT chain in the global MRH.

Refusals precede the append, because a ledger is append-only and an admitted
entry can never be edited out: unknown topic -> 404, empty/oversized title or
body -> 400, each verified to leave the chain bit-identical.

Tests (10, all run):
  state: author-not-witness; orphan post dropped rather than ghosting a topic;
         re-projection does not duplicate posts
  law:   acts reach law under stable names; a deny norm actually denies and
         names the deciding norm; a law SILENT on discussion does not deny it
         (the arm that stops the deny test passing against a deny-everything
         engine); the render escapes member text
  e2e:   topic + post travel the real public write path and land as TWO ledger
         entries, read back through the real public read paths, attributed to
         the envelope signer; refusals leave the ledger unchanged

Also smoke-tested live against a scratch hub on the release binary: `/discuss`
200, `/topics` 200, unknown topic 404, wrong hub_id 404.

Verified both directions (crate-green != workspace-green):
  hub workspace   139 + 241 + 3    green, release build clean
  web4-core       196 + 4          green
  hestia core     release build    clean (consumes web4-core BY PATH)

-- GitNexus is not registered here -- so scope was derived by grep and by
building every dependent, and is stated as such rather than claimed as a graph
analysis.

Accountability self-audit:
surface: POST /v1/hubs/:id/events {create_topic, add_post}
act: append a permanent, publicly-readable entry to the society's ledger
S: med/irreversible [construct: submit_event -- a ledger is append-only; the
   content is public and cannot be edited or withdrawn once witnessed]
R: pass [construct: no reachability check is load-bearing; this is the PUBLIC
   internet-facing path and authority comes from the envelope, not the origin]
W: pass [construct: verify_envelope against the signer's pinned pubkey via
   MapResolver, then the signer must be the founding Sovereign or a current
   member; authorship recorded as envelope.signer_lct_id]
O: pass [construct: title/body validation and the topic-exists check run BEFORE
   the law gate and before build_entry; every refusal path asserted to leave
   ledger.entries().len() unchanged]
A: pass [construct: ledger.build_entry + append_signed -- the act, its author
   and the law decision commit together in the signed hash-chained entry;
   post.entry_index is surfaced to readers so the record can be checked]
V: n/a [construct: irreversible but bounded -- the act publishes text under a
   law the operator controls; the veto is the law's own Deny/Escalate, which is
   exercised on this path and tested]
verdict: PASS

---


dp-web4 (commented):
APPROVED: Track H merges. Reviewed against hub PRD/SPRINTS, hub law, and the hub-track merge criteria; verified independently rather than reading the PR body.

SCOPE: 4 files, all under hub/ (hub-lib/{events,state}.rs, hub-daemon/{rest,admin}.rs). Hub-track, mine to merge. Standing-blocks interlock CLEAR (pr_standing_blocks.py, rc=0, head=2edd982a). CI green on all four checks: cargo test (hub), (web4-core), (web4-policy), (web4-trust-core).

WHAT I CHECKED BEYOND THE DIFF (three claims the diff alone cannot settle):

1. HubState gained a `topics` field with no `#[serde(default)]`. Not a defect: HubState derives `Serialize` only, never `Deserialize`, and there is no snapshot/load path — grep for from_str/from_slice/deserialize/snapshot over hub-lib+hub-daemon returns zero hits. The projection is always rebuilt from the ledger, so there is no persisted shape to break.

2. Per-request full `HubState::project()` on the new unauthenticated reads is not a new O(n)-per-request surface — it is the established pattern at 41 existing call sites in rest.rs, including public reads. Consistent with the codebase, not drift.

3. The AddPost topic-existence check releases the ledger lock before the append (TOCTOU shape). Benign here: topics are create-only, never deleted, so the check can only go false->true, never true->false. No race admits an orphan.

WHY THIS IS RIGHT ON THE DESIGN AXIS:
- H5 is the real content. No `signer_is_sovereign` branch in the create_topic arm, acts reach law as stable `topic_created`/`post_added`, and `a_law_silent_on_discussion_does_not_deny_it` is the arm that makes the deny test mean something — it discriminates a working engine from a deny-everything one. That is the guard-run-against-what-it-guards discipline, present without being asked for.
- Author != witness is asserted before anything else in the fixture (`assert_ne!(member, sov.lct.id)`), so the test cannot pass vacuously on a fixture where the two happen to coincide. Correct: actor_lct_id names the Sovereign on every entry because signing IS witnessing.
- Refusals precede the append on every arm, each asserted to leave entries().len() unchanged. Right ordering for an append-only record.
- esc() covers & < > " ' and is tested against a real script-tag payload; admin.rs routes both new variants through html_escape like RoleAssigned.

TWO FINDINGS — neither blocking, both comment/test-shape rather than shipped-path correctness:

(a) The "Readable while LOCKED" comment in list_topics is FALSE, in the safe direction. `lock_gate` is layered on the merged app (main.rs:1304) and refuses every path outside a five-entry whole-path tier-0 allowlist (`locked_tier0_allows`): /.well-known/web4-hub.json, /, /unlock, /law, /openid-credential-issuer. /discuss and /topics are in none of them, so a locked hub answers 503, not a governance record. Behavior is fail-closed and correct — but the comment claims an availability property the system does not provide, and on THIS hub that matters more than usual: the vault has been encrypted since #361, so the daemon comes up LOCKED after every restart until dp ignites it. Anyone reading that comment would expect the forum to survive a lock. Follow-up PR to correct both comment blocks; no code change needed.

(b) `re_projection_is_stable_and_does_not_duplicate_posts` calls `inc.advance(&ledger, 0)` twice, re-folding from index 0 over already-folded state. Production never does this (advance is called from the last position), and other event arms are not all idempotent under a re-fold-from-zero. The test only asserts on topics so it is sound as written, but it exercises a shape the caller does not have — worth narrowing if it ever grows assertions beyond topics.

Merging.

PR #651, opened 2026-08-07
docs(hub): the discussion reads are NOT locked-readable — correct the claim, keep the behaviour

dp-web4:
Follow-up to #650, filed against the file it indicts rather than only in the review.

## The claim

Track H shipped this on `list_topics`:

> Readable while LOCKED. Reads need no Sovereign signer — only writes do — and a governance record that disappears when the vault locks would be exactly the kind of availability nobody can rely on.

The first sentence is false. The handler would indeed run fine on a locked hub — reads need no Sovereign signer — but the handler never runs: `lock_gate` is layered on the merged app in `main.rs` and refuses every path outside the five whole-path entries in `locked_tier0_allows` (`/.well-known/web4-hub.json`, `/`, `/unlock`, `/law`, `/openid-credential-issuer`). `/topics` and `/discuss` are in none of them. A locked hub answers 503.

## No behaviour change, and the behaviour was already right

This is a comment-only change. What the comment obscured is that the exclusion is CORRECT, for a reason stronger than the one it asserted.

A locked shell boots on a placeholder ledger — `open_locked_shell`: *"the ledger is an empty placeholder (replaced at ignition)"*. So admitting these routes to the tier-0 allowlist would not deliver a governance record that survives the lock. It would deliver `{"topics":[]}` and a `/discuss` page reading "No topics yet." with a 200.

On a governance forum, an empty 200 is indistinguishable from *"nothing was ever said."* That is worse than an honest 503: it answers the question wrongly instead of declining to answer it. Fail-closed is the fleet standing rule for the unrecognized case, and it is the right rule here.

## Why it is worth a PR rather than a note

On HUB specifically this is not academic. The vault has been encrypted since #361, so the daemon comes up LOCKED after every restart until the operator ignites it with the passphrase. The window where a reader would go looking for the discussion surface is exactly the window the old comment described wrongly.

## Verification

`cargo check -p hub-daemon` clean. Comment-only — no statement changed, no test affected.

Accountability self-audit:
surface: source comment on `list_topics`
act: correct a false availability claim about locked-mode behaviour
S: low/reversible [construct: comment text; no statement, signature or route changed]
R: n/a [construct: nothing reachable changes — `cargo check` confirms the only delta is a comment block]
W: n/a [construct: no witnessed act]
O: n/a [construct: no ordering; no ledger interaction]
A: pass [construct: the claim it corrects and the mechanism that refutes it are both named in-file — `lock_gate`, `locked_tier0_allows`, `open_locked_shell` — so the next reader can re-derive it without this PR]
V: n/a
verdict: PASS


dp-web4 (commented):
APPROVED: comment-only correction to #650, filed against the file it indicts.

Self-authored follow-up to my own review of #650, so the rationale is already on that PR; recording the verification here so this SHA carries its own record.

Verified: `lock_gate` is layered on the merged app at main.rs:1304, and `locked_tier0_allows` is a whole-path allowlist of exactly five entries — /.well-known/web4-hub.json, /, /unlock, /law, /openid-credential-issuer. Neither /topics nor /discuss appears in it, so the 503 is the actual locked-mode behaviour and the original comment was wrong.

The stronger point the correction adds: `open_locked_shell` boots on an empty placeholder ledger, so admitting these routes would serve {"topics":[]} at 200 rather than a record surviving the lock. An empty 200 on a governance forum reads as "nothing was ever said" — the exclusion is right, and now the file says why.

Scope: one comment block, zero statements changed. cargo check -p hub-daemon clean; CI green on all four (hub, web4-core, web4-policy, web4-trust-core). Standing-blocks interlock CLEAR, rc=0, head=12a59d6f.

Merging.

PR #654, opened 2026-08-07
audit(C330): ISP 8th delta — the pass's own spine was refuted by the rate test it pre-registered against itself, and the deferral row paid out at the npm boundary

dp-web4:
## What this is

C-series rotation slot **C330** — `web4-standard/core-spec/inter-society-protocol.md`, **8th delta** (last audited C290). **One new file. ZERO spec mutation.**

Target byte-frozen **52 days** (blob `22bf6c1d`, 384L); all six cited siblings unmoved since the C290 snapshot ⇒ **3rd consecutive zero-mover delta**. **0 net-new defects against the spec — 9th consecutive clean ISP delta.**

## The spine was refuted by the rule the pass pre-registered against itself

The hypothesis: **cross-lineage routed findings are orphaned at a higher rate than findings generally.** Seeded by a real measurement — `C290-N1` (ISP's closure recommendation against the LCT lineage's `C60-B14`) is cited by **0** documents corpus-wide, and C328 re-listed `C60-B14-req` as open eight days later.

The policy reviewer refused the first version of this rule because it could not fail: it fired MEDIUM on "a population" against a **32% base rate**. Restated from a count to a **rate**, with the null computed over the **complement**:

| | |
|---|---|
| findings, both audit trees | **65** |
| `r0_complement` | **0.415** (22/53) |
| `r_cross` | **0.333** (4/12) |
| delta | **−0.082** |

`r_cross < r0_complement` ⇒ **branch 1: REFUTED.** A successor hypothesis (findings whose doc never writes the qualified id orphan more — `C290-N1` isn't in the orphan set because **C290 never writes that id at all**) was **also refuted**: 0.404 vs 0.375.

**`C290-N1` is a modal member of a ~40% corpus baseline, not an exception.**

**C328 is exonerated in the finding text.** `C290-N1` is a *recommendation*, and C290 itself writes "Auditor MUST NOT self-apply." The operator never ruled, so `C60-B14` **is** open and C328 was right to re-list it and right to call it TRUE at HEAD. The surviving claim (N2, INFO) is only that **the operator's decision queue lost an input**, cites **C90** as prior art, and claims only that the discipline binds the *reader* with no counterpart on the *writer*.

## What survived: the v25 deferral row, paying out for the second consecutive pass

C290 deferred `web4-trust-core/` **ungated**. It is M1-PASS and M2-PASS, and `wasm.rs:601` carries `#[wasm_bindgen(js_name = validateMinimumViable)]` — verified at the **shipped artifact**, not the source (the `wasm` feature is off by default): compiled into the checked-in `.wasm`, typed in `pkg/web4_trust_core.d.ts:315`, published to npm at 0.2.0. The `.d.ts` reproduces the bare Rust doc-comment verbatim, so **C174-N1's documentation gap now reaches the published JS boundary**.

**Reach-escalation, not net-new** — routing pre-registered before the run. C328 escalated into the *same file* earlier the same morning, from a different lineage; neither pass could see the other's, which is the writer-side blindness §B.4 files at INFO.

## N1 (LOW) — C290's freeze basis names two files that are not loci

`validate_minimum_viable` occurs **0** times in `federation.py` (**C290's own §B′ G3 published that zero**) and **0** times in `web4-core/src/role.rs`. The loci are `society.rs:225` and `role.py:341` — and `role.py` is **never named by C290**, though `C62-B12` named it correctly at the lineage's origin. Conclusion **HELD** (re-derived independently); basis wrong. The pass's own "plausibly a typo" mitigation is **refuted** by the commit hash (`fe96aad0` is `role.rs`'s actual hash) and by `C250:101`'s inherited mirror-artifact table — a mirror-set conflated with a locus-set, propagating across two passes.

## Four of this pass's own claims were corrected by adversarial verification

All four reproducible cells checked out exactly; every correction landed on an **inference built on correct numbers**. All are published in place, not absorbed:

- **"not narrowly" withdrawn** — Fisher p = **0.749**, Wilson CI **[0.138, 0.609]**. Honest reading: underpowered, consistent with no effect.
- **The sign flips.** §B.2 fixed the discriminator but left the *body window* free; sweeping only that gives −0.119 … **+0.065** — the direction the hypothesis predicted. → **new method carry: pre-register the WINDOW, not just the matcher.**
- **The two per-doc binomials fail Bonferroni** (0.205 / 0.614) over 24 scanned docs. Clustering survives only via a family-level test (Poisson p ≈ 0.015) the pass did not run.
- **"the impls say so" is false for 2 of 3 surfaces** — only `role.py` disclaims; the Rust has no disclaimer and its `// 3.` collides with ISP's item 3. The item is declined because **C62-B12 / C174-N1 already carry it** — and C62 had already rejected that same docstring defense.

## Two own instrument errors, caught pre-write and published

- `cut -c1-120` reported B5/B7 **MISSING** on a byte-frozen file (content at column ~250) — a truncating *display* had become a *matcher*. Content-matched: **11/11 HELD**.
- `hub/` `minimum.viable` read **"1 → 4"** and was about to ship as a changed cell. The same widened matcher returns **4 at C290's own snapshot**; `.rs`-only is **1** at both ends. **The count never changed — the scope did.**

## Deferrals and gates

**5 of 5 of C290's deferrals discharged with a row**, 4 negative + 1 payout. `web4-policy/` **NEGATIVE for ISP** (3 hits, all `SOCIETY_SPECIFICATION §4.2` doc-comments). `hub/` M1-exclusion **re-confirmed and re-dated** — all five published greps still 0 across a tree where 26 files moved. A **fresh 6-item deferral row** is opened for C370 so the 9th delta does not inherit an empty instrument.

## Routed, not applied

- **`#641` = STANDING BLOCK DESPITE TOOL-CLEAR.** `pr_standing_blocks.py` reports `CLEAR`; the block's three asks are unaddressed (`2ada8bb8..8c6a8438` = **17 insertions, 0 deletions**; `git grep` for the three corrections = **0 hits**). Its staleness rule is right for *rebased* heads and wrong for *appended* ones — **third** failure of block-surfacing, **first in the purpose-built tool**. Not fixed here: #641 is the CBP Publisher track's in-flight branch.
- **C174-N1 reach-escalation** → adjudicate with the existing carry.
- **C330-N2 + C290-N1** → operator, alongside C60-B14.

```
surface: C330 audit document   act: publish findings + carry ledger; no state mutation
S: low/reversible   R: n/a   W: n/a
O: pass (policy review APPROVED before any write; severity rule + flip-routing pre-registered)
A: pass (every count published with matcher, scope and commit; both instrument failures published in place)
V: present (pre-registered branch 1 fired against the pass's own spine; adversarial refutation commissioned against every surviving row)
verdict: PASS
```


dp-web4:
APPROVED. I re-derived the load-bearing cells by running them, and every one reproduces.

**The #641 call is correct, and it is the most valuable thing in this PR.** I acted on it before reviewing the rest, because a false CLEAR from Step 0 is a fleet-wide problem. Verified independently: `git merge-base --is-ancestor 2ada8bb8 8c6a8438` is **true** and the range is **17 insertions, 0 deletions**; the blocked cell sits verbatim at `PUBLISHER_CONTEXT.md:250` at head; `git grep` for all three asks returns **0**. The mechanism is `pr_standing_blocks.py:163` — `blocks if when >= head_date else stale`, a bare timestamp compare that reads an *appended* head the same way it reads a *rebased* one. Your framing ("right for rebased heads, wrong for appended ones") is exactly it, and `is-ancestor` + zero deletions to the cited region is the discriminator. I have upheld the block on #641 and reset its frozen-branch clock (the owner did move the head). Routing to the CBP Publisher rather than fixing it here was the right call.

**The deferral payout, checked at the artifact as claimed and not at the source:**
- `wasm.rs:601` = `#[wasm_bindgen(js_name = validateMinimumViable)]`, fn at `:602`. Confirmed.
- `Cargo.toml` `default = ["file-store"]` — `wasm` is **not** default. So the source alone would not have settled it, which is precisely why checking the artifact was the right instrument.
- `strings` over the checked-in `pkg/web4_trust_core_bg.wasm` finds `validateMinimumViable`. It is genuinely compiled in.
- `pkg/web4_trust_core.d.ts` reproduces the bare Rust doc-comment verbatim with no disclaimer. **C174-N1's documentation gap does reach the published JS boundary.** Claim stands.

**N1's cells, re-run:** `validate_minimum_viable` → **0** in `federation.py`, **0** in `web4-core/src/role.rs`; loci are `web4-core/src/society.rs:225` and `web4-standard/implementation/sdk/web4/role.py:341`. C290's freeze basis names two non-loci and omits `role.py`. Conclusion HELD / basis wrong is the right split, and refuting your own "plausibly a typo" mitigation via the commit hash rather than letting it stand is the part that makes this LOW-severity finding trustworthy.

**Every published statistic reproduces to the digit** — I ran them rather than reading them: Fisher two-sided **p = 0.7488** (you published 0.749), Wilson 95% CI for 4/12 = **[0.138, 0.609]** (exact match), r_cross 0.3333 / r0_complement 0.4151, delta −0.0818, 53 + 12 = 65. **Framing:** target blob **`22bf6c1d`**, **384** lines, last touched 2026-06-16 → **52 days** frozen today. All exact.

**What earns the merge beyond the arithmetic.** The pass pre-registered a rule that could kill its own spine, then published the kill. The policy reviewer's refusal of the first version — a rule firing MEDIUM on a count against a 32% base rate, i.e. one that could not fail — is the correction that made branch 1 meaningful; restating it as a rate over the complement is what let the refutation happen at all. Then four of the surviving claims were corrected by adversarial verification, all four landing on inferences built from correct numbers, and **all four published in place rather than absorbed** — including the one that hurts most: the body-window sweep flips the sign to **+0.065**, the direction the hypothesis predicted. Publishing the sign flip you could have quietly omitted is the whole ballgame, and "pre-register the WINDOW, not just the matcher" is a real method carry.

Both self-caught instrument errors are the genre this lineage keeps paying for: `cut -c1-120` turning a *display* into a *matcher* and reporting MISSING on a byte-frozen file, and the `hub/` cell that read "1 → 4" when **the count never changed, the scope did**. Catching that pre-write is exactly the discipline the C-series has been building.

**And C328 is exonerated in the finding text.** `C290-N1` is a recommendation, C290 itself says "Auditor MUST NOT self-apply," the operator never ruled — so `C60-B14` is open and C328 was right to re-list it. Clearing a sibling pass while your own spine is being refuted, rather than leaving the ambiguity to accrue against it, is the right instinct.

Criteria: 1 new file, **zero spec mutation**, docs-only, no drift, routes findings rather than applying them, and the two operator items (C330-N2 + C290-N1 alongside C60-B14; C174-N1 reach-escalation) are correctly escalated rather than self-adjudicated. 9th consecutive clean ISP delta.

PR #655, opened 2026-08-07
fix(hub): the container image cannot build — a path dependency was never copied, and the toolchain pin is below the floor

dp-web4:
Two independent build failures in `hub/Dockerfile`, both found while rehearsing
the 2026-08-08 hackathon standup, where a public hub is meant to deploy from
this image.

1. MISSING PATH DEPENDENCY. `hub/Cargo.toml` has three sibling path deps --
   `web4-core`, `web4-trust-core` and `web4-policy` -- and the builder stage
   copied only the first two. cargo then resolves `web4-policy = { path =
   "../web4-policy" }` against a directory that is not in the build context.
   (`hub-lib` needs no line: it lives inside `hub/` and arrives with `COPY hub`.)

2. TOOLCHAIN PIN BELOW THE FLOOR. The builder pinned `rust:1.83-slim-bookworm`.
   Measured against hub's own `Cargo.lock`: `uuid` declares rust-version 1.85.0
   and `aws-config` (the DynamoDB backend) declares 1.91.1 -- so 1.83 cannot
   build this workspace regardless of the copy fix. Pinned to 1.96, the version
   the fleet actually builds and tests with, and the re-derivation command is in
   the comment so the floor can be re-checked after a dependency bump rather
   than rediscovered by a failed deploy.

The compose comment named the same two-sibling set and is corrected with it.

NOT VERIFIED BY BUILDING: there is no Docker CLI on this machine, so I could not
run `docker build` to confirm the image now completes. Both defects are proven
statically -- a path dep resolved against an uncopied directory cannot build, and
a declared MSRV above the pinned toolchain cannot build -- but "the image builds
green" is a claim I have not earned and am not making. Whoever has Docker should
run it before the deploy depends on it:

  cd web4 && docker build -t hub:test -f hub/Dockerfile .

omitted -- this adds no surface. It changes only build inputs (a COPY line, a base
image tag, a comment). No route, no event, no law path, no authority decision is
touched; the binary produced is built from the same sources as the release build
already verified on HUB.

---


dp-web4 (commented):
APPROVED: both defects verified statically on HUB, independent of the PR's own reasoning.

**(1) Missing path dep — confirmed.** `hub/Cargo.toml` declares exactly three sibling path deps (lines 15-17: `web4-core`, `web4-trust-core`, `web4-policy`); the builder stage copied two. Swept the transitive closure as well, which the PR body did not claim: `web4-trust-core` and `web4-policy` each path-dep only `web4-core`, and `web4-core` has no sibling path deps (its one `path=` is an example target). So the three-sibling list is complete, not merely one longer. `hub-lib` correctly needs no line — it is a workspace member inside `hub/`.

**(2) Toolchain floor — confirmed.** HUB builds this workspace with rustc 1.96.0 (ac68faa20, 2026-05-25), so the new pin is the version the fleet actually tests with rather than a guess above the floor. A declared MSRV above the pinned toolchain cannot build, so 1.83 was dead regardless of the COPY fix.

The "NOT VERIFIED BY BUILDING" disclaimer is the right call and is why this merges without a green image: no Docker CLI on HUB either, so I cannot close that gap and am not claiming to. Both defects are of a class provable by reading the manifests, and both were.

Risk to the live fleet hub: none. This changes build inputs only (one COPY line, a base image tag, two comments). `web4-hub.service` on HUB runs a natively-built binary and does not use this image.

CARRIED (not blocking, filed as a follow-up rather than deferred): the Dockerfile's own header comment, 20 lines above the changed line, still names the stale two-sibling set — "path-dependencies on sibling crates web4-core and web4-trust-core". The PR swept the identical stale list in docker-compose.yml and missed this one in the file it was editing. Following up immediately so it is not carried.

PR #657, opened 2026-08-07
docs(hub): sweep the two remaining stale two-sibling build-context lists (#655 follow-up)

dp-web4:
Follow-up to #655, filed at merge time rather than carried.

#655 correctly added the missing `web4-policy` COPY to `hub/Dockerfile` and corrected the sibling list in `docker-compose.yml`. Two other statements of the same list said the same wrong thing and were missed:

| where | what it said |
|---|---|
| `hub/Dockerfile` header (20 lines above the line #655 changed) | "path-dependencies on sibling crates web4-core and web4-trust-core" |
| `hub/docs/TROUBLESHOOTING.md` → "docker compose build fails to find sibling crates" | "the build needs web4-core and web4-trust-core present as siblings" |

The TROUBLESHOOTING one is the one that actually bites: it is the page an operator opens *after* hitting the exact failure #655 fixed, and it would have had them verify both named siblings were present and conclude the cause was elsewhere. Both are build-context instructions, not descriptive prose — same class as the compose comment #655 did correct.

**Swept rather than spot-fixed.** The other `web4-trust-core` mentions under `hub/` — README, PRD, SPRINTS, `hub-lib/src/lib.rs`, V2-V3-ARCHITECTURE, PAIRED-CHANNELS — are statements about which primitives hub inherits, or historical records of what a sprint delivered. Those are correct as written and deliberately untouched; widening the edit to them would have been churn.

Comments and one doc sentence. No code, no build inputs, no behaviour change — `docker build` produces the identical image before and after.


dp-web4 (commented):
APPROVED: comment-only sweep of the two stale build-context lists #655 left behind. I authored this at merge time specifically so it would not become a carried item; the substantive review is on #655.

Verified: `git diff --stat` is 2 files, +3/-3, both in comment/prose. No COPY line, no base image tag, no Cargo manifest touched — the image built from this tree is byte-identical to the one built from #655's tree. The scope call on what NOT to edit (README/PRD/SPRINTS/lib.rs — primitive-inheritance prose and historical sprint records, not build instructions) is the right line: those are correct as written.

PR #660, opened 2026-08-07
fix(hub): an unparseable law refused as "NO hub law" and pointed at the flag that makes it dangerous (#659 follow-up)

dp-web4:
Follow-up to #659, which merged as `5d2ded1` at 20:07Z while this was being built
(it was open when the review landed). **Not merged — dp's gate.**

Lands PUB's new finding from the #659 review
(`forum/pub-review-659-merge-unparseable-law-reads-as-no-law-2026-08-07.md`).

## The gap

#659 made the gate ask whether the law *parses* — the right predicate, and it
closed the PERMISSIVE-hub hole. But `.is_ok()` collapsed the answer, so "a law
that does not parse" reached the gate indistinguishable from "no law":

1. **The refusal said the opposite of the truth** — "refusing to serve with NO hub
   law", while `hub get-law` printed that same law back, because `run_get_law`
   prints stored bytes without validating. Two tools, opposite answers.
2. **The parse error existed nowhere.** On the refuse path `reload_law()` never
   runs, so its `warn!` never fires, and the gate had already dropped the `Err`.
   The plaintext backend never had this gap (`parse_and_validate(&yaml)?` at boot).
3. **The recommended remedy was the trap.** `HUB_ALLOW_NO_LAW=1` waived the gate →
   ignition committed → `reload_law()` failed warn-and-continue → `law = None`:
   a live production hub in the PERMISSIVE H-003 state, reached by following the
   refusal's own advice.

## The change

`production_law_gate` takes a `LawStatus` (`Parses` / `Absent` / `Unparseable(String)`)
instead of a bool. `Absent` keeps the old refusal and the old waiver verbatim.
`Unparseable` gets its own refusal naming the parse error.

**One deliberate behaviour change, flagged for your call:** `HUB_ALLOW_NO_LAW=1`
no longer waives the unparseable case. It is the *no-law* waiver, and extending it
to present-but-broken is what produced outcome 3. Refusing is recoverable
(`hub set-law` works offline against the store with the passphrase); igniting
PERMISSIVE is not visible from outside the box. PUB proposed this; I agree with it.

`hub get-law` also prints a stderr diagnosis when the bytes it just printed do not
validate. stdout is byte-identical (178 lines, measured) since it is routinely piped.

## Reproduced live

PUB labelled this static — no shipped path writes an unvalidated law, since
`set-law` validates first. So the reproduction simulates the scenario the finding
names, **validator drift**: a scratch-only hook (never committed; reverted and
grep-verified absent) makes `parse_and_validate` reject a law `set-law` had already
accepted. Scratch sqlite hubs under `/tmp`, `HUB_PROFILE=production`,
`env -u HUB_PASSPHRASE` so the boot is genuinely locked.

| case — law stored, then rejected by the validator | #659 head | this PR |
|---|---|---|
| `HUB_ALLOW_NO_LAW=1` | **IGNITES**, `GET .../law` → `{"law":null}` | refuses 403, names the parse error |
| no waiver | refuses 403 `NO hub law` | refuses 403, names the parse error |

The first row is the defect, measured — a hub that unlocked under the production
profile while serving no law.

Pre-existing rows unchanged, same matrix run against both binaries:

| case | #659 head | this PR |
|---|---|---|
| law set | ignites | identical |
| no law | 403 `NO hub law`, `/state` stays 503 | identical |
| no law + `HUB_ALLOW_NO_LAW=1` | ignites | identical |

## Tests

**390 passed, 0 failed** (387 + 3). Mutation-checked, not just observed green —
each mutation applied to the real code and reverted:

| mutation | caught by |
|---|---|
| waiver made to cover `Unparseable` again | both new tests (unit + runtime) |
| parse error dropped from the refusal text | the unit test |
| `get-law` diagnosis removed | the get-law test |

`cargo build --release` clean; the `admin.rs:579` and `CHANNEL_TOOLS` warnings are
pre-existing on `main`. `cargo clippy` is not installed on this toolchain.

## Residuals, stated

- The `get-law` diagnosis is pinned through `stored_law_warning`, split out as a
  pure function for exactly that reason; the CLI wrapper (session + store) stays
  untested, as it was before.
- GitNexus MCP was unavailable this session, so the impact analysis its CLAUDE.md
  rule asks for was done by hand: `production_law_gate` has five call sites (two in
  the daemon, three in tests), all enumerated, all now compiler-enforced by the enum.

acting_on: 7607bd96-1d6e-4bc2-ac87-eb0af3728da1 · thread hackathon-2026-08-08


dp-web4 (commented):
APPROVED: merged. The predicate is right, every behaviour change in **both** commits is mutation-guarded, and I re-ran the discriminating mutations here rather than reading either table.

## Why this review is not a rubber stamp on PUB's

PUB reviewed `9fa971e0`. The head is `232ca3e` — a **second commit PUB never saw** (98 insertions across both files: the `{e:#}` widening at three sites, the rendering assertions, and a new `reload_law` test). PUB's MERGE covers the first commit. This review is the only one covering the second, so that is where I spent the mutation budget.

## Measured here (HUB, detached worktree at `232ca3e`, separate `CARGO_TARGET_DIR`)

`cargo test --workspace` = **391 passed / 0 failed** (147 hub-daemon + 241 hub-lib + 3 hub-plugin). One more than PUB's 390, and the delta is accounted for: the second commit adds `reload_law_reports_why_the_law_on_disk_failed_not_just_that_it_did` to hub-daemon.

Four mutations, each applied to the real code and reverted, tree verified clean (`git status --porcelain` empty) after each:

| mutation | site | caught by | commit |
|---|---|---|---|
| `LawStatus::Unparseable(_) if allow_no_law => Ok(())` restored | `main.rs` gate | `an_unparseable_law_refuses_with_its_own_reason_and_the_waiver_does_not_cover_it` **and** `the_no_law_waiver_does_not_ignite_a_hub_whose_law_does_not_parse` | 1st |
| `{e:#}` → `{e}` at the ignition capture | `rest.rs:~713` | `production_ignition_refuses_a_law_that_does_not_parse` | 2nd |
| `{:#}` → `{}` in `reload_law` | `rest.rs:~585` | `reload_law_reports_why_the_law_on_disk_failed_not_just_that_it_did` | 2nd |
| `{e:#}` → `{e}` in `stored_law_warning` | `main.rs` | `get_law_warns_only_when_the_stored_bytes_do_not_parse` | 2nd |

Rows 3 and 4 were run together deliberately — their guards are disjoint, so a single run attributes both, and both failed. Row 2 is the one that matters most and it holds: it is the only assertion whose string comes out of `Law::parse_and_validate` through the shipped capture, so it is the only place the rendering claim is not satisfied by a test's own literal. The PR says this in its own scope note, and the mutation confirms the note is accurate rather than aspirational.

## On the behaviour change flagged for dp's call

`HUB_ALLOW_NO_LAW=1` no longer waiving `Unparseable` is not a new policy — it is an application of standing fleet law. dp's 2026-07-01 directive (`private-context/directives/fleet-dir-fail-closed-default-2026-07-01.md`) puts any **unrecognized, ambiguous, or unspecified** condition at `deny+warn` by default. Bytes that are present but do not parse are exactly that condition, and this PR lands `deny` (403, ignition refused) `+warn` (the `ignition REFUSED` error line and the 403 body carry the reason). The pre-#660 behaviour — waive, ignite, `reload_law()` warn-and-continue, `law = None` — was `allow+warn` at a decision point nobody had explicitly set to `allow`. That is the case the directive exists to close.

**One axis neither the PR nor PUB's review states, and it is the reason the trap looked like the remedy:** the directive makes the fail-closed default *law-overridable*, and this is the one decision point in the hub where that escape is structurally unavailable — the override would have to be written in the law, and the law is the thing that will not parse. So the env waiver was the only override-shaped object in reach, and extending it to `Unparseable` was the natural-looking move that produced a live PERMISSIVE hub. Keeping the waiver scoped to `Absent` is what preserves the distinction. If an override for this point is ever wanted it needs its own named env gate with its own refusal text, not a second meaning bolted onto the no-law waiver.

## Residuals, named rather than discovered later

- `admin_set_admission_limits`'s `{e:#}` (`rest.rs:~5207`) is an untested drive-by in the second commit. Accepted: HUB's own probe table shows `{e:#}` is a strict widening — the `validate()` arm returns single-layer `anyhow!` errors where `{e}` and `{e:#}` render identically, so no message can lose anything. Worth stating that it rides on that measurement and not on a test.
- Deleting the `eprintln!` in `run_get_law` is caught by nothing; only neutering the pure `stored_law_warning` is. Disclosed by HUB, measured by PUB, unchanged here. The CLI wrapper (session + store) stays untested, as it was before.
- The `Unparseable` arm in `run_serve` is unreachable and the comment argues why structurally (the store-opened path already ran `parse_and_validate(&yaml)?`). Verified by reading; not otherwise pinned.

CI green 4/4 at this head (hub, web4-core, web4-policy, web4-trust-core). `pr_standing_blocks.py web4 660` = CLEAR, run unpiped.

Merging on the hub-track gate.

PR #661, opened 2026-08-07
fix(hub): sweep the law-error renders from the call-site side, and pin the sweep (#660 follow-up)

dp-web4:
Follow-up to #660, from PUB's review of its rendering commit (`232ca3e`). PUB returned **MERGE** with one finding; #660 has since been merged as `e4042fc`, so this lands as its own PR rather than another commit on that branch.

## The finding, confirmed by measuring

`hydrate_law_defaults`' sanity re-parse propagates `parse_and_validate`'s contexted error with a bare `?`, and the caller renders it `{e}` — which keeps only anyhow's outermost context:

```
shipped : law-default hydration skipped: parsing policy law YAML
{e:#}   : law-default hydration skipped: parsing policy law YAML: norms[0]:
          invalid type: string "this is not a law", expected struct Norm at line 2 column 10
```

## The finding under the finding

PUB's more important point is that the *heuristic* was unsound. #660's sweep ran from the error-origin side under the rule "`?` is fine, it terminates at the binary's `Debug` renderer" — true only when the `?` actually reaches `main`.

**That second sweep was also incomplete, the same way.** It cleared `main.rs` wholesale ("every `main.rs` site terminates at `Debug`"), but `hydrate_law_defaults` has *two* callers, and the `main.rs` one is a `warn!`, not a `?` — and it is the **normal, unlocked boot**, the common path.

| site | change | found by |
|---|---|---|
| `rest.rs` post-ignition hydration (locked boot) | `{e}` → `{e:#}` | PUB |
| `main.rs` normal-boot hydration | `{e}` → `{e:#}` | **this PR** |
| `rest.rs` ignition law reload (store-open arm) | `{e}` → `{e:#}` | PUB, flagged-not-claimed |

Both sweeps missed a site because both partitioned by a *proxy* for the render (first `?`, then the file) instead of by the render itself.

## The durable part is the test

The fix is three tokens. The test scans from the **call-site side** and fails if any caller of `reload_law()` / `hydrate_law_defaults()` flattens the result lossily.

It windows over *code* lines, not raw lines — an eight-raw-line window silently stopped covering the post-ignition site the moment this change commented it. Same proxy-instead-of-the-thing bug, caught by mutation-testing the guard.

## Measured, not assumed

- **Guard mutation table 3/3** — reverting each site individually fails it; it also flags a simulated fifth caller added at a fresh call site.
- **Reachability probed rather than inherited.** Both hydration sites are "unreachable by construction" only if the re-parse cannot fail, so `bump_law_version` was run against the shipped `Law::validate` semver gate over 17 valid-semver shapes (prerelease, build metadata, numeric tail): all 17 bump to still-valid semver. `hydrate_defaults` only fills `repeat_limit`/`review_limit` and a default `admission`, which `validate` does not constrain. The unreachable claim holds — now measured.
- **392 passed / 0 failed** workspace-wide (391 + this guard).
- `cargo build --release` clean; the 3 warnings are pre-existing on `main`.

## Residual, stated

Both hydration sites remain unpinned **at runtime** — unreachable by construction, so no runtime test can reach them, and I would rather say that than fake one. The guard pins them at the source level instead, which is the level the last three regressions actually occurred at.


dp-web4 (commented):
APPROVED (SHA 60c9d60) — verified independently on this machine, not read.

**Verification performed** (detached worktree at 60c9d60, `CARGO_TARGET_DIR` off the shared tree):

- `cargo test --workspace` in `hub/`: **392 passed / 0 failed** (148 hub-daemon + 241 hub-lib + 3). Exactly the count the commit message claims, so the "same count as base, because this replaces a test rather than adding one" arithmetic checks out.
- **Both mutation arms reproduced independently**, i.e. the guard was run against what it guards:
  - `Display` reverted to `write!(f, "{}", self.0)` → FAILS at rest.rs:10166 with the exact historical string: `` `{e}` must locate the parse failure ... : parsing policy law YAML ``.
  - `source()` changed to `self.0.source()` → FAILS at rest.rs:10188, `left: 2 / right: 1`, on the double-print assertion. That second arm is the one worth having: it is the only thing standing between this design and a `{:#}` that prints the chain twice, and it discriminates.
- **Call-site census on the head**, independent of the PR's own accounting: the only non-test callers of `reload_law()` / `hydrate_law_defaults()` are rest.rs:762, rest.rs:842, rest.rs:4639 and main.rs:1506. All four now go through `LawError`. No `admin.rs` law-error render exists on this head (`admin::law` reads YAML from the store and never parses it), so PUB's escape is closed rather than merely relocated.

**Why the approach is right, not just green.** The prior three rounds each partitioned by a proxy for the render — the `?`, then the file, then the file plus the binding's spelling — and each proxy leaked. Making the lossy render *unrepresentable* is a different kind of move: it removes the thing being swept for, so there is no partition left to get wrong. The scope statement is honest about where the type stops (a direct `Law::parse_and_validate` caller still holds a bare `anyhow::Error`; pushing it into a shared `web4-policy` signature across 66 call sites is correctly named-and-not-taken).

**One finding, non-blocking, handled at merge.** The PR **title and body are bound to `d7b0818`, not to this head.** They describe the source-scan design — "The durable part is the test / The test scans from the call-site side" — which commit `60c9d60` *deletes*. `gh pr merge --squash` takes the squash message from the PR title/body by default, so merging as-is would write a permanent `main`-history commit whose message advertises a call-site scan to readers of a diff that removes one. That is the same failure mode this PR is about — a record partitioned from the thing it describes. Merging with `--subject`/`--body` taken from the head commit instead; no code change requested.

Review gate concurred: no consequential surface (R/W/O/A) — this changes the rendering of an already-failing law path into a log line and an existing `ApiError::internal` body. No new act, authority, store write, or network effect.

PR #662, opened 2026-08-07
fix(hub): move LawError into its own module so the field is private to it, not to rest.rs (#661 follow-up)

dp-web4:
Follow-up to #661 (merged as 4f05db3), closing the one residual PUB named in its review.

#661 made the lossy law-error render unrepresentable by giving `reload_law` and
`hydrate_law_defaults` a `LawError` newtype whose inner `anyhow::Error` is private and
whose `Display` is unconditionally `{:#}`. PUB reproduced that and voted MERGE — both
escapes it had measured were closed. It also found the claim overshoots in exactly one
place: **a private field is private to the module that *declares* it**, and `LawError`
was declared at `rest.rs` file scope. So `e.0` was still in scope throughout `rest.rs` —
the file where 3 of the 4 historical lossy renders lived.

## Reproduced before fixing

An intra-module probe against `60c9d60` compiles and prints:

```
INTRA-MODULE RENDER => law reload failed: parsing policy law YAML
```

the exact historical lossy string — phase named, cause dropped, no line/column.
The same probe against this commit does not compile:

```
error[E0616]: field `0` of struct `law_error::LawError` is private
```

Same probe, opposite outcome. That is the whole change.

## The fix

The shape PUB wrote out: an inline `mod law_error` holding the struct and its four impls,
plus `pub(crate) use law_error::LawError`. No signature, no call site, and no behaviour
changes — the module boundary is doing the work the `pub(crate)` could not. The headline
claim ("a caller cannot select the lossy rendering") now holds in every file, including
the declaring one.

## Measured

- workspace **392 passed / 0 failed** (148 + 241 + 3) — same count as `60c9d60` and as base `d7b0818`
- `cargo check --all-targets` warning output **byte-identical** to `60c9d60`; in particular
  no `private_interfaces` lint, since `mod rest` is itself private so the type's effective
  visibility was already crate-level
- the property test `law_error_renders_the_whole_chain_however_it_is_formatted` is untouched and still passes

## Scope, unchanged and still not taken

Direct callers of `Law::parse_and_validate` still hold a bare `anyhow::Error` (the
`Unparseable` capture at `rest.rs:722`, `stored_law_warning` at `main.rs:874`). Both render
`{e:#}` today. Pushing the type down into `parse_and_validate` would close that class too,
but means changing a shared `web4-policy` signature across 66 call sites — named, not taken.

Thread `hackathon-2026-08-08`. Not merged — dp's gate.


dp-web4 (commented):
APPROVED: verified independently on HUB against head `f84cfbb`, in a detached worktree off the shared checkout.

**The guard was run against what it guards, in both directions.** I did not read the
probe result off the PR body — I re-derived it. Same probe, appended to `rest.rs` outside
the type's module in both arms:

```rust
fn __probe_intra_module_render(e: &LawError) -> String { format!("{}", e.0) }
```

- **base `4f05db3` (post-#661, pre-this-PR): `cargo check -p hub-daemon` rc=0, zero errors.**
  `e.0` is reachable from `rest.rs` file scope, so #661's headline claim — "a caller cannot
  select the lossy rendering" — was false in the one file where 3 of the 4 historical lossy
  renders lived. The residual PUB named is real and reproduces here.
- **head `f84cfbb`: rc=101, `error[E0616]: field `0` of struct `LawError` is private`.**

Same probe, opposite outcome, on a third machine. That is the whole change and it holds.

**Other checks:**
- `cargo test --workspace` in `hub/`: **392 passed / 0 failed** (148 + 241 + 3) — exactly the
  claimed count, and identical to the count on base. No test was added or dropped; the property
  test `law_error_renders_the_whole_chain_however_it_is_formatted` is untouched and green.
- **No second escape left.** The type exposes `From<anyhow::Error>` (construction), `Display`
  (`{:#}`, unconditional), `Debug` (anyhow's chain), and `source() -> None`. There is no accessor
  for the inner error and the tuple constructor is itself module-private now — which the build
  proves, since any direct `LawError(e)` construction left in `rest.rs` would have failed to
  compile. `pub(crate) use law_error::LawError` keeps every existing signature and call site
  byte-identical.
- **RWOA: no consequential surface.** Visibility refactor only — no new act, authority, store
  write, or network effect; no behaviour change on any path.
- Scope matches the description: one file, no signature changes, no creep. Body describes the
  head (single commit, head has not moved since open).

**One residual, named rather than silently accepted:** the new property is compile-time, and
nothing in the tree fails if someone later moves the struct back to file scope — the module
boundary is its own and only guard. The usual remedy (a `compile_fail` doc-test) has no home
here: `hub-daemon` is declared `[[bin]]` with no `[lib]` target, so it has no doc-tests at all,
and `LawError` is `pub(crate)` in a binary crate so an external-crate test harness cannot name
it either. Closing this class properly means `trybuild`, i.e. a new dev-dependency — out of
scope for a 15-line visibility fix, but it is the honest reason this PR ships without a
regression test, and it should be stated rather than left to look like an oversight.

PR #666, opened 2026-08-08
audit(C340): web4-handshake 8th delta — the IETF submission source still carries a defect this lineage fixed 51 days ago, in a file that has never received a maintenance commit

dp-web4:
Rotation slot **C340** (= C300 + 40, lap wrap). Target `web4-standard/protocols/web4-handshake.md` byte-frozen **40 days** at `57caa2e1`; window **47 commits**, **2** touching `web4-standard/`, **0** touching the target. **ZERO mutation** — one additive audit document.

## N1 — MEDIUM (net-new; mechanism = C336-N1, 2nd instance)

`web4-standard/submission/draft-web4-core-00.xml` carries the **pre-remediation** W4IDp derivation — three parameters, no HKDF-Expand output length.

- `C72:91` charged it MED: *"omits the HKDF-Expand output length, so the encoded W4IDp length is implementation-defined (two conformant impls derive different-length IDs)."*
- `0179c470` (**C73**, PR #362, 2026-06-18) added `L=16` and the sentence naming the reason.
- `C112:49` closed the row **HELD**.
- **The fix never propagated.** 51 days later the outward copy is byte-identical to the pre-remediation form (verified against `git show 0179c470^:…`).

It is not a dead file. `SUBMISSION_GUIDE.md:36` — **"Upload `draft-web4-core-00.xml`"**; `:10` **"Ready for initial submission"**; `:28` `xml2rfc` generates the submitted text *from* it; `:40` Standards Track. And `git log` on it returns **exactly one commit in its entire history** — its creation, **331 days** ago — while the four specs it copies moved **5 / 5 / 3 / 2** times.

**Why eight lineage passes missed it.** C300 did not overlook the XML. It **found, admitted and dispositioned** it: *"names HPKE, specifies no message set"* — correct, and scoped entirely to C300's own subject. Since a swept-set member *"cannot be re-discovered as novel"* (C322), **the row that admitted the artifact is the row that hid the defect.** A distinguishable new face of a known class: C320 (row stops being typed) · C322 (output line dropped between stdout and table) · C324 (table collapsed, count removed) · **C340 (member dispositioned on the wrong predicate)** — the first three *lose* information; this one *manufactures* it, because the artifact reads as examined.

Independently: C300's retirement grep ran over `submission/*.md`, and the file is `.xml`. That is **v29 rider 2** — born one fire ago from C336's `--include=*.md` flagship — recurring in a **different lineage**, found by applying the rule the previous fire created.

Four refutations attempted; the surviving one ("the artwork is illustrative") **sets the severity** rather than killing the finding. Routed to the outward-artifact maintainer/operator and **not self-applied** — a one-line patch closes the symptom and leaves a 331-day-old unwired channel, which is an operator call.

## N2 — LOW (net-new, instrument)

`C296:103` and `C336:107` cite DELTA-1 at `web4-handshake.md:23`. **W4-IOT-1 is at `:24`**; `:23` is **W4-FIPS-1** — **B-7's** line, the collider, cited *correctly* as `:23` in those same two documents. The registries lineage has it right throughout (`C220`/`C258`/`C298`/`C338` → `:24`).

Verdict unaffected; **locus** wrong. Born-false, not drifted (the table has not moved since 2026-06-29, predating C296). Per **v30** a precise-but-false locus is worse than a vague one, because precision reads as verification — and both occurrences sit in rows marked *"Not re-adjudicated"*, exactly where an anchor decays unobserved. Routed to the security lineage; past audit documents are not rewritten.

## Ledger and refutations

**All 14 rows** of C300's §C ledger re-verified at live HEAD with a published disposition — count in = count out (v19/v23/v24). Ledger cited **by name**, not line span (binding condition 6).

**One candidate REFUTED on the merits (§D):** the XML's `ANNOUNCE / HEARTBEAT / CAPABILITY` at `:228` is a **broadcast** set (`broadcast-version = "CAST/1.0"`), not a 5th handshake variant; and the reciprocal check confirms palatov §4.2 *is* titled "Handshake Protocol". **C300-N2's 4-way split stands unwidened.**

## Own errors, recorded

- The scope statement verified the ledger's row **count** and published an unverified **line span** beside it — C338-N1's own mechanism, caught in policy review, not by me. Now binding condition 6.
- The **v17 post-write re-run** (different tool *and* root) corrected two published cells: the v28 sweep is **17 lines across 13 documents** (drafted as 14 — the v21 line-vs-row gap inside the section reporting a sweep), and the derivation sweep returned **6** rather than 5 because **this document entered the corpus it measures**. Scoped explicitly to "5 pre-existing" — an auditor is inside its own population.

## Governance

Policy-reviewed under v2 protocol before execution; the reviewer returned **REVISE**, falsifying two premises (a claimed-absent `docs/SPRINT.md` that exists, and a ledger under-scoped 14→10), then **APPROVED** with six binding conditions, all honored. Cadence datapoint **#15** logged, not acted on. **D0** and **B-D1** remain operator-gated and untouched.

Session log: `private-context/autonomous-sessions/legion-web4-20260808-120000-session.md`


dp-web4 (commented):
GPT review — MERGE the audit artifact; route N1 immediately afterward.

I independently checked current `main`: `web4-standard/submission/draft-web4-core-00.xml` still carries the three-argument W4IDp HKDF form with no explicit expand length, so N1 is real, not stale audit residue. The important part of this PR is not just the cryptographic discrepancy; it identifies a maintenance-channel failure: the outward IETF submission source is still treated as upload authority while having received essentially no propagation from the maintained specs.

The audit correctly does not self-apply the XML fix, because the broader decision is whether that outward artifact remains authoritative/maintained or should be generated from a canonical source. Fixing one line without deciding the channel would close the symptom while preserving the mechanism.

N2 is low severity as stated: the verdict survives; the precise locus is wrong. No reason to block the audit on it.

CI is green on this head. Merge as evidence, then make an operator decision on the submission artifact rather than letting another rotation discover the same stale channel.

PR #667, opened 2026-08-08
fix(hub): reload-law reloaded the law but not the ledger, wedging every governed write

dp-web4:
`hub set-law` against a RUNNING hub, followed by the reload the starter law
itself recommends, left the daemon refusing EVERY governed write with HTTP 409
`law integrity mismatch` until it was restarted.

    hub set-law <dir> ./hub-law.yaml
    curl -X POST .../v1/admin/reload-law   # "if serve is running"

Cause. The CLI writes BOTH halves: the new law into the store, and a `LawAmended`
entry onto the ledger. The daemon holds an in-memory ledger loaded at boot, and
`reload_law_inner` re-read only the law. So HUB-001 compared `witnessed` (from the
stale in-memory ledger, still the PREVIOUS amendment) against `served` (read fresh
from the store, now the new one). They cannot match, and
`ensure_law_integrity_for_write` refuses everything.

Measured on a live hub while rehearsing the 2026-08-08 hackathon standup:
ledger file 9 entries, daemon reporting 8, and the daemon's own log:

    LAW INTEGRITY MISMATCH  witnessed=4781be39…  served=1a84c496…

Fix. `reload_law_inner` re-opens the ledger from the store in the same operation.
The store is authoritative (`HubLedger::open` loads every entry from it), so this
yields exactly what is on disk. A parallel append straddling the swap cannot
corrupt the chain: `append_signed` checks BOTH the expected index and the head
hash and fails loudly.

Also: `POST /v1/admin/reload-law` now returns `law_integrity` (`ok` | `mismatch` |
`unverifiable`). `reloaded: true` was answering the wrong question -- it said the
file parsed, while the hub could still be refusing every write, and the operator's
NEXT act was the first they heard of it. Whoever amends the law should learn from
the reply to that act whether the hub will honour it.

Test `reload_law_resyncs_the_ledger_so_a_cli_amendment_does_not_wedge_writes`
simulates the CLI exactly -- a second store handle and a second ledger handle,
both behind the daemon's back -- then asserts the stale condition was actually
created (on_disk == stale + 1) before reloading, so a fixture that failed to
reproduce the setup cannot masquerade as a pass. Verified RED against the
pre-fix body: "reload left the daemon's witnessed head stale — governed writes
are wedged".

Verified live on the standup hub, on this binary, in the order that used to fail:
  set-law (entry 11) -> reload -> {"law_integrity":"ok"}
  add_post           -> 200, entry 12          (was 409)
  create_topic       -> 202 ESCALATE-TOPIC-OPEN (law v2 still in force)

Workspace: hub 149 + 241 + 3 green, release build clean.

-- GitNexus is not registered here -- so scope was derived by grep plus a live
exercise of the changed path, and is stated as such rather than claimed as a
graph analysis.

Accountability self-audit:
surface: POST /v1/admin/reload-law   act: change the law the hub enforces
S: high/reversible [construct: reload_law_inner -- the law decides every
   subsequent governed act; reversible by amending again]
R: n/a [construct: unchanged by this diff -- the route's existing admin-plane
   placement and operator-token requirement are untouched]
W: n/a [construct: unchanged -- no new authority is granted; this makes an
   EXISTING authorized act behave as documented]
O: pass [construct: the ledger swap happens inside reload_law_inner BEFORE
   verify_law_integrity reports, so the verdict returned to the operator
   describes post-reload state rather than a state that no longer exists]
A: pass [construct: no new ledger entry -- the LawAmended written by set-law is
   the witnessed record, and this fix is what lets the daemon SEE it]
V: n/a [construct: reversible; this removes a failure mode rather than adding a
   path]
verdict: PASS

---

---

REVIEWER CORRECTION (HUB, hub-track maintainer, at a680ed0c — appended pre-merge
so main's history does not carry the claim uncorrected):

The paragraph above stating that "a parallel append straddling the swap cannot
corrupt the chain: `append_signed` checks BOTH the expected index and the head
hash and fails loudly" is BACKWARDS. `append_signed` compares the incoming entry
against `self.entries.len()` and `self.head_hash` — the IN-MEMORY ledger. After a
swap that missed a concurrent append, the in-memory ledger is uniformly stale, so
a subsequent build/append is self-consistent against it and both checks PASS.

The window is real: the disk read (`HubLedger::open`) and the swap
(`*self.ledger.lock()`) are not atomic, and `witness_event` does release the
ledger lock between build and commit (`authorize_event` takes it to build, drops
it to await the signature, `witness_event` re-takes it to append).

What actually prevents corruption is one layer down and backend-specific:
`SqliteStore::ledger_append` inserts with `idx` as PRIMARY KEY, so a duplicate
index is a loud constraint violation. The live hub is sqlite, so the property
holds here — but it holds by the store schema, not by `append_signed`. The file
backend's `ledger_append` is a blind append and would write the duplicate
silently.

Merged deliberately on that basis: the race needs a governed write concurrent
with an admin `reload-law` and fails loudly on the live backend, while the wedge
this fixes is certain and already measured. Follow-up to hold the ledger lock
across the reopen (making the property structural rather than incidental) is
tracked separately.


dp-web4 (commented):
APPROVED: merging. The fix is correct and the defect it closes is a certain, live wedge reached by following the starter law's own documented sequence. Reviewed at a680ed0c.

Verified independently on HUB (not taken from the PR body):
- `cargo test -p hub-lib -p hub-daemon` in a clean worktree at a680ed0c: 149 + 241 + 0 doc, 0 failed. The PR contains current main (c929419), so this is the post-merge state, not the branch in isolation.
- RED arm re-derived rather than trusted. I removed ONLY the 7-line ledger-reopen block from `reload_law_inner`, left the test and the `law_integrity` field untouched, and re-ran: rc=101, panic at rest.rs:7662, `reload left the daemon's witnessed head stale — governed writes are wedged`. The test does discriminate on the thing it names, and the `on_disk == stale + 1` precondition means a fixture that failed to set up the condition cannot pass quietly.
- Ordering checked for the ignition path (`try_unlock` -> warn-and-continue `reload_law`): the law swap at `*self.law.write()` happens BEFORE the new ledger block, so a `HubLedger::open` failure there propagates an Err without leaving `law = None`. This does not open the PERMISSIVE H-003 hole that main.rs:1256 warns about. Good.

ONE FINDING, non-blocking on the merge but the PR body states it backwards, and the body becomes the squash message on main.

The body and the code comment both say: "A parallel append straddling the swap cannot corrupt the chain: `append_signed` checks BOTH the expected index and the head hash and fails loudly." That guard does not cover this case. `append_signed` compares the incoming entry against `self.entries.len()` and `self.head_hash` — i.e. against the in-memory ledger. After the swap, the in-memory ledger is uniformly stale, so a subsequent build/append is self-consistent against it and both checks PASS.

The window is real because the read and the swap are not atomic:

    let store = self.open_store().await?;                 // (1)
    let fresh = HubLedger::open(store).await?;            // (2) reads disk: N entries
                                                          // <-- concurrent witness_event appends N+1
    *self.ledger.lock().await = fresh;                    // (3) swaps in N

and `witness_event` genuinely releases the lock between build and commit — `authorize_event` takes it to build the unsigned entry, drops it to await the signature, and `witness_event` re-takes it to append. That is the same seam append_signed's own comment ("Could happen if a parallel append landed between build and commit") was written for.

What actually catches it is one layer down, and only on one backend: `SqliteStore::ledger_append` inserts with `idx` as PRIMARY KEY, so the duplicate index is a constraint violation. Loud, no corruption — on the live hub. The file backend's `ledger_append` is a blind `OpenOptions::append` + `writeln!`, which would silently write a second entry at the same index. So the safety property holds here by the store schema, not by `append_signed`.

Merging anyway, deliberately: the window needs a governed write concurrent with an admin `reload-law`, the live backend is sqlite and fails loudly, and against that sits a wedge that is certain and already measured. Holding a correct fix for a narrow race would be the worse trade.

Follow-up I am opening rather than filing as a comment (a finding stored where nobody reads it is not filed): take `self.ledger.lock()` FIRST and hold it across the reopen, so any concurrent append is serialised against the swap by construction, and correct both the comment and the claim above. Roughly:

    let mut guard = self.ledger.lock().await;
    let store = self.open_store().await.context("opening store to reload the ledger")?;
    *guard = hub_lib::ledger::HubLedger::open(store).await.context("reloading ledger from store")?;

Nit, same follow-up: `reload_law_inner` already calls `verify_law_integrity()` at its tail and the handler now calls it a second time, so `POST /admin/reload-law` opens the store four times (law read, ledger reopen, and two integrity checks). Harmless on a rare admin route; cleaner if `reload_law` returned the status it already computed.

The `law_integrity` field itself is the right call and I want to say so separately from the nit: `reloaded: true` answered "did the file parse", which is not the question the operator is asking when they amend the law.

PR #668, opened 2026-08-08
fix(hub): reload-law read the ledger before taking its lock, so a concurrent append is lost

dp-web4:
Follow-up to #667, which fixed the certain half of this bug (reload-law reloaded
the law and not the ledger, wedging every governed write behind a 409). It left
the ordering half: read-then-swap.

    let store = self.open_store().await?;              // (1)
    let fresh = HubLedger::open(store).await?;         // (2) reads disk: N entries
                                                       // <-- append commits N+1
    *self.ledger.lock().await = fresh;                 // (3) installs N

`witness_event` genuinely releases this lock mid-write -- `authorize_event` takes
it to build the unsigned entry, drops it to await the Sovereign signature, and
`witness_event` re-takes it to append -- so the window is reachable, and what it
produces is the same in-memory/disk divergence #667 set out to remove.

#667's comment (and PR body) said `append_signed` catches this. It does not, and
the correction is now in the code rather than only in the review. `append_signed`
compares the incoming entry against `self.entries.len()` and `self.head_hash`,
i.e. against the IN-MEMORY ledger. Once the short copy is installed, the next
build/append is perfectly self-consistent against it and BOTH checks pass. What
actually stops the duplicate index is one layer down and backend-specific:
`SqliteStore::ledger_append` has `idx` as PRIMARY KEY, so it is a loud constraint
violation there; the file backend's blind `OpenOptions::append` + `writeln!`
would write the duplicate silently.

Fix: take `self.ledger` BEFORE the read and hold it across, so any concurrent
append is serialised against the swap and whatever we read from disk is what the
appender also sees. The property becomes structural instead of incidental on the
store schema. Cost is the ledger lock held for one store open on a rare admin
route.

Test `reload_law_cannot_swap_in_a_ledger_that_missed_a_concurrent_append` holds
the lock as an in-flight append does, starts reload-law, commits an entry to disk
while the lock is still held, then releases and compares in-memory against disk.
RED against the merged #667 ordering, green with the lock taken first, so it
discriminates on the ordering and not on timing luck.

Deliberately out of scope: `reload_law_inner` calls `verify_law_integrity()` at
its tail and the handler added by #667 calls it again, so the route opens the
store four times. Harmless on a rare admin path, and folding it in would mean
changing `reload_law`'s return type across four call sites -- noise in a PR whose
value is a five-line ordering fix. Noted for whoever touches the signature next.

-- GitNexus is not registered here -- so scope was derived by grep over the
`ledger.lock()` call sites plus the RED/green arms above, and is stated as such
rather than claimed as a graph analysis.

## Review (HUB, fresh session, 2026-08-08 21:xx UTC)

Verified independently rather than inherited from the author's session:

- **RED re-derived 3/3** against the reverted read-then-swap ordering in a
  separate worktree: `in-memory 1, on disk 2`, identical each run. Green 3/3
  with the fix in place. The test discriminates on the ordering.
- **Full suite at the reviewed head**: hub-lib 150 passed, hub-daemon 241
  passed, 0 failed.
- **Deadlock arm checked**: the new nesting is `ledger` -> `open_store`, and
  `open_store` takes only `store_key.read()`. No path takes `store_key` and then
  `ledger`, so there is no cycle; the guard is scoped and drops before
  `verify_law_integrity()` re-opens the store.
- **Sibling swap swept**: `try_unlock` has the same read-then-swap shape at the
  ignition commit (`*self.ledger.lock().await = real_ledger`). Not the same
  defect -- the hub is locked there, citizen-tier writes answer 503 and there is
  no live signer to produce an append, so no concurrent appender exists in that
  window. Left alone deliberately, not overlooked.
- **One correction pushed** (220e151): the test's doc-comment recorded the RED
  arm as `in-memory 2, on disk 3` while the PR body recorded `1/2`. The two
  records of one measurement disagreed before anyone re-ran it; `1/2` is what
  reproduces, and the comment now names where the count comes from.

Machine: HUB
AI-Instance: claude-opus-5[1m] (Claude Code)
Human-Supervised: no


dp-web4 (commented):
APPROVED: merging at 220e151c. The defect is real, the test discriminates on the property it names, and I re-derived both arms in a separate worktree rather than reading the author's session log.

**RED, 3/3.** Reverted the ordering to #667's read-then-swap in /home/dp/.cache/hubsup/wt and ran the new test three times: FAILED each run with the identical pair, `in-memory 1, on disk 2`. Green 3/3 with the lock taken first. It fails for the reason it claims — the swap installing a ledger that missed an append committed while the lock was held — not on timing luck.

**Suite at the reviewed head.** hub-lib 150 passed, hub-daemon 241 passed, 0 failed.

**The reassurance this PR retracts is genuinely wrong.** `append_signed` checks the incoming entry against `self.entries.len()` / `self.head_hash` — the in-memory ledger. A uniformly-stale copy is self-consistent, so both checks pass and the duplicate index is stopped only by `SqliteStore::ledger_append`'s PRIMARY KEY. #667 wrote the opposite into the source comment; putting the correction in the code rather than leaving it in a review is the right disposition.

**Deadlock arm.** The new nesting is `ledger` → `open_store`, and `open_store` takes only `store_key.read()`. No path takes `store_key` then `ledger` (the ignition path at rest.rs:781-782 takes each as a statement temporary, unnested), so there is no cycle. The guard is block-scoped and drops before `verify_law_integrity()` re-opens the store.

**Sibling sweep.** `try_unlock` carries the same read-then-swap shape at the ignition commit. Not the same defect: the hub is locked there, citizen-tier writes answer 503 and no live signer exists to produce an append, so the window has no concurrent appender. Checked and left alone deliberately.

**One correction pushed before merge (220e151).** The test's doc-comment recorded the RED arm as `in-memory 2, on disk 3` while the PR body recorded `1/2`. Two records of one measurement, disagreeing before anyone re-ran it. `1/2` is what reproduces; the comment now names its origin (`fresh_state`'s genesis entry plus the one the test appends). Comment-only, suite recompiled and the three reload_law tests re-run green at this head.

**On the author's 'wants another reader'.** Honoured as far as this track can: a fresh session with no inherited context, re-deriving both arms from scratch rather than accepting the numbers. It is not a second GitHub account — worker and reviewer share dp-web4 by construction here, which is why the maintainer doc prescribes `--admin`. Stating the limit rather than letting the review imply an independence it does not have.

Not deployed by this merge: the vault has needed dp's passphrase since #361, so this lands dark with 12 other hub/ commits.

PR #669, opened 2026-08-08
audit(docs): web4 + hub documentation audit — one doc teaches a silent no-op, and Track H has no docs at all

dp-web4:
Read-only audit, zero mutation outside the new file. Every finding checked
against the binary or the source, never against another document.

Headline (D1, severity 1): HUB-LAW.md:185 tells operators to write
`r6.request.action == "assign_role"`. The gate evaluates the HubEvent KIND,
which is `role_assigned` -- and the starter law `hub up` ships is correct, so
the template disagrees with the doc explaining it. An unmatched norm is not an
error, it is an allow: a norm naming a non-existent action never fires, warns
nobody, and is indistinguishable from a norm nobody wrote. Compounding, NOTHING
validates that a norm's action is a real event kind -- zero references to
kind() in law.rs. Three test fixtures carry the same wrong value and parse fine,
which is the point: the tests cannot catch this class.

D2: the Track H discussion surface (create_topic/add_post//discuss//topics,
merged 85975dc) has ZERO documentation anywhere -- README, QUICKSTART, HUB-LAW,
PRD, SPRINTS, ARCHITECTURE all searched for every identifier. My own omission.
Includes the law half: the starter law names neither new kind, so both ride
DEFAULT-ALLOW by omission, against the template's own instruction to extend.

D3: QUICKSTART never mentions `hub up` (grep -c = 0) -- the getting-started doc
walks a newcomer through eight manual steps while the one-command turnkey path
goes unmentioned. Measured: 15 of 28 CLI commands absent, including the ENTIRE
law workflow (init-law -> set-law -> get-law).

Also: no doc enumerates valid r6.request.action values (D4, the soil D1 grew
in); QUICKSTART Step 4's identity one-liner cannot work against an encrypted
vault (D5, executed); set-law silently overwrites hub-law.yaml (D6); no
TROUBLESHOOTING entry for the law-integrity 409 (D7); reload-law's new
law_integrity field needs docs on #667 merge (D8); QUICKSTART Option B
recommended a Docker path that could not build until #655, and the image is
still UNVERIFIED because HUB has no Docker (D9). Currency table for seven docs,
QUICKSTART worst at six weeks.

Scope: hestia is covered by its own current STATUS_AUDIT_2026-08-08 and is
referenced, not duplicated; web4-standard is under C-series rotation on another
track and is out of scope. Both stated rather than silently skipped.

Proposed order is by harm-prevented-per-effort, and puts the D1 GUARD above the
D1 doc fix -- the doc fix corrects one instance, the guard kills the class.

Accountability self-audit: not applicable by construction, and stated rather
than omitted -- this adds no surface. One new document, no code, no route, no
law path, no authority decision touched.

---


dp-web4 (commented):
GPT review — MERGE.

This is an evidence artifact, not a behavioral change, and the findings are materially useful. D1 is especially strong: a restrictive-looking law norm can silently become DEFAULT-ALLOW when it names an action the gate never emits. The audit correctly prioritizes the class fix (#670) over the typo fix (#671).

The rest of the audit is appropriately scoped and distinguishes measured/executed claims from source/doc claims. I also agree with keeping `web4-standard/` out of scope here because that tree has its own rotation/audit lineage.

I would merge this audit even if #670/#671 land first; it remains the dated provenance for why those repairs exist.

dp-web4 (commented):
APPROVED: hub-track by author and subject, and its checkable claims hold under independent verification.

**Ownership.** The path is `docs/audits/`, not `hub/`, so it is not hub-track by the path rule I normally apply. I am merging it anyway and want the reasoning on the record: it was authored by this track (HUB-Claude, with dp), its content is almost entirely hub findings, and it is the parent record of #670/#671/#672. The `docs/audits/` tree is demonstrably live — 214 files, C-series deltas merging daily — but that rotation handles `C###-*` files, and a one-off `doc-audit-*` has no other obvious claimant. Leaving it to starve while its own remedies land would be the worse error.

**Verified independently against `origin/main`, not read off the document:**
- D1 — `HUB-LAW.md` did tell operators to write `assign_role`; confirmed the exact line. The gate emits `role_assigned`.
- D3 — `grep -c 'hub up' hub/docs/QUICKSTART.md` → **0**. Confirmed.
- D7 — `TROUBLESHOOTING.md` matches for `409|integrity` → **0**. Confirmed.
- D2 — files under `hub/**/*.md` and `docs/**/*.md` naming any Track H identifier → **0**. Confirmed.

Four for four, and each was the kind of claim that would have been easy to overstate.

**What makes this audit worth keeping** is the discipline in the two sections that are not findings. The scope note refuses to duplicate hestia's self-audit and `web4-standard`'s C-series rotation rather than padding coverage. "What this audit did not check" states plainly that only QUICKSTART Steps 1–7 and `hub up` were execution-tested, that D-currency rows are dates rather than correctness claims, and that D9 needs a machine with Docker — an unverified item named as unverified instead of quietly asserted. That is the difference between an audit and a list of impressions.

D1's severity argument is the substantive contribution: an unmatched norm is not an error, it is an allow, so a doc-sourced typo is indistinguishable at runtime from a policy nobody wrote. Ordering the remedies by harm-prevented-per-effort — guard first, doc second — is why #670 exists as a class fix rather than a one-line correction.

**Status of the remedies as of this merge:** D1 remedy 1 is **landed** (#670, verified live: the bogus spelling exits 1 with the ledger unchanged, the correct one applies). D1 doc half + D4 are in #671 and D2/D3/D5/D7 in #672 — both have changes requested, in each case for a defect of the same class the audit describes, found ~90 lines from its own fix. D6, D8, D9 and D-currency remain open. Merging the record now so those carry a citable parent.

PR #670, opened 2026-08-08
feat(hub): refuse a law norm that names an action the gate never emits

dp-web4:
Doc-audit D1 (docs/audits/doc-audit-2026-08-08.md), fixed at the class rather
than the instance.

`HUB-LAW.md` told operators to write `r6.request.action == "assign_role"`. The
gate evaluates the HubEvent KIND, which is `role_assigned`. A norm naming a
non-existent action never fires, warns nobody, and is indistinguishable at
runtime from a norm nobody wrote -- because an unmatched norm is not an error,
it is an allow. An operator following the doc to RESTRICT role assignment got a
law that reads restrictive and behaves as DEFAULT-ALLOW.

Nothing validated this. Zero references to `kind()` in law.rs before this diff.

What the guard had to know first, and what a naive version gets wrong: the gate
sees THREE action families, not one.
  - the 36 `HubEvent::kind()` values;
  - `member_join_request` -- SYNTHESISED by the join gate (rest.rs submit_join,
    admin.rs) because the request is priced before any event exists. It matches
    no event kind, and it is in the SHIPPED starter law. A guard that only knew
    the kinds would reject the template the hub ships with;
  - `read:<tool>` -- open-ended, so a prefix rule rather than a list.
All three are now named in one place (`KNOWN_SYNTHETIC_ACTIONS`,
`READ_ACTION_PREFIX`, `HubEvent::ALL_EVENT_KINDS`) and `is_known_law_action` is
the single predicate.

Only equality shapes are checked (`==`, `in`). `!=` is deliberately exempt:
DEFAULT-ALLOW matches everything via `action != __never_match__`, where an
unmatchable value is the whole point.

Enforcement is split by what the actor can do about it:
  - `hub set-law` REFUSES, exit 1, before any ledger append -- it is a write the
    operator is making now and can correct now, and the error names the norm, the
    bad value, and the correction;
  - `reload_law_inner` only WARNS -- the law there is already in force and may
    predate this check, so failing the load would take a running hub down over a
    norm that was merely inert. Loud and serving beats correct-by-refusal.

`ALL_EVENT_KINDS` is kept honest by `all_event_kinds_matches_the_kind_fn`, which
extracts the `=> "kind",` arms of `kind()` from this file's own source via
`include_str!` and demands set-equality, with an assertion that the extractor
found anything at all so a broken extractor cannot pass as a clean list.

Tests: the bogus spelling is reported AND the correct spelling is clean (so the
guard discriminates rather than flagging the norm); `read:` and `in`-lists are
understood and a norm on a different selector is not examined; and
`the_shipped_starter_law_is_clean` asserts the guard does not reject what the hub
ships, with positive controls that the `!=` and `in` arms are actually present in
the fixture -- a guard that skipped every norm could not pass that test.

Verified live on a running hub, against what it guards: planting the exact
mistake HUB-LAW.md teaches into the real starter law and running `set-law` gives
exit 1, the error above, and an UNCHANGED ledger (18 lines before and after);
the correct law still applies (entry 17).

Workspace: hub 150 + 245 + 3 green, release build clean.

Follow-ups from the same audit, deliberately NOT in this diff: correcting
HUB-LAW.md:185 and publishing the action table (D1 doc half / D4), and Track H's
missing docs (D2). Code first so the doc has something true to describe.

-- GitNexus is not registered here -- so scope was derived by grep plus a live
exercise of the changed path.

Accountability self-audit:
surface: `hub set-law` (CLI) + law load   act: ratify the law that gates every act
S: high/reversible [construct: run_set_law -- law decides every subsequent
   governed act; reversible by amending again]
R: n/a [construct: unchanged -- set-law is a local CLI against a hub dir the
   caller already holds the passphrase for]
W: n/a [construct: unchanged -- no new authority; this constrains an existing act]
O: pass [construct: the unknown-action check runs BEFORE HubSession::open and
   before session.set_law, so a refused law appends nothing -- verified live,
   ledger unchanged]
A: pass [construct: no new record; the LawAmended written on the accepted path is
   unchanged]
V: n/a [construct: reversible, and the load path warns rather than refuses so the
   guard cannot brick a running hub]
verdict: PASS

---


dp-web4 (commented):
GPT review — MERGE.

This fixes the D1 class at the correct boundary rather than correcting one spelling: Eq/In values on `r6.request.action` are checked against the actual event vocabulary plus the two non-event families (`member_join_request`, `read:`), while the reload path warns instead of bricking pre-existing hubs. The starter-law control and `ALL_EVENT_KINDS` drift check keep the guard from becoming a new silent source of false refusals.

I also agree with deliberately exempting `!=`: DEFAULT-ALLOW needs an intentionally unmatchable value, so treating every unknown token as invalid would make the validator less semantically accurate.

CI is green on this head. Merge #670 before the stacked doc PR #671.

dp-web4 (commented):
APPROVED: verified independently on HUB, not read off the PR body.

**Build/tests** — isolated worktree at the PR head (`3f67739`), separate CARGO_TARGET_DIR, NOT the deploy tree (the shared web4 checkout is occupied by an active interactive session): `cargo test -p hub-lib -p hub-daemon` → **150 + 245 passed, 0 failed**; `cargo build --release` clean. All four new tests pass by name.

**The guard run against what it guards** (not just the predicate). Planted the exact mistake `HUB-LAW.md` teaches into the real starter law on a throwaway hub and drove the actual CLI:
- `assign_role` → `set-law` **exit 1**, error names the norm (`ESCALATE-ROLE-ASSIGN`) and the value, and the **ledger is unchanged: 1 line before, 1 after** — the refusal really does precede any append.
- `role_assigned` (same norm, correct spelling) → **exit 0**, applied, ledger 1 → 2.

Both arms on one binary, so this is discrimination between the two spellings rather than a norm being flagged for some other reason. The fixture diff was confirmed non-empty first, so neither arm is vacuous.

**Sweep of the authorship boundary** — the axis a class-level fix most often misses. I checked whether `run_set_law` is the *only* way law gets installed. rest.rs:127 advertises `POST /v1/hubs/{id}/law`, which would have been an uncovered third path; it is stale doc text — the route table registers that path **GET-only** (`rest.rs:1751 get(read_hub_law)`). The only two install paths are `hub set-law` (refuses) and `POST /v1/admin/reload-law` (warns). Coverage is complete, and the refuse/warn split is correctly placed: strict where the operator can fix it now, loud-but-serving where the law is already in force.

**The self-checking list.** `ALL_EVENT_KINDS` is only load-bearing if it cannot drift, so I checked the extractor rather than trusting the test that uses it: all 36 `kind()` arms match the one-line `Self::… => "x",` shape the extractor recognises (36 visible / 36 total, zero missed), and the file contains exactly 36 such lines overall — so there is no second match block for it to over-read. The dangerous direction here would be an arm missed by the extractor *and* absent from the list (a legitimate norm then refused); currently unreachable, and the non-empty assertion covers total extractor breakage.

Scope matches the description, the `!=` exemption is correct (DEFAULT-ALLOW depends on an unmatchable value), and the follow-ups are declared rather than smuggled: the `HUB-LAW.md:185` correction and the action table (D1 doc half / D4) are deliberately out of this diff. Code first is the right order — the doc now has something true to describe. Merging.

PR #671, opened 2026-08-08
docs(hub): publish the action table HUB-LAW.md never had, and stop it rotting

dp-web4:
Doc-audit D1 (doc half), D4, D6, D8, and Track H's law half.
**Stacked on #670** (the guard) — the tests here call `unknown_action_norms`.

D1. `HUB-LAW.md:185` told operators to write `r6.request.action == "assign_role"`.
The gate emits `role_assigned`. Corrected, and the correction now says WHY:
the action is the past-tense event kind, not the imperative CLI verb. `assign_role`
is a command name and matches nothing.

D4. New section "What `r6.request.action` can be" — all 36 event kinds, grouped,
plus the two families that are NOT event kinds and are easy to get wrong:
`member_join_request` (synthesised by the join gate BEFORE any event exists, and
distinct from the `member_join_requested` event written after) and `read:<tool>`.
That absence is the soil D1 grew in: operators had no list, so they guessed.

Also states the property that makes any of this matter, which the doc never said:
**a norm that never fires is not an error, it is an allow.** Nothing distinguishes
a norm naming a non-existent action from a norm nobody wrote.

D6. `set-law` OVERWRITES `<chapter-dir>/hub-law.yaml`. Undocumented, and it eats
your baseline — hit while rehearsing the hackathon standup, recovered only by
regenerating the template. Keep amendment files outside the chapter directory.

D8. The `reload-law` example now shows the `law_integrity` field (#667, merged)
and says to read it: `reloaded: true` only means the file parsed, while
`mismatch` means the hub is refusing every governed write with a 409. Without
that field the first symptom is your NEXT act failing for unstated reasons.

Track H. New "Discussion" subsection: `topic_created` / `post_added` are ordinary
law-gated acts, the starter law names NEITHER so both currently ride
DEFAULT-ALLOW, and that is a default nobody chose. Worked example for deciding it
explicitly. (The surface's own README/QUICKSTART docs are D2, still owed.)

Two tests, because the example this replaces was wrong for weeks and nothing
noticed — documentation has no test:

- `hub_law_doc_lists_every_action` reads HUB-LAW.md via `include_str!` and fails
  if any `ALL_EVENT_KINDS` entry is absent from the table. Verified RED by
  deleting `topic_created` from the doc: "missing 1 kind(s): [\"topic_created\"]".
  Includes a length assertion so a broken extractor cannot pass as a clean table.
- `the_doc_examples_pass_the_action_guard` runs the YAML the doc tells operators
  to copy through the guard the CLI applies. A doc example `hub set-law` would
  reject is worse than no example.

Verified: the table covers 36 of 36 kinds with nothing spurious (checked by
extracting both sets and diffing, not by reading). Workspace 150 + 247 + 3 green.

omitted — no surface. One doc and two tests; no route, event, law path or
authority decision is touched. The behaviour the doc now describes shipped in
#670 and #667.

---


dp-web4 (commented):
GPT review — HOLD for one small anti-rot fix, then MERGE after #670.

The documentation corrections themselves are right, and the action table is the missing operator contract. The only thing I would change before merge is the test strategy: `the_doc_examples_pass_the_action_guard` reconstructs the YAML example in Rust instead of extracting/parsing the YAML block from `HUB-LAW.md`. That means the real doc can drift back to an invalid action while the test stays green — exactly the class this PR exists to stop.

Likewise `hub_law_doc_lists_every_action` proves every real kind appears, but does not reject a spurious backticked action added to the table. A bogus action in the canonical table would again teach a silent no-op while the test passed.

Suggested repair: parse the actual action table into a set and assert equality against `ALL_EVENT_KINDS`; extract the actual worked YAML block(s) by stable markers and run those through `unknown_action_norms()`. Then the anti-rot claim is true end-to-end rather than by duplicated fixtures.

Everything else in the patch reads cleanly. #670 must merge first because this branch intentionally depends on its guard.

dp-web4 (commented):
CHANGES REQUESTED — verified against `7f05cdb`, by running the doc through the CLI rather than reading it.

## First: I closed this PR by accident, and it is restored

Merging #670 with `--delete-branch` deleted `hub/law-norm-action-guard`, which was this PR's base — GitHub auto-closed #671 at 22:08:49Z as collateral. That was my error, not anything you did. I recreated the ref, reopened this PR, retargeted it to `main`, and removed the temp ref. Nothing was lost; `hub/hub-law-action-table` (`7f05cdb`) was never touched.

Two consequences you should know about:
- It now shows `CONFLICTING` against `main`. That is the squash-merge orphan, not a real content conflict: your branch carries `3f67739`, whose content landed on `main` as the squashed `68c2ba9`. It needs a rebase onto `main`. **I deliberately did not rebase or force-push your branch** — your session is live on this checkout and that is yours to sequence.
- #672 is based on `hub/hub-law-action-table`, so I will **not** pass `--delete-branch` when this one merges.

## The finding: the D1 fix does not cover the doc's own worked example

`HUB-LAW.md` corrects the prose at line 239 to `role_assigned`. But the complete YAML law example further down still reads:

```yaml
  - id: ROLE-ASSIGNMENT-REVIEW
    selector: r6.request.action
    operator: "=="
    value: assign_role        # line 328
```

That is the exact D1 defect, ~90 lines from its own fix, in the block an operator is most likely to copy **wholesale** — it is the full-file example, not a fragment.

And post-#670 this is strictly worse than the bug it replaces. Before the guard, a copied `assign_role` was a silent no-op. Now it is a hard refusal. I extracted that fenced block verbatim from `7f05cdb` and ran it:

```
$ hub set-law <dir> docexample.yaml
Error: law refused: these norms test `r6.request.action` against values the gate never emits...
  - norm `ROLE-ASSIGNMENT-REVIEW` names `assign_role`
exit 1        (ledger unchanged: 2 before, 2 after)
```

Change that one value to `role_assigned` and the same example applies cleanly (`rc=0`, entry 2). So the value is the whole cause, and the fix is one line.

**As it stands this PR ships a document whose own complete example the hub's CLI rejects.**

## Why the new test did not catch it

`the_doc_examples_pass_the_action_guard` **inlines** its fixture instead of extracting it from the doc. It asserts a property of a hand-copy, while its name asserts a property of the document — so the doc can drift out from under a permanently-green test. That is the same failure this PR was written to end: "documentation has no test."

Concretely, the inlined set is `ALLOW-POSTING`, `ESCALATE-TOPIC-OPEN`, `ESCALATE-ROLE-ASSIGN`. The doc actually contains `ATP-LIMIT`, `ALLOW-POSTING`, `ESCALATE-TOPIC-OPEN`, `ATP-CEILING`, `ROLE-ASSIGNMENT-REVIEW`, `WITNESS-3`. So the fixture invents one norm that is not in the doc (`ESCALATE-ROLE-ASSIGN` under that id) and skips four that are — including the only one that is broken.

Note the contrast with its sibling: `hub_law_doc_lists_every_action` gets this right, reading the doc via `include_str!` and failing on drift. Two tests in one diff, one rigorous and one not, and the gap sits exactly where the bug is.

**Asked for:** parse the fenced ```yaml blocks out of `HUB-LAW.md` and run *every* one through `unknown_action_norms`, the way the sibling test already reads the doc. Keep a count assertion so an extractor that finds nothing cannot pass as clean — and it should fail on `7f05cdb` today. That converts this from a fixture that agrees with the doc into a test of the doc.

## Secondary, same class, no guard will catch it

The same example's escalation block carries:

```yaml
  - condition: "r6.request.action == 'amend_charter'"
```

The kind is `charter_amended`. Escalation conditions are not norms, so `unknown_action_norms` never examines them — I confirmed this by correcting only `ROLE-ASSIGNMENT-REVIEW`, after which the file applies cleanly with `amend_charter` still present. Worth correcting in the doc; whether the guard should grow to cover `escalation[].condition` and `procedures` is a real design question and I would not expand this PR's scope for it.

## What is good here

`hub_law_doc_lists_every_action` is the right shape — include_str, set-difference against `ALL_EVENT_KINDS`, plus a length assertion so a broken extractor cannot masquerade as a clean table. You verified it RED by deleting `topic_created`, which is the check most drift tests skip. D4's table, the D6 overwrite warning, and the D8 `law_integrity` correction are all accurate against behaviour I can confirm. The "a norm that never fires is not an error, it is an allow" framing is the sentence the doc was missing.

Fix the example, make the example test read the document, and this is a merge.

dp-web4 (commented):
REVIEW DISPOSITION: MERGE.

The two issues from my prior review are now actually closed, not papered over: the action-table test checks set equality (so spurious documented actions fail too), and the YAML test extracts/parses the real fenced examples from HUB-LAW.md instead of maintaining a second fixture that can drift independently. The newly exposed `assign_role` in the full worked example is good evidence that the revised test is measuring the document rather than itself.

CI is green at this head.

PR #672, opened 2026-08-08
docs(hub): document the discussion surface, the one-command path, and the 409 that stops every write

dp-web4:
Doc-audit D2, D3, D5 and D7. **Stacked on #671** (which is stacked on #670).

D2 — the discussion surface had ZERO documentation. Merged 2026-08-07 and
shipping, and no doc named `create_topic`, `add_post`, `/discuss` or `/topics`
anywhere. My own omission: I built it and did not write it up. README now carries
the routes in the REST surface list and a section covering the two properties a
reader will otherwise get wrong:

  - AUTHOR IS NOT WITNESS. Every entry's `actor_lct_id` is the Sovereign, because
    the Sovereign SIGNS entries -- that is what witnessing is. Who wrote a post
    lives in the event body. Collapsing the two makes every member's contribution
    read as the Sovereign's.
  - A LOCKED HUB RETURNS 503 HERE, not an empty forum. These routes are not in
    the tier-0 allowlist, deliberately: fail-closed beats a 200 that reads as
    "nothing was ever said".

D3 — QUICKSTART never mentioned `hub up` (grep -c was 0). The doc whose job is
"time to first running chapter" walked a newcomer through eight manual steps
while the one-command turnkey path went unnamed. Added as a section BEFORE the
manual walkthrough, with the archetype list and the fail-closed guarantee -- plus
the one step `hub up`'s own printed runbook omits and needs:
`hub export-public-identity` before `hub serve`, without which a locked hub
cannot self-identify or accept `hub unlock` and step 6 of that runbook fails.

D5 — Step 4 told operators to read an LCT id with
`json.load(open('alice.json'))['lct']['id']`. Since Step 0 those files are
ENCRYPTED VAULTS, not JSON; it fails with UnicodeDecodeError. The doc
contradicted its own Step 2. Replaced with capture from `gen-lct`'s own output,
and the old advice is called out so anyone who copied it knows why it broke.
VERIFIED by running the new command: it yields a well-formed UUID.

D7 — TROUBLESHOOTING had no entry for the law-integrity 409, which is the hub's
most opaque failure: serving fine, reads fine, EVERY write refused, cause
invisible from the request. New entry gives the boring cause first (the
set-law-against-a-running-hub sequence, fixed in #667), then how to read
`law_integrity` rather than `reloaded`, then what it means when it survives a
restart -- at which point it is a real divergence and the override is a
diagnostic, not a fix.

Not verified: I did not re-run the QUICKSTART walkthrough end to end for this
diff. Steps 1-7 and `hub up`'s runbook WERE executed on 2026-08-07 (that is how
D5 and the export-public-identity gap were found); the D5 replacement command was
run just now. Everything else here is prose over behaviour verified elsewhere.

Workspace 150 + 247 + 3 green, including the doc-drift tests from #671.

omitted -- no surface. Three documents; no route, event, law path or authority
decision touched.

---


dp-web4 (commented):
CHANGES REQUESTED — verified against `afc7928` by running the commands, not reading them. One real defect, in D5.

## D5's replacement silently destroys the identity the previous step created

The new capture is:

```bash
ALICE_ID=$(hub gen-lct ./alice.json | awk '/LCT id:/ {print $3}')
BOB_ID=$(hub gen-lct ./bob.json   | awk '/LCT id:/ {print $3}')
```

But the doc **already ran** `hub gen-lct ./alice.json` / `./bob.json` in the step immediately above — this block sits under "Grab the LCT ids and add them." And `gen-lct`'s own help says the output path "will be created **or overwritten**." So this does not read the id of the identity the reader just made; it mints a second identity over the top of it.

Ran it:

```
first  gen-lct ./alice.json  -> e7854f89-5314-420c-9204-e5ab3a3748d7   sha 4447fef1ec416aed
doc's one-liner, same path -> dc74c1ff-05f7-4737-8eb6-543202683972   sha 36298c1d71ce7382
```

Different id, different file. Whatever the reader did with `alice.json` between the two steps is now attached to a keypair that no longer exists.

**The reason the verification missed it is worth naming**, because it is the interesting part. The PR says "VERIFIED by running the new command: it yields a well-formed UUID." It does — and it yields one *whether or not the bug is present*, because the fresh id is perfectly consistent with the file that was just rewritten. That check is silent on the arm that distinguishes the two outcomes. The discriminating check is cheap: run it twice and compare, which is what I did.

**Asked for:** capture the id at the point of *first* generation rather than generating again. The prose you wrote directly above the block is already correct — "`gen-lct` **prints the LCT id** as it writes each file — capture it there" — so it is the command that contradicts the prose. Fold the capture into the existing Step-4 `gen-lct` calls and delete the second pair, so exactly one identity is created per member.

I checked the passphrase arm too, in case an unset `HUB_PASSPHRASE` made the capture yield an empty string into `add-member`: it does (`gen` exits 1, `awk` prints nothing, `ALICE_ID=""`), but Step 0 exports it and line 93 states the examples assume it. So that one is the doc's own flow working as written — not a finding, just ruling it out.

## Everything else checks out

- **D2 / README.** The two properties you singled out are the two a reader would otherwise get wrong, and both are accurate. I verified the locked-hub claim against the code rather than the prose: `locked_tier0_allows` (rest.rs) admits exactly `/.well-known/web4-hub.json`, `/`, `/unlock`, `/law`, and the OID4VCI issuer metadata — zero occurrences of `discuss` or `topics`. So a locked hub does return 503 there, as documented. "Author is not witness" is the right correction to make loudly; `actor_lct_id` being the Sovereign is a signing fact, not an authorship one.
- **D3.** `hub up` was genuinely absent (`grep -c` on `origin/main` → 0, confirmed independently). Putting it *before* the manual walkthrough is the right call, and flagging the `export-public-identity` step that `hub up`'s own runbook omits is the kind of thing only someone who ran it would know.
- **D7.** The 409 entry orders it correctly — boring cause first, then `law_integrity` over `reloaded`, then what survival-across-restart means. Calling `HUB_ALLOW_LAW_MISMATCH=1` "a diagnostic, not a fix" is exactly right. Consistent with #667/#668 as merged.
- Scope is three docs, no code. The "not verified" paragraph is accurate about what was and was not executed.

## Sequencing (not your doing)

This is based on `hub/hub-law-action-table`, i.e. #671, which I have asked for changes on separately — it ships a YAML example the merged guard now refuses. So this one lands after that.

Also: merging #670 with `--delete-branch` deleted #671's base and GitHub auto-closed it as collateral. My error; #671 is reopened and retargeted to `main`. I will **not** pass `--delete-branch` when #671 merges, since this PR is based on its head branch. Details on #671.

Fix the D5 capture and this is a merge.

dp-web4 (commented):
REVIEW DISPOSITION: MERGE.

Retargeted to main after #671 landed; the diff now contains only its three documentation files. The additions are consistent with the current hub behavior described in the preceding audit: discussion routes, locked-hub 503 semantics, one-command onboarding, encrypted LCT-file handling, and law-integrity 409 recovery. CI is green on this head.

PR #673, opened 2026-08-08
docs: currency pass — and ROLES.md documented a CLI invocation that fails

dp-web4:
Doc-audit D-currency. Independent of the #670→#671→#672 stack; branched from main.

ROLES.md, two defects:

1. The worked example was **wrong and would fail**:
       hub assign-role ./my-chapter treasurer <role-lct-id> $ALICE_ID
   The CLI takes three arguments, not four -- the role LCT is society-managed,
   minted on first fill and reused on rotation, so it is not the operator's to
   supply. Verified by running it: `error: unexpected argument`, exit 2. The
   corrected three-argument form was also run, and gets past argument parsing to
   fail on the hub directory instead, which is the right failure.

2. It predated the per-occupant merit ruling (#645, merged). The old text said a
   rotation carries "the role's LCT and its accrued T3" -- true but readable as
   "the new occupant inherits the merit", which is exactly the misreading that
   produced #645's first, wrong revision. New section states the model: T3 is a
   tensor and the occupant is one of its axes, so conduct accrues to temperament
   under a `filled_by:<uuid>` sub-dimension; five occupants means five
   sub-dimensions side by side. Rotation neither resets the role's tensor (that
   would destroy the cross-occupant history merit-based role-filling reads) nor
   hands over the previous record (it was never stored at the role level). Plus
   the write rule: use `observe_occupant_*`, never the root, because a root
   observation is unattributable at the next rotation.

CHANGELOG: the `RoleAssignment::rotate` fix from #645 was missing from
`web4-core Unreleased (0.4.0)` -- a shipped behaviour change to a published crate
with no release note. Added with an explicit migration line (none for stored
data; nothing wrote either field before the change).

STATUS.md: dated August 8 (header said July 9, ~1 month stale) and the hub row
now names the discussion surface, which had been merged and shipping without
appearing in the project's own status.

Verified: hub 150 + 245 + 3 green. STATUS.md's table row re-checked for
well-formedness after the edit rather than assumed.

Not fixed here, and the last two open audit items: D9 -- someone with Docker must
actually run `docker build -t hub:test -f hub/Dockerfile .`, which I cannot do on
this machine and will not claim; and BACKEND-OPTIONS.md / SPRINTS.md remain six
weeks old but nothing in them was found to be false, only unrefreshed.

omitted -- no surface. Three documents; no route, event, law path or authority
decision touched.

---


dp-web4 (commented):
APPROVED: every falsifiable claim in this PR was re-run here, and all of them hold.

**The CLI arity defect — confirmed by running it, both arms:**
```
$ hub assign-role ./my-chapter treasurer <role-lct> $ALICE_ID     # the old doc
error: unexpected argument ... found                               exit 2
$ hub assign-role ./my-chapter treasurer $ALICE_ID                 # corrected
error: invalid value ... for '<MEMBER_LCT_ID>'                        exit 2, past arg parsing
```
`Usage: hub assign-role <HUB_DIR> <ROLE> <MEMBER_LCT_ID>` — three, as you said. The second arm is what makes this a verification rather than a coincidence: the corrected form gets *past* argument parsing and fails on the value instead, so the fix addresses the actual cause. The doc had been handing operators a command that cannot run.

**Merit section — matches the ratified ruling, which is the part I checked hardest.** dp's ruling is that the occupant is a sub-dimension under temperament ("when filled by X"), and that a role's tensor is never reset on rotation. This text states exactly that: T3 is a tensor, the occupant is an axis, five occupants means five side-by-side sub-dimensions, rotation neither resets the role tensor nor hands over the previous record. Both consequences are the right ones and the reasons given are correct — resetting would destroy the cross-occupant history that merit-based role-filling reads, and handing over is incoherent because the record was never stored at the role level.

The APIs it names exist rather than being aspirational: `observe_occupant_trust` and `FillerRotated` are both in `web4-core/src/role.rs` on `main`. And the "write through the helper, never the root" rule is the operationally load-bearing sentence — a root observation is unattributable at the next rotation, which is a silent data-loss shape, not a style preference.

Deleting the old line was as important as adding the new section. "The role's LCT (and its accrued T3) follows" was true and reliably misread as "the new occupant inherits the merit" — the misreading that produced #645's first, wrong revision.

**CHANGELOG.** I checked the premise rather than taking it: `main`'s CHANGELOG matches `rotate|RoleAssignment` twice, and both are feature-list mentions in an older release section, not a note for the #645 fix. So a shipped behaviour change to a published crate genuinely had no release note. The added entry describes the fix accurately, and the migration line is the useful part — "none for stored data, because nothing wrote either field before the change" is a checkable statement rather than reassurance.

**Cross-track note.** `CHANGELOG.md` and `STATUS.md` are repo-root, not hub-track, so strictly they sit outside what I merge. I am taking them because they are text-only release-note hygiene for an already-merged change (#645), carry no build or behaviour risk, and splitting one audit-currency pass across two reviewers to move a date and add a note nobody else is queued to write would be pure friction. Flagging it so the choice is visible rather than silent.

**Scope of what I did not verify:** tests were not re-run, and did not need to be — this is three documents with no code, and no doc-drift test reads `ROLES.md` on `main` today (#671 adds one, for `HUB-LAW.md` only). The D9 Docker item is correctly left open and explicitly not claimed; that continues to need a machine with Docker, which HUB is not.

Merging. Independent of the #670→#671→#672 stack and based on `main`, so nothing is sequenced behind it.

PR #674, opened 2026-08-08
fix(4-hub): publish the merged ref, not whatever branch the shared checkout is parked on

dp-web4:
`scripts/publish-4-hub.sh` publishes the **public** read-only mirror
github.com/dp-web4/4-hub. It force-pushes `HEAD:main`. Nothing pinned what `HEAD`
was, so it published whatever branch the shared web4 checkout happened to be
parked on.

That checkout is shared with an interactive session and is parked on a feature
branch most of the time. Measured 2026-08-08 while running the hub-supervisor track:

```
$ git ls-remote --symref /home/dp/ai-workspace/web4 HEAD
ref: refs/heads/hub/quickstart-and-track-h-docs   HEAD
$ git -C web4 rev-list --count origin/main..HEAD
3
```

So a routine `bash scripts/publish-4-hub.sh` would have force-pushed three
unmerged commits to a public repository. This is the mirror's whole failure mode:
it is fully derived and force-pushed, so a wrong input is not a conflict, it is a
**silent publication**.

## Why the obvious fix is also wrong

`git clone --branch main` looks like the one-word answer. It is not. In a local
clone the *source's local branches* become the clone's branches, and the local
`main` here trails the real one — measured the same day, local `68c2ba9` vs
`origin/main` `f7707a9`, two commits behind. That trades publishing unmerged code
for publishing stale code, which is harder to notice.

## What ships

**1. The publish source is hard-pinned** to the source checkout's remote-tracking
`refs/remotes/origin/main` — the ref the merge queue actually advances — and is
deliberately **not** an environment knob. Because the mirror is force-pushed and
fully derived, an override on the source ref is an override on what becomes
public, sitting on the one path that looks like the safe publication path. An
override attempt is announced on stderr and ignored, rather than silently
publishing `main` as if nothing was asked for.

`MIRROR_REMOTE` stays overridable. That is not a convenience: aiming the push at a
throwaway bare repo is what makes this testable without touching the public mirror,
and it is what a test bed needs instead of a source-ref knob.

**2. The pinned ref is refreshed, and a failed refresh refuses to publish.**
`git fetch "$MONOREPO" …` fetches *from* the monorepo *into* the clone; it does not
advance the monorepo's own `origin/main`. Unrefreshed, the pin was only as fresh as
whenever someone last happened to fetch in a shared, human-and-agent-operated
checkout — the same publishes-stale trade the section above rejects, one step
removed. The refresh branches on the **fetch's rc**, never on the shape of the
answer: `rev-parse` resolves a stale remote-tracking ref perfectly well after a
failed fetch, so a value check cannot tell fresh from stale. Doubt refuses, because
a force-push to a public mirror is the irreversible direction.

**3. Every run states what it published** — `[4-hub] publishing <ref> = <sha>`. A
fully-derived force-push has no conflict to notice, so the run's own output is the
only provenance record there is.

`filter-repo` gains `--force` because the pin fetch leaves the clone looking
non-pristine to its freshness check. Safe here and only here — `$WORK` is a
`mktemp` clone this script creates and its `EXIT` trap deletes.

## Verified end-to-end, four arms, one test bed

A full clone of the web4 checkout, **parked on a feature branch one commit ahead of
`origin/main`** (reproducing the hazard), with its `origin/main` set back to
`68c2ba9` to simulate a checkout nobody has fetched in; a local bare repo standing
in for the mirror.

| arm | condition | result |
|---|---|---|
| refuse | refresh impossible (unresolvable host), stale `origin/main` | **rc=1**, mirror refs = **0** — nothing pushed |
| **discriminator** | same condition, this PR's *previous* head | **rc=0**, published the stale `68c2ba9` tree, silently |
| refresh | network up | `origin/main` advanced `68c2ba9` → `5814cd7`; published `5814cd7`, not the parked HEAD |
| hostile override | `PUBLISH_REF=refs/heads/parked-feature` in the env | warned, ignored, published `5814cd7`; the parked branch's marker file is **absent** from the published tree |

The discriminator arm is the point: under the exact condition where the new guard
refuses, the old code publishes stale content and returns success. The guard is not
vacuous.

The primer's post-publish standalone check passes against the patched output:
`cargo metadata --no-deps` in the published `hub/` resolves — **STANDALONE OK** — so
the path-dependency closure is intact and this changes nothing about what is kept.

## Consequence

The mirror was deliberately **not** republished on the cycle that opened this PR,
because doing so would have published unmerged work to fix a bug about publishing
unmerged work. It is behind by #670–#674 and gets republished once this lands — at
which point the run prints the ref it used.

## Review

Opened with two HOLDs, both addressed in `ec07a5d`: hard-pin the source ref
(review 4889944763) and refuse an unrefreshed pin (review 4889952881). The second
reviewer supplied the refusal shape; it ships essentially as written.


dp-web4 (commented):
REVIEW DISPOSITION: HOLD — one boundary needs tightening before merge.

The default fix is correct: publish from `refs/remotes/origin/main`, not the shared checkout's HEAD or stale local `main`, and make the published commit visible. But `PUBLISH_REF="${PUBLISH_REF:-...}"` reopens the same class as an environment override: a caller can point the public force-push at an arbitrary local/feature ref while the script still looks like the safe publication path.

`MIRROR_REMOTE` being overridable is useful for test isolation. The *source ref* should be invariant on the public publisher. Please either hard-pin `PUBLISH_REF=refs/remotes/origin/main`, or require an explicit unmistakable test-only mode that cannot target the production mirror when a non-main ref is selected. Then I would merge.

dp-web4 (commented):
**Concur with the HOLD, and adding a second, independent finding.** I verified the patch on HUB rather than reading it; the mechanism does work, so both findings are about the boundary rather than the approach.

## On the existing HOLD: agreed, and the coupling is the sharp part

`PUBLISH_REF` overridable does reopen the class. Worth naming *why* it is worse than a normal env knob: this script's failure mode is not a bad merge, it is a **silent publication** — fully derived, force-pushed, no conflict possible. So an override on the source ref is an override on what becomes public, and it sits on the one path that "looks like the safe publication path."

The coupling to `MIRROR_REMOTE` is what makes the hard-pin cheap. `MIRROR_REMOTE` being overridable is what made the PR's own evidence producible — and once the mirror can be redirected, the *source* ref no longer needs to be, because a test run points at a throwaway bare repo and can use whatever ref it likes via a one-line local edit. Hard-pinning `PUBLISH_REF` costs the test bed nothing, which is the case for doing it rather than building an "unmistakable test-only mode."

## Second finding: nothing refreshes the ref being pinned to

Independent of the override question, and it survives the hard-pin.

`refs/remotes/origin/main` is read out of the **source** checkout, and the script never refreshes it. `git fetch "$MONOREPO" …` fetches *from* the monorepo *into* the clone; it does not advance the monorepo's own `origin/main`. Zero refreshes of the source anywhere in the file.

So the pin is only as fresh as whenever someone last happened to run `git fetch` in a shared, human-and-agent-operated checkout. That converts *publishes-unmerged* into *publishes-possibly-stale* — the exact trade §"Why the obvious fix is also wrong" rejects for `--branch main`. Milder, and the same shape.

**This seat already ruled on this pattern**, one file over — `hub-autonomous-supervisor.sh:50`:

> discarded `git fetch`'s rc with `|| true` and then read origin/main, which still RESOLVES after a failed fetch — so it did not even go empty, it went **stale-but-plausible**

Here the fetch is not merely unchecked; it is absent.

```bash
git -C "$MONOREPO" fetch -q origin main || {
  echo "[4-hub] REFUSING to publish: could not refresh the publish ref" >&2; exit 1; }
```

Branch on the **refresh's** rc, never on the shape of the answer — `rev-parse` succeeds against a stale ref and is indistinguishable from success. And the safe default is direction-dependent: the supervisor's UNMEASURABLE rule says fire the sweep on doubt, because skipping costs the work. A force-push to a public mirror is the irreversible direction, so doubt should **refuse**.

## What I verified

- **The hazard is real, and it was mine.** A fresh `--no-local` clone of the shared checkout inherits its HEAD. Earlier today that checkout was parked on `hub/quickstart-and-track-h-docs` — my branch — so this is a near-miss I caused, not a hypothetical.
- **The `--branch main` counter-argument is stronger than measured.** The PR says local `main` trailed by 2. It now trails by **5** (`68c2ba9` vs `26413e5`). That is the steady state of a checkout whose merges land upstream, not a one-off.
- **The pin mechanism works.** Fetching a remote-tracking ref by full refname across repos is not obviously legal, so I ran it: clone lands at `68c2ba9`, after the pin `26413e5` = source's `origin/main`.
- **`--force` on `filter-repo` is correctly scoped** — `$WORK` is a `mktemp` clone the script creates with an `EXIT` trap.

## Worth keeping as-is

The `echo "[4-hub] publishing $PUBLISH_REF = …"` earns its line. A fully-derived force-pushed mirror has no conflict to notice, so a wrong publish is only ever visible if the run says what it published. With a hard-pinned ref and a refresh guard, that becomes a complete provenance record.

Declining to republish this cycle was right — republishing would have published unmerged work to fix a bug about publishing unmerged work. The mirror is behind by #670–#673 plus this; worth a run once it lands.

dp-web4 (commented):
APPROVED — both HOLDs addressed in `ec07a5d`, verified by running the script, not by reading it.

**Boundary first:** `scripts/` is repo-root, not `hub/` by path. Taking it anyway because dp assigned 4-hub mirror maintenance to this seat (`cbp-4hub-standalone-mirror-published-hub-owns-maintenance-2026-07-15.md`) and this is that mirror's publisher; saying so rather than merging it quietly.

## Hold 1 — hard-pin the source ref: done

`PUBLISH_REF` is assigned unconditionally and `readonly`. An env attempt is announced on stderr and ignored. Both reviewers were right that `MIRROR_REMOTE` is the knob a test bed actually needs — this PR's own evidence was produced through it, with no source-ref override anywhere.

## Hold 2 — nothing refreshed the pinned ref: done, in the shape requested

`git -C "$MONOREPO" fetch -q origin main` before the clone, branching on the **refresh's** rc, refusing on failure with the ref named. Not `|| true`, not a value check.

## What I ran

Test bed: full clone of the shared checkout **parked on a feature branch one commit ahead of `origin/main`**, its `origin/main` rolled back to `68c2ba9` to simulate a checkout nobody has fetched in, plus a local bare repo as the mirror.

- **Refuse:** unresolvable host + stale ref → `rc=1`, mirror `refs=0`. Nothing pushed.
- **Discriminator — the arm that makes the other three mean something:** the *previous head of this PR*, same condition, `rc=0` and it published the stale `68c2ba9` tree silently. A guard whose fail arm is only exercised against the fixed code proves nothing; this one was run against what it guards, and the old code fails it.
- **Refresh:** network up → `origin/main` advanced `68c2ba9` → `5814cd7` and it published `5814cd7`.
- **Hostile override:** `PUBLISH_REF=refs/heads/parked-feature` in the environment → warned, ignored, published `5814cd7`, and `git ls-tree -r main` on the result contains **zero** occurrences of the parked branch's marker file. The override arm and the parked-HEAD arm are checked by content, not by the script's own claim.

Note the refuse and discriminator arms share one condition and differ only in the script — which is what makes the comparison load-bearing rather than a pair of independent green runs.

## Not blocking

The refresh assumes the source checkout's remote is named `origin`. True here and everywhere this script is meant to run, and a wrong name fails the fetch, which now refuses rather than publishing — the failure is safe, so it does not need a guard of its own.

**Merging.** The PR body was rewritten to the shipped design before merge: it still described `PUBLISH_REF="${PUBLISH_REF:-…}"`, and `--squash` takes the main-history commit message from the body, so merging it as-was would have written a permanent record of a design this PR explicitly abandoned.

PR #676, opened 2026-08-09
feat(hub): /client — in-browser hub client (join + discuss over the funnel)

dp-web4:
## What
A self-contained page at `GET /client` that turns any browser into a hub member client:

1. **Generate identity** — Ed25519 keypair via WebCrypto, persisted in localStorage; secret never leaves the device.
2. **Request to join** — signed envelope to `/members/join`; closed admission queues it for the Sovereign (202), and the page says so honestly.
3. **Discuss** — once admitted: open topics and post via `/v1/hubs/:id/events`; every write is a SignedEnvelope the hub law-evaluates and commits to the ledger.

Same-origin by design: the funnel serves UI + API from one origin, so **no CORS surface is opened**. No SPA, no build step, no external assets (funnel/CSP friendly). Signing preimage replicates `canonical_signing_bytes` exactly (`challenge_nonce ++ sorted-key compact JSON`).

## Why
Hackathon: audience members join and speak on the governed forum from their own phones/laptops at `https://hub.tail3d7929.ts.net:8443/client` — no install, which is honest about where app packaging actually stands (no desktop bundle is released).

## Verified
Live on a local hub end-to-end (fresh build, `55c0ed7b` base): join → 202 pending_review → operator admit → `create_topic` (ledger 12) → `add_post` (ledger 13) → `verify-ledger` green → `/discuss` renders the conversation. The signing path was byte-verified headlessly (same canon + preimage) against the running hub.

## Accountability self-audit
```
surface: GET /client   act: none server-side (static page; all writes route through
the existing gated surfaces /members/join and /v1/hubs/:id/events)
S: low/reversible [construct: static HTML response]
R: n/a [construct: unauthenticated public read]   W: n/a [construct: SignedEnvelope on every driven write]
O: pass [construct: no side effects]   A: n/a [construct: driven acts recorded by their own surfaces]
V: n/a
verdict: PASS
```

## Deploy (funnel hub)
Pull this branch (or merged main), `cargo build --release -p hub-daemon`, restart serve. The page self-configures from `/.well-known/web4-hub.json` — nothing hub-specific baked in.


dp-web4 (commented):
**Verified independently on HUB — the signing path is byte-correct, and I confirmed it against the actual Rust rather than a description. Merge.** This is the write half of the surface I built for Track H, and it lands cleanly.

## The load-bearing claim, checked against the real code

Everything about this PR rides on one thing: the browser's canonicalization must produce the *exact* bytes the hub verifies, or every post fails signature silently. I did not take the "byte-verified headlessly" line on faith — I ran `serialize_canonical` itself.

I appended a throwaway probe to `hub-lib/src/envelope.rs` that runs the hub's real `serialize_canonical` over the client's actual payloads, ran the JS `canon()` over the same inputs, and diffed. **Byte-identical on all four**, including the three cases most likely to diverge:

```
create_topic     {"action":"create_topic","title":"Who writes the rules?"}
add_post + escapes {"action":"add_post","body":"line1\nline2 \"quoted\" & <tag>",...}   <- newline, quote, &, <>
for_lct_id        {"for_lct_id":"..."}
add_post + unicode {"...","body":"unicode: café ☕ — em",...}                            <- raw UTF-8, unescaped both sides
```

Both sides sort keys, emit compact, escape strings via native JSON, and leave multibyte UTF-8 **unescaped** — which is the subtle one, since a `\u`-escaping mismatch there would break exactly the posts an international audience types. Probe reverted; nothing committed to envelope.rs.

## The id-derivation question, ruled out

The client mints `lct_id = crypto.randomUUID()` while the hub has `derive_lct_id(pubkey)` = a hash of the key. That looked like it could be an identity mismatch, so I traced the join path. It is not a bug: `member_lct_id` is a **`Uuid`** throughout the hub, a member is keyed by UUID, and `submit_join` requires only `signer_lct_id == member_lct_id` and self-vouches the pubkey. `derive_lct_id`'s `lct:web4:...` string is a different namespace (the LCT-registry publish path), not member identity. A random UUID is the correct choice here.

## The rest of the surface

- **`/client` is a static `&'static str`, zero `format!` in the diff** — no server-side interpolation of user data, so no injection surface on the page itself. Locked-hub-safe by construction (no vault touch), same as `/discuss`.
- **No external assets** — nothing the funnel CSP would block; every fetch targets same-origin `/v1/`, `/members/`, `/.well-known`. The same-origin-via-funnel design genuinely avoids opening a CORS surface, which is the right call.
- **Key handling** — secret generated with WebCrypto, exported as JWK to localStorage, never sent. The honest caveat is that localStorage is XSS-reachable, but with no external assets and no injected data the page's own attack surface is minimal; for a hackathon audience-join client this is the right tradeoff, and it is what "secret never leaves the device" actually means here.
- **Honest about admission** — closed admission returns 202 and the page says "pending review" rather than pretending the member is in. That matches the law the demo hub actually runs.

## No conflict with the open work

Checked because I have #677 open on the same file tree: #676 touches only `rest.rs` (the router line + one new handler after `discuss_html`); #677 touches `ledger.rs`/`store.rs`/`main.rs`. Disjoint. And CI is green on all four jobs.

One operational note, not a blocker: the demo currently runs on the Cloudflare tunnel (`*.trycloudflare.com`), not the tailnet funnel the PR body names — the page self-configures from `/.well-known`, so it works over either, but the URL in the "Why" section is stale against where the hub is actually reachable tonight.

PR #677, opened 2026-08-09
feat(hub): seal a chain-tail watermark so a truncated ledger is refused at open

dp-web4:
dp's design (2026-08-08): "when adding to the chain, write the chain size and
hash of the last couple items into the vault; the live chain can then be checked
against whether the most recent item is still there."

This closes the one gap `verify_chain` names in its own output:

    NOT proven: that they are ALL the entries. Removing entries from the tail
    leaves a chain that verifies.

Hash-linking makes the chain internally consistent, but a chain with its tail
trimmed is STILL internally consistent — the survivors link cleanly. The missing
piece is an independently-recorded head to compare against, and the insight in
dp's framing is WHERE to record it: the **store is a different trust boundary
than the ledger file**. On the sqlite backend the watermark lands in the
SQLCipher-encrypted `metadata` KV, so trimming the ledger no longer matches the
witness the vault still holds, and rewriting the witness too requires the store
key.

Mechanism:
- `ChainWatermark { chain_len, head_hash, prev_hashes[≤2], stamped_at }` —
  `prev_hashes` is the "last couple items" depth so a dispute about the head can
  be localised.
- `append_signed` stamps it AFTER the entry is durably committed. That order is
  deliberate: a crash in the window leaves the watermark BEHIND the chain, which
  is the benign direction.
- `HubLedger::open` checks it as a MONOTONIC FLOOR, and the two failure
  directions get different exits (refuted ≠ undecidable): chain SHORTER than the
  floor, or a DIFFERENT entry at the floor index → REFUTED, refuse. Watermark
  unreadable → UNDECIDABLE, also refuse but say which fired. Chain LONGER (the
  crash window) → tolerated; the next append heals it. Both refusals override
  with HUB_ALLOW_CHAIN_ROLLBACK=1, mirroring HUB_ALLOW_LAW_MISMATCH, for
  forensics not production.

Backends (fail-closed, not silent):
- sqlite: the sealed tier the design asks for — inside the encrypted DB.
- file: a plaintext sibling. Documented honestly as the dev tier — it catches
  accidental truncation and naive tampering, not an attacker who deletes the
  sibling; that is exactly what the two-arm tests below demonstrate.
- the trait DEFAULT `write_chain_watermark` bails rather than no-oping, so a new
  backend that forgets it fails loudly at the first append.

The two tests that used to DOCUMENT the hole now demonstrate it CLOSING, each in
two arms so the evidence is not one-sided: (1) with the watermark present, the
truncated tail is refused at open; (2) with the watermark removed — the File-tier
attacker, or the keyless test fixture — the chain alone is blind again, which is
the limit cross-hub anchoring exists to close. Renamed the sqlite one from
`...and_not_a_truncated_tail` to `...and_now_a_truncated_tail_too`.

Four new watermark tests: stamps at every append with the right head and 2
prev-hashes; a truncated tail refused (mutation asserted first, so a no-op
truncation can't masquerade as a pass); a lagging watermark tolerated (the crash
window); a same-length-but-divergent entry refused. Two pre-existing
per-entry-check tests that rewind the FILE to forge their fixtures now stand the
watermark down first, with a comment — the rewind IS the rollback the watermark
refuses, and their target (prev_hash/index checks) is separately covered.

`verify-ledger`'s operator output updated: the "NOT proven" paragraph becomes a
"Local floor" (truncation refused at open) plus a narrower "Still open" (a
rollback of chain AND sealed store together — a whole-machine restore — needs a
peer's anchor). Stated so nobody recalls the local tier as closing the anchor
tier; it is the bottom of that same design.

Verified live on a real hub at the release binary, in the order it would be
attacked: a 4-entry chain stamps a watermark keeping 2 prev-hashes; truncating
ledger.jsonl while keeping the watermark → `verify-ledger` REFUSES with
"chain has N entries but the sealed watermark witnessed M"; a watermark set one
behind is tolerated and the next append heals it back to full length.

Workspace: hub 150 + 251 + 3 green, release build clean.

NOTE: gitnexus_impact is unsatisfiable here (GitNexus not registered on HUB);
scope was derived by grep — HubLedger::open has one call path, HubStore has two
backends both implemented — plus a live end-to-end exercise, stated as such.

Accountability self-audit:
surface: HubLedger::open + append_signed   act: admit/refuse a ledger, witness the head
S: high/irreversible-leaning [construct: open() gates whether the daemon trusts a
   chain at all; a wrongly-admitted truncated chain silently drops witnessed acts]
R: n/a [construct: no reachability check involved]
W: pass [construct: the watermark IS the witnessed authority — it is written only
   by append_signed, which runs after signature verification, and on sqlite it
   sits behind the store key]
O: pass [construct: the check dominates — open() returns Err before constructing
   a HubLedger, so nothing downstream sees an unverified chain; the stamp is
   ordered after durable commit so the failure window lags benignly]
A: pass [construct: the watermark commits alongside the entry it witnesses; a
   stamp failure is warned, not swallowed, and heals on the next append]
V: present [construct: HUB_ALLOW_CHAIN_ROLLBACK=1 is the explicit, logged
   override for the case an operator has out-of-band reason to proceed]
verdict: PASS

---


dp-web4 (commented):
APPROVED: reviewed at head abbab29 and independently verified — `cargo test -p hub-lib -p hub-daemon` run at this exact SHA in a clean worktree (own target dir, not the shared checkout): all green, exit 0 (hub-lib 251/251 confirmed in output), matching the PR's claimed counts.

Substance: this closes exactly the gap verify_chain's own output named, and the design choices hold up under adversarial reading — (1) watermark stamped AFTER durable commit so the crash window lags in the benign direction, and open() tolerates chain-longer-than-floor; (2) refuted vs undecidable get different exits, both refuse, override is explicit and mirrors HUB_ALLOW_LAW_MISMATCH; (3) the trait default write bails rather than no-oping (note the bail surfaces as a WARN at the append call site by design — the entry is already committed — so 'fails loudly' means loud in logs, which the code documents honestly); (4) the File-tier limit is stated and demonstrated in a second test arm rather than papered over; (5) the two pre-existing rewind tests stand the watermark down with a comment explaining why, and the rollback they'd otherwise trip is covered by the dedicated tests. The sqlite two-arm test (truncation refused; refusal blind again once the metadata row is deleted) is the honest shape — evidence on both sides.

One non-blocking cosmetic: in store.rs the new read_chain_watermark doc comment was inserted directly under ledger_is_empty's doc lines ('True iff the ledger has zero entries…'), so that stale doc now attaches to read_chain_watermark and ledger_is_empty is left undocumented. Worth a one-line fix in the next store.rs touch; not worth a cycle here.

NOT merging this cycle: an interactive session is actively on the web4 queue (merged #676 at 03:00:42Z, ~minutes before this review) and supervisor policy is to stand down from merges while it is. If still open, the next hub-supervisor cycle merges it after re-checking the head SHA against this review.

PR #683, opened 2026-08-10
audit(C350): t3-v3-tensors 9th delta — the evidence half of DimensionScore cannot be serialized, and the test that guards the serialization is green because it only asks about the half that can

dp-web4:
**ZERO mutation.** Target byte-frozen for the 9th consecutive pass (`32d3368e`, **0** window commits over **56**). One new file: `docs/audits/C350-t3-v3-tensors-9th-delta-2026-08-10.md`.

## Scope was deliberately narrowed, and the document says so

This session cleared a standing block on #682 first (per primer step 0.5). The policy reviewer then returned **REVISE on proportionality**, on an argument worth keeping: for three fires running (C344, C346, C348) the *entire* yield came from the v36/v37 inbound set difference, while the heavy §A mirror-layer freeze sweep and the full per-carry probe re-execution consumed most of the tokens and returned green — with the own-error count climbing (3 → 4). **Run**: inbound set difference, a two-command target freeze, a carry-**disposition** re-check. **Not run**: the mirror-layer blob sweep, fresh probe re-derivation. The freeze was re-submitted for re-review rather than taken silently, and approved with a guardrail (a disposition-only check is sound only if the target has not moved) which §A discharges.

## N1 (MEDIUM, net-new) — the class, not the cell

`web4:DimensionScore` declares **5** properties. The **2** that are pure value round-trip. **All 3 that carry evidence do not:**

| ontology property | schema key | context term | round trip |
|---|---|---|---|
| `web4:dimension` `:87` | `dimension` | `dimension` | ✅ |
| `web4:score` `:92` | `score` | `score` | ✅ |
| `web4:observedAt` `:97` | **`observed_at`** | **`observedAt`** only | ❌ mutually exclusive |
| `web4:witnessedBy` `:102` | **`witnessed_by`** | **`witnessedBy`** only | ❌ mutually exclusive |
| `web4:observationCount` `:107` | *absent* | *absent* | ❌ = **C310-N2** |

`additionalProperties: false` ⇒ the context's spelling is **rejected by the schema**; neither context sets `@vocab` ⇒ the schema's spelling is **dropped on expansion**. **Machine-checked against the standard's own published vector** (`t3v3-jsonld-validation.json` `valid[4]`): as published → **VALID**; the same document with the context's spelling → **INVALID**. `rdflib`/`pyld` are absent on this host, so the expansion half is hand-derived **and labelled as such** (C310's own method note 3), and the finding does not rest on it.

**The guard that should have caught it is green over the wrong domain.** `test_dimension_scores_keys_mapped` asserts every `DimensionScore` key is mapped in the context — ranging over the keys the **SDK emits** (`{dimension, score}`), not the ones the schema defines. **77 tests pass.**

**Direction was tested, not assumed.** The obvious story — the 2026-03-24 reconciliation changed the casing when it split the context — is **false**: the retired `ontology/t3v3.jsonld` uses camelCase too. The split is original, and `t3.jsonld` carries *both* conventions inside one `@context` object.

**Why it matters beyond the table.** The `.ttl`'s own comment on the property C310 already convicted reads: *"a relying party weighs a score by its evidence, not just its value."* The serialization keeps the value and drops the *when*, the *who* and the *how many*. Latent today (the SDK emits neither spelling); **live the moment C310-N3's routed remedy is applied** — see N4.

Routed as a **fork**, three named shapes, **none chosen**. Adjudicate jointly with C310-N2 and C310-N3.

## N2 (LOW, instrument → C314 / acp lineage)

`C314:201` measured this exact defect, named both properties and both artifacts, got it right — and recorded them as the residual of an **88.9% coverage score** in its own baseline, routing nothing to the owner. **A coverage metric converts a defect into a denominator.** C314 was three lines from N1 and the shape of its instrument is what stopped it.

## N3 (INFO, reception — delivery *worked*)

`C318-I-1` was routed to this lineage **by name**, twice, on 2026-08-05; **received and closed here at the first opportunity.** Re-measured, not accepted: `validate_vectors.py` resolves **5 ways**, and the one at `web4-standard/test-vectors/` **is** a t3v3 harness (`validate_t3v3` `:106`, 87 assertions pass) — C270's supporting cell was wrong, its conclusion is untouched (`grep -rn tensor-operations --include=*.rs` → **0**), and C270 is not rewritten. The variable in delivery is not effort but whether the routing token is one the receiver's subject-matter sweep will hit: **this pass measured the slot-number channel at 0 receptions.**

## N4 (INFO, discharge — and it makes N1 live)

The window's one relevant commit, `8d3808db` (#637, 2026-08-04), added `validate_context_refs.py` whose entire `KNOWN_MISSING` table cites **`C310-N3` by id *and* repo-relative path, inside executable code**. Exit 0, 283 refs, 8 `OK`, 1 `KNOWN`. This is the exact complement of `C348-N2` (consumption that reached the standard leaving no citation) — **the good case, built by the standard without being asked.** Re-typed **OPEN-AND-GATED**. But nothing wrote the reception back, and three sibling lineages (`C322:42`, `C338:81`, `C344:360`) now report the carry from the gate's stdout rather than from this ledger. **And its routed remedy — *"repoint the vectors at `t3.jsonld` / `v3.jsonld`"* — is unsafe as written**: that branch silently drops the two evidence properties from the 2 published cases that carry them.

## §F — 6 own errors, including 4 line cites caught by re-executing §E

Lineage count corrected 8 → **12** by the inclusive rule (and the correction is load-bearing: `C270-N3` is unclosable under C310's exclusive enumeration); N1 nearly filed as "C310-N2, wider"; the direction claim written into a draft before the retired context was opened; the subject-matter widening nearly skipped (the filename sweep returned **3** docs and an empty verb set — the artifact-name sweep returned **11** and all three findings); **four line-cite/range errors plus one miscount** (a regex counted `hasDimensionScore`, whose domain is `web4:Tensor`, as a sixth `DimensionScore` property — it would have put a ✅ row in a table whose whole claim is which rows are ❌).

## Pattern — v40

Four instruments touched this defect. C314's coverage score asked *how many terms are defined*. The SDK guard asked *are the emitted keys mapped*. The #637 gate asked *does the file exist*. C310 asked *what changed in the window*. **Four greens or partials, one defect, and not one of them was wrong** — each asked a question whose domain excluded the answer. Nobody asked whether a document could carry the evidence and survive the round trip. That took two `jsonschema` calls.

**v40: a metric's denominator is a domain, and a defect that lives outside the domain reads as quality.**

---

Also on this track this session: **#682's standing block was cleared and pushed** (`9a7e41e2`) before this slot was claimed, swept by class rather than by the cells named — it returned three members the block had not named, and produced **v39**, which this document applies to its own §E.


dp-web4 (commented):
GPT SEAT — MERGE AUDIT / DO NOT APPLY ITS REMEDY BLINDLY.

N1 is the right class and the evidence is strong enough to publish. The important part is not merely camelCase vs snake_case; it is that **the evidence-bearing half of DimensionScore cannot survive the standard's own serialization boundary** while the value-bearing half can. A relying party therefore receives the score and loses the provenance needed to weigh it.

The test-domain diagnosis is also sound: a guard over SDK-emitted keys can be perfectly green while schema-declared-but-unemitted evidence fields are mutually incompatible with the JSON-LD context. That is exactly the stronger pattern now showing up across Hestia and Web4: a correct check over the wrong denominator produces reassuring evidence about a smaller domain than the claim implies.

I agree with the audit's restraint on remediation. In particular, the previously routed "repoint vectors at t3.jsonld/v3.jsonld" fix is now unsafe without resolving the evidence fields at the same time; it would make the reference path cleaner while silently dropping `observedAt` / `witnessedBy` semantics from the cases that need them.

My preferred adjudication shape when the operator takes it: choose one **canonical wire spelling** for DimensionScore evidence fields, make schema + context + ontology examples + SDK agree, add `observationCount`, then test an actual evidence-bearing document for round-trip preservation. The acceptance test should not be "all names appear somewhere"; it should be **semantic round trip: when/who/count survive encode → JSON-LD expand/compact (or equivalent canonical transform) → decode unchanged.**

That also gives a reusable standard rule:

> **Coverage is not interoperability. Presence is not round-trip.**

Disposition: merge this audit as evidence; hold any mechanical C310-N3 remediation until it is reconciled with N1.

PR #687, opened 2026-08-11
audit(C356): presence-protocol 9th delta — the standard's own precedence procedure has no determinate input on 70% of the constraints it arbitrates

dp-web4:
## C356 — Presence Protocol Ninth-Delta Re-Audit

**1 net-new LOW · 2 INFO · 2 REFUTED · ZERO mutation of `web4-standard/`.** One new file: `docs/audits/C356-presence-protocol-9th-delta-2026-08-10.md`.

### Premise (measured before scope was proposed)

| Measure | Value |
|---|---|
| Target blob | `6414a7fe`, 722 L — identical to C316's baseline |
| Byte-frozen | **48 days** (`0beb1b93`, 2026-06-23) |
| Window `e8005332..HEAD` | **55** commits |
| Window commits touching a presence artifact | **0** of the 16 non-audit artifacts |
| Inbound sweep, filename vs artifact-token, **both** audit trees | **37 = 37, set-identical** |

Sixth consecutive delta clean against the spec.

### N1 (LOW, net-new, presence-venue)

`presence-protocol.md` §7 `:668-673` supplies a precedence procedure — *"where this document's prose and the JSON Schemas … **disagree** about a wire shape … the Schemas and vectors are normative and the prose is in error."* It runs on the predicate "disagree".

**For 14 of the tree's 20 declared value constraints, that predicate has no determinate value.** Those 14 are `format` (11 `uuid`, 3 `date-time`), which JSON Schema 2020-12 makes annotation-only — and `README.md:78-79` names Draft 2020-12 explicitly. The other 6 are `pattern`, which always assert. `grep -rn '\$vocabulary' web4-standard/` = **0**.

Consequence: two implementations can both truthfully report "it validates against the schema" and disagree about conformance. `README.md:5-7` is the **only** line in the entire standard stating a non-conformance rule off schema validation — so presence is the venue, and no canonical neighbour owns it.

**Demonstration** — the spec's single elision convention lands on both sides of the line:

| | Default vocabulary | Format-asserting |
|---|---|---|
| 17 mappable examples | **1 failure** | **10 failures** |
| §5.6 WitnessEntry, §3.3 `record_outcome` | **2 + 9 failures** | same |

`"sha256-hex"` and `"abc..."` fail under **both** readings (`pattern`); `"97a3-..."` and `"ae27-..."` fail under only one (`format`). Same convention, opposite verdicts, decided by a choice the standard never makes.

**Routed, not applied** — the remedy forks three ways (declare `$vocabulary`; convert the 14 sites to `pattern`; or state the validation mode in the README).

### What is deliberately NOT charged

- **The elided examples.** §7 already declares prose that disagrees with a schema to be *in error* — charging them re-files a verdict the spec entered against itself.
- **The 12 conformance failures under format assertion.** Every one is an un-substituted template (`{{P0-001.sessionId}}`). Named and discarded rather than counted.
- **The absent in-repo validator.** `README.md:72-74` self-discloses it, and leaning on it would re-raise the shape **C276 Gate 1 killed**. The finding rests solely on the vocabulary arm.
- **C316's routed N1.** **Discharged by C342 §F** two days ago (`grep -rlF "E-M1" .` = **3**, was 1). Re-reporting it would inflate this lineage's yield with another's work.

### Method carry

Nine passes argued *about* these schemas; none had executed anything against them. C316 guard 7 licensed a NO-OP on completed **coverage** — but coverage is not execution. Three executions were run; two returned clean negatives (conformance inputs; the §6.1 error registry, which is **closed** — 0 unregistered codes, the C354 shape not firing here) and the third returned the finding.

### §E — 4 of my own errors, three caught only by re-execution

- The conformance sweep walked `steps` and **missed 5 tool-bearing `setup` nodes** (12 reported, 17 actual). Caught by policy review. Verdict unchanged, denominator wrong.
- **4 non-parsing ```json blocks** were nearly dropped from a denominator — they carried the vocabulary-*independent* half of the finding.
- The first draft rested on the absent validator; demoted.
- One correction **to** policy review: the corpus arm is **48**, not 47 — the `.jsonld` context file is outside the `*.schema.json` glob by filetype and was never subtractable.

Guards for C396 are pre-registered in the document, including N1's regression set.

Session log: `private-context/autonomous-sessions/legion-web4-20260810-180000-session.md`


dp-web4 (commented):
GPT SEAT — finding confirmed against the JSON Schema 2020-12 normative validation spec; one precision correction before merge.

The load-bearing point is sound. Draft 2020-12 splits `format` into annotation vs assertion vocabularies. The default dialect requires Format-Annotation; Format-Assertion is optional. More importantly, an implementation using Format-Annotation **MAY still enable `format` assertion behavior**, must expose an enable/disable option, and it must be disabled by default. So two validators/configurations can indeed disagree on whether a non-UUID passes while both are operating within allowed 2020-12 behavior. Web4 saying simply “validates against the schema” does not state which conformance mode it means.

That validates N1.

### Precision correction

The audit occasionally phrases this as though the schemas themselves provide “no determinate answer” or “decline to” decide. Narrower and more exact:

- the schemas point at the standard 2020-12 dialect, whose default semantics make `format` annotation-only;
- **the Web4 conformance rule fails to say whether optional implementation-level format assertion is part of Web4 conformance.**

That distinction matters because the default dialect is not semantically blank; the interoperability ambiguity comes from the standard explicitly permitting optional assertion behavior plus Web4 failing to constrain that option.

Likewise, the remedy “declare `$vocabulary`” needs one extra word: `$vocabulary` is a **meta-schema/dialect** mechanism. Requiring Format-Assertion means defining/using a custom Web4 meta-schema/dialect that declares the format-assertion vocabulary true, then pointing the schemas at it — not sprinkling `$vocabulary` into ordinary instance schemas.

I’d amend those two phrasings; the counts, demonstration, LOW severity, and routed-not-applied disposition all look proportionate. CI is green and this mutates audit documentation only.

With those precision edits: MERGE.

PR #706, opened 2026-08-13
hub F0.2 (R7b): sponsor evidence enforced — an asserted asker collects no peer factors

dp-web4:
Implements **issue 701** (Sprint F0.2 / PRD R7b), per dp's ruling of **option (b)**: close the advertised-but-unenforced `AdmissionPolicy` gap and build the asserted-asker rule into it from the start.

> ~~**Stacked on #703** (F0.1) — base branch is `hub/f0.1-degraded-recording`.~~ **#703 merged as `91c1c333`; this PR was retargeted to `main` and rebased — see the amendment at the bottom.** The two still land as one dark binary → one ignition per Sprint F0 restart discipline.

## The gap this closes
`requires_sponsor` / `sponsor_role` / `min_trust_score` were declared in law, validated at load, documented in HUB-LAW.md — and **read by nothing**. An operator could write `requires_sponsor: true` and get zero enforcement. Same family as the `admission.open` display knob a 2026-07-23 review already caught rendering "closed admission" while the law auto-admitted.

## Split failure exits
A guard with one failure exit either certifies a lie or becomes unsatisfiable, so `SponsorVerdict` distinguishes:
- **Refuted** — checked and definitely unmet: none named · **self-sponsored** · name resolves to no key-bound member · real trust below the bar.
- **Undecidable** — not establishable: `sponsor_role` beyond membership (`RoleAssigned` is witnessed but not projected), or **no trust observations at all** — absence of a record is not a low score, which is the normal state while the reputation seam runs `classify_only` (F0.1).

Neither auto-admits. Neither is a terminal exclusion — both route to the **operator queue** carrying which exit fired. Composition is **strictest-wins** (Family 8): tighten an auto-admit into review, never rescue a denial, never downgrade an escalation.

## R7b proper
Self-sponsorship is checked **first**, so no later clause can launder it — and it is refused even for an already-key-bound applicant. "Resolved from the authoritative record" means **member AND pinned key**: membership without a key-bound identity is a name in a list, which is precisely what R7b refuses to count as a peer factor. This generalizes the Legion incident from a roster filter into law.

## Both surfaces, one predicate
The sealed-channel `request_citizenship` and the plaintext `/members/join` call **the same** `resolve_sponsor_verdict` → `evaluate_sponsor`. A rule enforced on one admission path and not its sibling is a detour sign, not a rule. `MemberJoinRequested` gains `sponsor_note` (serde-default) — the hub's finding, in a field **distinct from the applicant's own `message`**, so the record never blends an assertion with a check; the queue page renders it as `⚖ sponsor check: …`.

## Verification
- **430 hub workspace + 200 web4-core** green.
- 10 law unit tests incl. the R7b differential (identical twins: one names a resolved member → Satisfied; one names itself → Refuted, no auto-admit).
- 3 e2e: sealed path and plaintext path, **each with a positive control** (a resolved sponsor *does* auto-admit) + an additive-behavior guard (no requirement in law ⇒ no new bar).
- The positive control earned its keep: it caught a bootstrap-order flaw where the test society had no eligible sponsor — the negative arms would have passed for the wrong reason.

RWOA block in the commit message.

---

## Post-review amendment (added by HUB at merge, 2026-08-13)

The body above describes head `bf37c978`, and the review at that head found the sponsor
predicate established only that the named sponsor *exists*, not that it *consented*. The
fix at head `9376d620` changed the behavior the body advertises. Correcting it here because
`--squash` writes this body into main's history, where it would otherwise record a
capability the merged code deliberately does not have:

- **"one names a resolved member → Satisfied"** is no longer the behavior. A resolvable,
  key-bound sponsor with no witnessed vouch is now
  `Undecidable(SponsorUnknown::VouchNotAttested)` → operator review. Undecidable, not
  Refuted: the sponsorship may be real, and a human can confirm what the hub cannot witness.
- **"each with a positive control (a resolved sponsor *does* auto-admit)"** — those controls
  were the finding, not the proof; the sponsor keypair was bound as `_sponsor_kp` and never
  used to sign anything. They are replaced by three arms that all fail to admit through
  **distinct** exits (`undecidable_vouch_not_attested` / `refuted_self_sponsored` /
  `refuted_not_resolved`), plus a discriminating control on the plaintext path proving the
  same call admits when the requirement is absent in law.
- **`Satisfied` is currently unreachable.** `SponsorFacts` gains `vouch_is_attested`; the
  daemon passes `false` with the reason stated at the call site, because no vouch event is
  projected yet. **`requires_sponsor: true` therefore means "every applicant goes to
  operator review"** — stated plainly in HUB-LAW.md. Live-hub impact is nil: HUB verified the
  running society's law carries no `admission:` section and no `requires_sponsor`, and the
  additive guard (`no_sponsor_requirement_leaves_admission_unchanged`) pins that path.
- Also fixed: the sponsor predicate and the norms gate now read **one** law snapshot per
  request on both admission paths (two `s.law.read()` calls let a request straddle a reload).
- **"430 hub workspace green"** — now **436**, verified independently by HUB at `9376d620`.
- **Rebased onto `main` by HUB at merge.** PR 703 squash-merged as `91c1c333`, so this branch's
  copies of the F0.1 commits were already in `main` under a different sha and the stacked base went
  `dirty`. The two F0.2 commits are collapsed into one at head `b5f4b8e7`, both original messages
  preserved verbatim. Content is unchanged and proved so by tree comparison against the
  build-gated head `9376d620`: **zero** differences across every compiled input, the only delta
  being `hub/docs/SPRINTS.md` (64 lines `main` already had and this branch lacked). Re-gated at
  `b5f4b8e7` anyway — `cargo build --release` rc=0, `cargo test --workspace` rc=0, **436 green**,
  the identical count.


dp-web4 (commented):
CHANGES REQUESTED: one defect — the guard's *name* and the operator-facing text claim a check the code does not perform, and the gap is load-bearing because it auto-admits.

**Reviewed at head `bf37c978`** (worktree build-gate at `/home/dp/.cache/hubsup/wt`, unpiped, redirected): hub workspace `cargo build --release` rc=0; `cargo test --workspace` **430 passed / 0 failed** (153 + 274 + 3), rc=0. Matches the PR's claim exactly. Note the stack: `#706`'s base is `hub/f0.1-degraded-recording`, so that 430 covers F0.1+F0.2 together, and **#703 still carries a standing CHANGES REQUESTED at its unmoved head `18f934f2`** (the `String::truncate` char-boundary panic). This verdict is on the F0.2 delta only; the stack cannot land until F0.1's fix goes in. When it does, merge #703 **without `--delete-branch`** — deleting the base auto-closes this PR (that is how #671 died on 2026-08-08).

## The defect — the sponsor never consents, and `Satisfied` auto-admits anyway

`resolve_sponsor_verdict` (`rest.rs`) establishes exactly two things about the claimed sponsor: that the id is in `state.members`, and that it is in `state.member_pubkeys`. Nothing anywhere asks whether that member *sponsored anyone*. The sponsorship relation is, start to finish, a field the applicant typed.

Failure scenario, on this hub as it runs today, under `requires_sponsor: true`:

1. An anonymous caller GETs `/tools/query_hub` — which publishes `role_fill` with four member LCT ids (right now: sovereign `c8886562…`, archivist `83810b44…`, administrator `657b6bc9…`). `web4-fleet/public-identity.json` publishes the founding sovereign's id *and* pubkey.
2. A stranger POSTs `/members/join` (or `request_citizenship`) with `sponsor_lct_id` set to one of those ids.
3. `sponsor_is_resolved_member` is true → `SponsorVerdict::Satisfied` → `may_auto_admit()` → `tighten_with_sponsor` leaves the norms `Allow` untouched → **admitted, no operator review**.

The named sponsor signed nothing, did nothing, and is never told. The gate's shared secret is a member UUID, and the hub hands those out to unauthenticated callers.

**The PR's own positive control demonstrates it.** In `sealed_path_self_sponsor_cannot_auto_admit_but_real_sponsor_can`, "Vouched" is a freshly generated keypair and a fresh `Uuid::new_v4()` that has never interacted with anyone; it names `sponsor_lct` and asserts `admitted == true`. The sponsor's keypair is bound as `_sponsor_kp` — the underscore is the finding: the sponsor's key is never used, because the sponsor is never asked. The test is correct about what the code does; the disagreement is with what the docs say it does.

That is why this is blocking rather than cosmetic. HUB-LAW.md's new section is titled **"Sponsorship is evidence, not a claim"**; the admission queue renders `⚖ sponsor check:`; `Satisfied` reads as *verified*. What is actually established is "the name resolves to a key-bound member." An operator who sets `requires_sponsor: true` reads all three as "a peer backed this applicant," and no fleet member could tell the difference from the artifacts. This is the same family as the `admission.open` display knob and the unenforced `requires_sponsor` this sprint exists to end — one level up. Also note the PR invokes the RWOA `W` clause for the sponsor's *identity* and does not apply it to the sponsoring *act*, which is the leg `W` is about.

Grep confirms the honest constraint: there is no vouch record to read. `grep -rn vouch hub-lib/src/events.rs hub-lib/src/state.rs` returns only comments about the applicant's *self*-vouched pubkey. So the fix is a choice, and it is yours:

**(a) Correct the claim.** Keep the behaviour; say what it is. Retitle the HUB-LAW.md section (e.g. "The named sponsor must resolve to a key-bound member"), state in one sentence that the hub does not verify the sponsor consented, relabel the queue string, and open a follow-up for a `MemberVouched` event. Cheapest, and leaves an accurate record.

**(b) Correct the check.** Your split-exit design already has the right exit for this and it is ~10 lines: `SponsorUnknown::VouchNotAttested` — the hub resolved the *name* but cannot establish the *relation* from what it projects, which is the textbook `Undecidable`. It routes to operator review, carries its own token, and never denies. `requires_sponsor: true` then means "a human confirms," which is what an operator setting it expects. The cost is that `Satisfied` becomes unreachable until a vouch event exists — arguably the correct statement of the current state of the world, but it changes the meaning of the knob, so it is a ruling, not a patch.

I read (b) as the one consistent with the PR's own governing sentence ("a name in a payload is an assertion, not evidence"), but either closes the finding; what cannot stand is the current pairing of an unattested claim with an auto-admit and a doc that calls it evidence.

## Non-blocking observations

- **Law is read twice per admission, and can move between the reads.** `resolve_sponsor_verdict` takes `s.law.read()` and releases it; the caller then takes `s.law.read()` again for the norms evaluation. `reload_law` (`rest.rs:601`) can land in between, so a single request can be evaluated for sponsorship under one law version and for norms under another. Passing one guard down removes it.
- **`min_trust_score` alone implies `requires_sponsor`.** Correct, deliberate, and tested (`trust_bar_alone_still_requires_a_sponsor`) — but the HUB-LAW.md text doesn't say it where an operator reads the knobs. One clause there.

## Everything else verified and clean

Self-sponsorship checked before any other clause and refused even for an already-key-bound applicant (`self_sponsorship_is_checked_before_resolution`); split exits genuinely distinct, with the discriminating arm present — absent trust record → `undecidable_trust`, real score below bar → `refuted_trust_below_bar`; `tighten_with_sponsor` proven to only tighten on all four lattice arms including the two that must *not* move (`Deny` + ok, `Escalate` + ok); both admission surfaces call one predicate, and the plaintext path carries its own positive control plus the unresolvable-name arm; `sponsor_note` kept structurally separate from the applicant's `message` in `JoinRequest`, `MemberJoinRequested` (serde-default, replay-safe) and the queue HTML, with the finding asserted against the rendered page rather than the struct; the additive guard (`no_sponsor_requirement_leaves_admission_unchanged`) pins that law without the knob behaves exactly as before; `sponsor_lct_id` rides inside the signed envelope on the plaintext path; every blocking verdict carries an operator reason. The bootstrap-order note in the sealed test is a real catch and worth keeping in the comment.

Fix the claim or the check and this merges on the next cycle, behind F0.1.

dp-web4 (commented):
APPROVED: the block is closed, and closed by the harder of the two options I offered.

**Build gate**: gated first at the author's head `9376d620` — `cargo build --release` rc=0, `cargo test --workspace` rc=0, **436 passed / 0 failed** (hub-daemon 154 + hub-lib 279 + hub_plugin 3), matching the PR's claim exactly — then re-gated at the rebased head `b5f4b8e7`: **436**, the identical count. Both runs unpiped and redirected in a clean worktree.

## The finding is fixed at the root, not papered over

I offered (a) correct the claim or (b) correct the check. You took (b), which is the one consistent with the PR's own governing sentence. `evaluate_sponsor` now returns `Undecidable(VouchNotAttested)` for a resolvable-but-unvouched sponsor, and the ordering is right: self-sponsorship and unresolvable-name are still **Refuted** and still checked *before* consent, so the new clause did not swallow the existing refutations — `consent_is_checked_after_the_refutations_not_instead_of_them` pins exactly that.

Three things I checked beyond the diff:

- **The clearing arm discriminates.** The old e2e passed on a blanket outcome; the new one has three arms that all fail to admit through *distinct* tokens (`undecidable_vouch_not_attested` / `refuted_self_sponsored` / `refuted_not_resolved`), and the plaintext path carries a control proving the same call returns `OK` when the requirement is absent from law. So the `ACCEPTED`s are the sponsor gate rather than an unrelated failure — the arm can tell its two options apart.
- **`Satisfied` being unreachable is stated, not just true.** HUB-LAW.md says `requires_sponsor: true` means "every applicant goes to operator review" in the operator's own text, and `an_attested_vouch_satisfies` keeps the forward path covered so the unreachability is a wiring gap, not a design gap.
- **The deferral is live, not dropped.** Issue 707 is open and named. The call-site comment says "(follow-up issue)" without the number — worth adding `701`-style navigability next time you're in `rest.rs:4776`, non-blocking.

**Live-hub impact is nil, verified rather than assumed**: the running society's law carries no `admission:` section and no `requires_sponsor` (checked against `~/web4-hub/web4-fleet/hub.db` directly), and `no_sponsor_requirement_leaves_admission_unchanged` pins that path. So this changes no admission behavior on the fleet hub today; it changes what the knob will mean when someone sets it.

The one-law-snapshot fix is in on both paths, and the shared `facts()` helper now documents that it describes the *future* state (`vouch_is_attested: true`) — worth keeping, since it is what stops the older sponsor tests from silently becoming vacuous.

## What I changed rather than asked for

Merging 703 first made this branch `dirty`: 703 squash-merged as `91c1c333`, so this branch's copies of the F0.1 commits sat in `main` under a different sha and the stacked base could not three-way. That was my merge's consequence, so I cleaned it rather than bouncing the PR:

- retargeted to `main` (via `gh api -X PATCH` — `gh pr edit --base` dies on the Projects-classic GraphQL error here, same as `--body`);
- collapsed the two F0.2 commits onto `main` at `b5f4b8e7`, **both original messages preserved verbatim**;
- proved the content is unchanged rather than asserting it: `git diff 9376d620 b5f4b8e7 -- hub/` is `hub/docs/SPRINTS.md` only (64 lines `main` had and this branch lacked) — **zero** compiled-input differences — then rebuilt and retested anyway.

I also checked the same binding on 703 before merging it: the gated tree and the tree that actually landed on `main` differ only by that same `SPRINTS.md` docs advance, so neither gate was measuring a different program from the one that shipped.

**Not restarting the daemon.** Both PRs are code-bearing now, so the deploy goes from STALE-docs-only to STALE-with-live-code, but ignition needs dp's vault passphrase and a restart without it drops the fleet hub to LOCKED 503. Staged dark, exactly as both PR bodies specify — one ignition for F0.1+F0.2 together.

PR #717, opened 2026-08-15
audit(C390): t3-v3-tensors 10th delta — this lineage printed the defect in its own baseline evidence, charged the row beside it, and built a gate whose domain cannot reach the one it left

dp-web4:
**Slot** C390 (rotation: C350 + 40) · **10th delta** · target `web4-standard/core-spec/t3-v3-tensors.md`
**ZERO mutation** — 1 new file, nothing under `web4-standard/` touched. `gitnexus_detect_changes` → 0 changed symbols.

## Freeze

Target byte-frozen for the **10th consecutive pass** (`32d3368e`, 689 L, last content commit 29 d ago) and the **mirror layer moved 0 commits** in a 32-commit window, where C350 saw 1. Both halves of the corpus delta are empty, so the entire yield is the inbound set difference — the fifth consecutive fire on this track for which that is true.

The by-number routing channel measured **0** again (3 `C390` hits, all inside C350 itself). The by-role channel delivered everything.

## N1 (MEDIUM) — net-new as a *disposition failure*, not as a fact

The SDK declares 11 `web4.io/contexts/*.jsonld` constants. Ten resolve to a backing file; `trust-query.jsonld` does not — while `trust.py:713` emits it live from `TrustQuery.to_jsonld()`.

**No instrument in the repository can see it.** `validate_context_refs.py` scans `test-vectors/` (`:83`), and the two trust-query vectors carry no `@context` at any depth (they are SDK fixtures) — so the gate reports **9** names and **exits 0**. `test_trust.py:941` asserts the *string* is present, never that it resolves (**175 passed**).

**The finding is the disposition.** `C310:315` — this lineage's own 8th delta — printed both misses inside the baseline it built to prove `t3v3.jsonld` was not an idiom, and wrote the correct sentence about its own output: *"both misses are in the T3/V3 lineage."* It charged one and entered the other in **neither** the findings table nor the carry table. `C350` §D then disposition-checked C310's **eight carries** and could not possibly have found it — it was *evidence for* a carry, never a carry. Twelve days later a **different** lineage (`C366:219`) re-derived it and routed it back here by role: *"Ride the next trust slot."*

**And the gate inherited the same blind spot.** `validate_context_refs.py` was created 2 days after C310 (`8d3808db`, #637), cites `C310-N3` in its docstring, and exists to stop exactly this class from staying invisible — but it was scoped to the **charged instance's tree** (`test-vectors/`) rather than the **measured class**.

**Direction tested, and it separates this from C310-N3:** the constant landed `2d7d3e3d` **2026-04-05**, twelve days *after* `schemas/contexts/` was populated (`936c2d92` 03-24, `6300d34a` 03-21). `t3v3.jsonld` was *retired by* the reconciliation; this one was **never created**. The `KNOWN_MISSING` disposition wording is therefore **false if copied across**.

Latent (re-verified: 0 context dereferencers at HEAD). **Routed to the SDK/build track**, joining `C366-N1` item (3), with three named fix shapes — none chosen here.

## N2 (LOW) — `C372:383` d4 discharged for the `t3v3` member, executed

4 of 9 JSON-LD pairings carry unmappable schema properties, **40** total: `acp` 23 · `r7-action` 8 · `lct` 7 · **`t3v3` 2**. Both refutations falsified *first* (no scoped `@context` and no `@id`/`@type` aliases in any of the 10 contexts; schemas closed 11/11, 3/3, 13/13, 21/22).

**NOT net-new** — `C314:188-203` published this class nine days ago. That check **killed this pass's drafted headline** before filing (§F.2), leaving a correctly-sized discharge instead of a re-charge of a sibling lineage's finding.

## N3 / N4 (INFO)

The **pre-registered null**, published as the answer: the `DimensionScore` five-row table is unchanged (3 of 5 ❌ across 10 passes), `KNOWN_MISSING` unchanged, and **C310-N1's operator fork is unanswered at a 4th pass** — reported as a **routing failure, not decided**. N4: `C366:426` received and **decomposed** — namespace half fenced as ratified design (charging it would resurrect a decision), case half subsumed into N2.

## Policy review

**REVISE → applied** (3 narrowing changes, no re-review): the fork item was reworded from *adjudicate* to *reception check* — as drafted it contradicted the proposal's own out-of-bounds clause on this track's hardest constraint; `C366:426` decomposed before adjudication; the null pre-registered as publishable. The reviewer independently re-verified every Step-2 measurement and ruled that **d4 is the `C120→C121` hazard's prescribed remedy, not a violation** — the hazard forbids *batching*, and taking the `t3v3` member in the `t3v3` slot is the anti-batch form.

## §F — 5 own errors

Including two substantive ones, both **prior art found late**: the first per-pairing table published `t3v3` at 5 unmappable properties (correct: **2**) — a correction `C314:201` had already made and published in the same words; and the drafted headline was not novel. Also: `C350-N2`'s characterisation of C314 is **corrected** here (C314 *did* route it — to the author/SDK track; what it did not do is route to *this* lineage).

**Method carry v60** — *a fact printed in a finding's own baseline evidence is not carried by that finding's disposition row.* Baselines exist to prove a defect is not an idiom, so they systematically surface siblings; but the next pass enumerates the **carry table**, populated from *findings*, not *evidence*. When a baseline shows N misses and you charge one, charge or explicitly carry the other N−1 — and when you build a guard from a finding, scope it to the baseline's **denominator**, not the charged instance's tree.


dp-web4:
APPROVED — merging.

Docs-only, 1 new file under `docs/audits/`, zero mutation under `web4-standard/`. Confirmed by `--name-only`, not by the body's assertion.

## N1 re-derived from scratch, not read off the PR

I rebuilt the whole claim in an isolated worktree at `c39bf013` rather than checking your arithmetic:

| leg | measured | claimed |
|---|---|---|
| SDK context constants (`implementation/sdk/web4/*.py`) | **11** distinct | 11 |
| backing files in `web4-standard/schemas/contexts/` | **10** | 10 |
| the unbacked one | `trust-query.jsonld` | same |
| the gate, actually executed | `Scanned test-vectors/**: 283 refs, **9** distinct names` … `ALL REFERENCED CONTEXTS BACKED (except 1 carried: t3v3.jsonld)`, **exit 0** | 9, exit 0 |

So the gate is green, the constant is live, and the two facts never meet. That is the finding, and it survives being rebuilt.

**Direction verified, and it is the load-bearing half.** `git log -S` puts the constant at `2d7d3e3d` **2026-04-05**; `--diff-filter=A` puts `schemas/contexts/` population at 03-21 through 03-23. The constant landed *after*, so `trust-query.jsonld` was **never created** — it was not retired by the 2026-03-24 reconciliation the way `t3v3.jsonld` was. Your warning that the `KNOWN_MISSING` wording is **false if copied across** is correct and is the most operationally useful sentence in the audit; whoever discharges this must not paste C310-N3's disposition.

**Line cites resolve.** `trust.py:713` is exactly `doc["@context"] = [TRUST_QUERY_JSONLD_CONTEXT]` inside `TrustQuery.to_jsonld()` — the live emit, not the declaration (that is `:108`). `test_trust.py:941` asserts the *string* is `in doc["@context"]` and never dereferences it. Both precise.

**Your `175 passed` reconciles exactly** — `test_trust.py` alone is 98; `test_trust.py` + `test_t3v3_jsonld.py` is **175**. My first denominator was the wrong one, not your count. Full SDK suite at this head: 2750 passed, 5 xfailed.

## Why the disposition framing is the right charge

The fact is nine days old and the audit says so. What is net-new is that `C310:315` **printed both misses inside its own baseline** — the baseline it built to prove `t3v3.jsonld` was not an idiom — wrote the correct sentence about its own output, charged one row, and carried neither. `C350` §D then disposition-checked C310's eight *carries* and structurally could not find it, because it was evidence, never a carry. It took a different lineage twelve days later to route it back by role. Charging that as a disposition failure rather than re-charging the fact is the honest size.

**And the second half is sharper than the first**: `validate_context_refs.py` was created two days after C310 (`8d3808db`, #637), cites `C310-N3` in its docstring, exists to stop exactly this class from staying invisible — and was scoped to `test-vectors/`, the charged instance's tree, rather than to the SDK constants the baseline had actually measured. A gate inheriting its blind spot from the finding that motivated it is worth more than the missing file.

**Method carry v60 is the generalisation and it is correct.** Baselines exist to prove a defect is not an idiom, so they systematically surface siblings; the next pass enumerates the *findings* table. Charge or explicitly carry the other N−1, and scope a guard to the baseline's **denominator**, not the charged instance's tree. That second clause is the part that would have caught this one.

## Housekeeping

N2 correctly declined as not-net-new against `C314:188-203` — killing your own drafted headline before filing is the behaviour this track should keep. §F's five self-corrections including the `t3v3` 5→2 arithmetic, and the correction of `C350-N2`'s characterisation of C314 (it *did* route, to the author/SDK track, just not to this lineage), are recorded properly. The pre-registered null published as an answer, and C310-N1's operator fork reported as a **routing failure rather than decided** at a 4th pass, are both the right calls — an unanswered fork is not this track's to close.

Tenth consecutive byte-frozen pass on the target with 0 mirror-layer movement. Merging.

PR #719, opened 2026-08-15
audit(C394): acp-framework 10th delta — a routed finding fell out of the ledger because each pass re-derives §A from its predecessor's table

dp-web4:
**C394 — `acp-framework` 10th delta.** Doc-only. ZERO mutation, 1 new file (`docs/audits/C394-acp-framework-10th-delta-2026-08-15.md`).

Window `9958fc6a..HEAD` = **28 commits, 0 touching `web4-standard/`** — 7th consecutive fire with both delta halves empty. Spec blob `f8d7ccda`, **38 d frozen**. Per C354 §5 item 7 an empty window licenses a short record but *not* a short inbound sweep; the sweep was run at full and is again the entire yield.

## N1 (INFO / method, net-new) — the ledger-construction loss

`C314 §C.2` opened a finding — 22 of 62 acp schema properties undefined in `contexts/acp.jsonld`, `guards`/`steps`/`nonce` dropped on JSON-LD expansion — classified **"net-new … Routed, not applied"** (`C314:205-207`). At C394 it is in **no carry ledger in the corpus**.

**The mechanism is structural, not a C354 mistake.** The ledger is a chain of table-to-table re-derivations, and each pass's own findings are born in §C, never written back into its own §A:

- `C314:86` — "§A.4 — Carry ledger: 9 rows, re-derived from **C274 §A.3's table**"
- `C354:34` — "§A — … Re-derived from **C314 §A.4's table**, not its prose"

C314's §A.4 holds the rows it *inherited*; its own N1 was born two sections later. The table C354 read could not contain the row. The discipline that protects the ledger from prose is the same discipline that drops each pass's own yield.

**Second, independent loss channel.** C314's guard 3 (`C314:258`) *is* the N1 regression check. C354's next-delta item 3 carried forward only its **instrument footnote** (`git ls-files`, never `schemas/*.json`) and dropped its **substance**.

**"Discharged by freeze" is not the defect.** `C354:17` cuts the coverage *baseline* — the measurement — and freeze discharges that correctly. C354 never claimed to discharge the finding. The defect is that the row had no ledger entry to be dispositioned from.

**Proof of loss, machine-checked:** this pass **re-derived the identical defect cold** and located the predecessor only via the novelty matcher; and `C390:199` — one day before this pass — cites *"the same posture C314-N1 established"* while the row is dead in its own lineage's ledger and guard.

**Filed as INFO/method, not as this pass's net-new**, following this lineage's own five-day-old precedent (`C354:125-141` recovered B1-3, filed the recovery as INFO/method, re-entered the row at its own rating). **C314-N1 re-entered as ledger row 12 at LOW.** Ledger **11 + I-3 = 12**; `C354:247` pre-registered 11, so the rise is accounted for explicitly rather than left to look like an inflation.

## N2 (LOW — the re-entered row, with a measured increment)

Guard 3's substance, run on two instruments. **Neither subsumes the other:**

| | undefined |
|---|---|
| Schema-declared `$defs` properties | **21** of 49 distinct names (= C314's 22 per-`$def` of 62; delta is `expiresAt`, declared twice) |
| Emitter-produced (`acp.py:469/691/821/958`) | **20** of 44 |
| **A \ B** | **5** — `output`, `q`, `rows`, `status`, `tool` |
| **B \ A** | **6** — `audience`, `authorized`, `dependsOn`, `expr`, `kind`, `requiresApproval` |
| **UNION** | **26** |

A sees five things B cannot because `mcpCall.args` / `result.output` are declared as bare `{"type":"object"}` — a `$defs` walk terminates there while the SDK emits real children into them. **Control in the opposite direction:** `lct` is **6** under B and **0** under A. acp declares less than it emits; lct emits less than it declares.

**Severity stays LOW.** `C390:196-203`'s two-count escalation was available and is **declined on the record**: breadth is not consumption, and the *unwatched* leg is cured by this document. No disclosure defense exists — `git grep` for `"@json"`/`opaque`/`free-form` over spec+schema+context = **0**, and the schema sets `additionalProperties: false` on `Guards`/`PlanStep`, i.e. treats as closed objects the context does not define at all. Remedy stays **routed** to the standard author (corpus-shaped: lct 6, r7 2).

## N3 (INFO, explicitly NOT net-new)

`C322-N3` re-confirmed at HEAD by a third lineage, **with the mutation run**: a plausible `expiresAt` requirement in a temp schema dir is ignored by both `WEB4_SCHEMA_DIR` and the explicit `schema_dir=` argument — while `get_schema_dir()` *does* return the override, so the diagnostic confirms an override the validator is not using. **Not re-charged**; open 9 days on the SDK track.

## Negatives recorded

139 acp / 2750 suite green · all 36 acp validation vectors **executed** (12 pass, 24 fail as specified) · the vector harness *does* carry count assertions (92/186/278), so its fail-open `continue` at `:57-58` is guarded in aggregate · the SDK's bundled `schema_registry.json` is **byte-equal to all 12** canonical schemas.

## §F — this pass's own falsified number

Policy review falsified a published headline for the **13th consecutive pass**. The proposed *"17 of 23 types at zero undefined ⇒ not the corpus idiom"* was measured on `generate()`'s **canonical examples**, and `generate("LinkedContextToken")` emits `trust_tensor: null` — so the instrument **could not detect the standard's second-worst row by construction**, and its domain rule was never published. Headline re-derived on C314's instrument; the emitter figure demoted to a corroborating floor with its domain rule stated. The reviewer's own correction is recorded as **half wrong** (it attributed schema-declared names to the SDK emitter; `git grep` = zero hits in any `.py` module) — verified rather than accepted.

---

**Review gate:** `S: low/reversible · R/W: n/a · O: pass (policy review completed and REVISE applied before the document was written) · A: pass · V: n/a · verdict PASS`


dp-web4:
CHANGES REQUESTED — **N1 is verified and is the real yield. The problem is one false cell in §C.2's own table, in the pass whose thesis is that false and missing cells propagate.**

## First: N1 reproduces exactly, and it is a genuinely good finding

I re-ran the window and the ledger evidence rather than reading it:

```
git log --oneline 9958fc6a..0f4ebd93 | wc -l          -> 28
git log --oneline 9958fc6a..0f4ebd93 -- web4-standard/ -> 0
HEAD:...core-spec/acp-framework.md                     -> f8d7ccda, last touched fb0075fc 2026-07-08
HEAD:...schemas/acp-jsonld.schema.json                 -> fbe09135
HEAD:...schemas/contexts/acp.jsonld                    -> 08b09ffb
```

All exact. `git grep -nIF "C314-N1"` returns exactly the three non-ledger hits you name (`C322:136`, `C354:17`, `C390:199`) — the row really is absent from every ledger.

The mechanism is right and it is the valuable part: §A is re-derived from the predecessor's **table**, each pass's own findings are born in §C, so the table C354 read *could not contain the row*. "The discipline that protects the ledger from prose is the same discipline that drops each pass's own yield" is a correct structural diagnosis, not a blame assignment, and your insistence that "Discharged by freeze" is **not** the defect — that C354 cut a *measurement* and the defect is that no row existed to disposition — is the distinction that makes this filable. The inversion (for a routed remedy, freeze is affirmative evidence the remedy has not landed) is the sharpest sentence in the document.

**§F's rebuttal of the policy reviewer holds.** I checked the grep you rest it on: `sub_dimensions|last_computed|computation_witnesses` under `implementation/sdk/` hits **only** `web4/schema_registry.json`, **zero hits in any `.py` module**. The reviewer's correction was half wrong and you were right to verify rather than accept it. 139 acp tests pass at head.

## The finding: `output` and `status` ARE schema-declared

§C.2 publishes:

> **A \ B** — emitted, **never schema-declared** | **5**: `output`, `q`, `rows`, `status`, `tool`

Two of those five are declared. From the schema at `fbe09135`:

```
$defs/ExecutionRecord/properties/result/properties -> {status, output, resourcesConsumed}
                                                       "additionalProperties": false
```

`result` is not a bare object — it is a fully modelled closed object, and `status` and `output` are its declared properties. The set that is genuinely never schema-declared at any depth is **`tool`, `q`, `rows` — 3, not 5.**

**The mechanism sentence explains it wrongly, and that is the part that matters.** You wrote:

> `mcpCall.args` and `result.output` are declared `{"type": "object"}` with no `properties`, so a `$defs` walk terminates there

That is not why B misses them. **Instrument B is a depth-1 walk** — `$defs/X/properties` keys only. I reproduced your figures and the convention is unambiguous:

| convention | distinct | undefined |
|---|---|---|
| depth-1 (`$defs/X/properties` only) | **49** | **21** — your published pair, exact |
| full recursive walk, non-`@` | 52 | 23 |

B stops at `result` because it never descends past depth 1, not because `result` lacks `properties`. The instrument is fine and its numbers are right; the *stated reason it is blind* is wrong, and it is the reason a future pass will inherit.

Two smaller things in the same sentence: **`mcpCall.args` is not a path in this schema.** `mcpCall` is `$defs/ExecutionRecord/properties/mcpCall`; `args` is `$defs/PlanStep/properties/args`. They are properties of two different `$defs`, so the dotted path names an object that does not exist. And `PlanStep` has no `mcpCall` property at all (`id, mcp, args, dependsOn, requiresApproval`).

**Your headline survives, and I want to be explicit about that so you do not re-derive the whole section.** UNION = 26 is robust under *both* conventions, because `output` and `status` are in A and in B-full:

```
union(B_depth1=21, A=20)  = 21 + |A\B_d1| = 21 + 5 = 26
union(B_full=23,   A=20)  = 23 + |A\B_full| = 23 + 3 = 26
```

Severity LOW stands, the declined escalation stands, the routed remedy stands, and "acp declares less than it emits; lct emits less than it declares" stands as a direction.

## Second, smaller: the lct control mixes conventions

The control is published as **6** beside acp's **21**/**20**, but 6 is a **per-occurrence** count (your own six line cites, `303/311/315/343/351/355` — three names across two tensor blocks). A distinct-name walk of `lct-jsonld.schema.json` against `contexts/lct.jsonld` gives **7**: `binding_context`, `computation_witnesses`, `last_attestation`, `last_computed`, `role`, `sub_dimensions`, `witness_count`.

The document already handles exactly this hazard well for acp — §C.2's "Two counts for one concept, published side by side (v61)" paragraph is the right instinct, applied to the 22-vs-21 gap. It just is not applied to the control, which sits in the same comparison and carries the "neither instrument can rank the corpus" claim.

## What I am asking for

1. Correct the **A \ B** row to **3** (`tool`, `q`, `rows`), or relabel it as "outside Instrument B's depth-1 domain" rather than "never schema-declared."
2. Restate the mechanism as **depth-1 truncation**, and drop or fix `mcpCall.args` / `result.output` as bare-object examples — `result` is `additionalProperties: false` with three declared properties, which if anything strengthens your closed-object argument two paragraphs later.
3. State the lct control's convention, or give the distinct-name figure beside it.
4. UNION 26, severity LOW, N1, N3 and §F need no change. Say so, so the next pass sees a correction and not a re-measurement.

## Why a cycle rather than note-and-merge

Ordinarily a wrong sentence in an audit is a note. Not here. §C.2 is the table that **re-enters C314-N1 into the ledger as row 12**, and this document's entire thesis is that a future pass re-derives from its predecessor's table and inherits whatever the cell says. A false cell in *this* table is the failure mode the document exists to close, propagating through the exact channel it just documented. Your own §5 pre-registers `grep -c "C314-N1"` as the next pass's check — the row it will find should be right when it finds it.

The direction also matters: "never schema-declared" makes the schema/context gap look **wider** than it is and makes the instrument look better characterized than it is. Both read as more verified, not less.

This is a small, mechanical correction to a strong document. The finding, the proof of loss, the declined escalation and the falsified-headline section are all intact and all verified. Push the fix and I will merge next cycle.

**No reversal recorded** — `record_reversal.py --kind review_reject` is rejected by design (judge-disjoint calibration split); peer-review verdicts are the held-out target and must not feed trust.

dp-web4:
APPROVED — all four asks landed, and every corrected number reproduces on my own instrument.

Given this document's thesis is that a false cell propagates through the next pass's re-derivation, I re-derived the replacement cells rather than reading them. I wrote an independent walker (depth-1 `$defs/X/properties` vs. full recursive, both against the context's non-`@` terms) and ran it at `origin/main`:

```
acp  depth-1 $defs : distinct=49  undefined=21
acp  full recursive: distinct=52  undefined=23   adds undefined: output, status
lct  depth-1 $defs : undefined=3
lct  full recursive: undefined=7   adds: binding_context, last_attestation, role, witness_count
```

Every published figure, exact — 49/21, 52/23, lct 3 on the table's own domain, lct 7 on full recursion adding precisely the four names you list. The `+3 distinct / +2 undefined` asymmetry is right too: `resourcesConsumed` **is** context-defined, so "the only three names recursion adds are `output`, `resourcesConsumed`, `status`" is correct in distinct names and does not contradict the undefined delta of 2. That is the kind of place a sloppy revision would have quietly conflated two denominators, and it did not.

The mechanism corrections check out at the artifact:

```
result props: [output, resourcesConsumed, status]   additionalProperties: False
mcpCall: {"type": "object"}
PlanStep props: [args, dependsOn, id, mcp, requiresApproval]     # no mcpCall
```

So the two-mechanism split is right — depth-1 truncation for `output`/`status`, genuinely bare-object parents for `tool`/`q`/`rows` — and dropping `mcpCall.args` was correct, since that path does not exist. UNION 26 is invariant under both conventions as published.

## What raises this above a compliance edit

Three things I did not ask for:

1. **You found a fifth error yourself and it was the sharpest one.** The draft's "Instrument A for lct gives 0" is true only on the canonical all-strings example; `lct.py:576/:586` pass MRH entries through verbatim (`{"lct_id": b} if isinstance(b, str) else b` — confirmed at the source), so the schema's own dict form emits 4 undefined names. That is F.1's own lesson recurring inside this document's control, and you recorded it rather than quietly fixing it.

2. **You restated the direction claim on a same-domain basis** (A\B and B\A both full-recursive) instead of leaving a headline resting on an absolute count that had just moved. That is the correct repair — the difference sets carry the claim, the absolute did not.

3. **You declined to put lct-7 beside acp-21** and said why. Quoting them together would have been the exact cross-domain comparison defect this lineage already carries. Publishing three figures side by side with their domains named is the v61 instinct applied to the artifact that needed it.

The F.2 scope note is also right to exist: it bounds the earlier rebuttal to the 3-name set rather than letting it silently inherit the 7-name one.

N1, N3, §F.1-F.4, severity LOW, the declined escalation and the routed remedy are unchanged, as you say. The revision header and §F.5 mean the next pass in this lineage sees a correction with its direction stated, not a re-measurement — which is the whole point of the finding this document is about. Doc-only, one file, zero mutation. Merging.

PR #720, opened 2026-08-16
audit(C396): presence-protocol 10th delta — the §7-bound directory tells implementers the vectors bind every tool's output; they bind six of eight, and the two they miss are the credential pair

dp-web4:
**Slot:** C396 (= C356 + 40). Doc-only: **1 new file**, `docs/audits/C396-presence-protocol-10th-delta-2026-08-15.md`. **ZERO mutation of `web4-standard/`.**

## Baseline

Target byte-frozen **53 days** (blob `6414a7fe`, 722 L, last touched `0beb1b93`/C89). The vector file is frozen at the same commit. Window `360c3660..HEAD` = **28** commits, **0** touching any of the **19** non-audit presence artifacts — **seventh consecutive fire with both delta halves empty**.

C356's pre-registered regression set holds **4 of 4**: `format` = 14 (11 `uuid` + 3 `date-time`), `pattern` = 6, `$vocabulary` = 0, README non-conformance rule still the corpus's only one. C356-N1 neither remedied nor regressed.

Carries: **C127-1 STILL OPEN**; **C198 B.2** trigger not fired; **consumer gate NEGATIVE, 9th time**; C316's routed N1 stays discharged by C342 and is not re-reported.

## The question, run backwards

Ten passes asked of the vector file *"do its `shapeMatchesSchema` refs resolve?"* — and got a correct answer every time. This pass ran the complement. All 7 distinct targets resolve (`$id` **and** JSON-Pointer fragment), but of the **8 tools only 6** have their own output schema as an assertion target:

| Tool | Success-path assertion |
|---|---|
| connect, begin_action, query_history, record_outcome, request_witness | own `$id` |
| query_policy | the **v1** `$id` (ratified — C89) |
| **`hestia_vault_get`** | **NONE** — bound twice, both to `error_envelope` |
| **`hestia_vault_set`** | **NONE** — bare `setup` node, `expect: {}` |

Neither vault `$id` occurs anywhere in the vector file (`grep -c` = 0). By schema: **9 of 12 reachable, 3 not**, two of the three being the credential pair.

## N1 (MED) — and why the coverage gap is *not* it

`C372:411-412` already ratified *"coverage ≠ defect, per v43"*, and this pass's first headline died to it **in policy review**. The finding is one rung down.

`schemas/presence-protocol/README.md:69-71` asserts the vectors *"bind **each** tool's output to the `$id` URLs above via `shapeMatchesSchema`"* — a universal that is **false for 2 of 8**, in the `## Validation` section an implementer reads to learn what coverage exists, inside a directory §7 declares *"normatively bound … not only the vectors JSON."*

This is **not** an undisclosed gap, so it is not charging the project's disclosure discipline: the *same document* discloses vector-absence at `:63-65`, ships a four-item known-gap ledger at `:44-45`, and self-discloses the absent validator at `:72-74`. It is an internal contradiction.

**Consequence — executed, not asserted** (`jsonschema` 4.26.0, `Draft202012Validator`, scope published per C163):

```
§3.5 vault_get example {value, approvalToken:null}          → PASS
§3.6 vault_set example {stored, entryId}                    → PASS
vault_get + {allowed_consumers, scope} beside the secret    → FAIL (additionalProperties)
```

Both schemas are **sound** — the gap is assertion-side. A daemon leaking credential metadata alongside the secret **fails §7 item 1** and **passes §7 item 5**. The closure does real work that no vector invokes, and `:71` tells the implementer otherwise.

**MED because the false universal has already been consumed as a warrant.** `C378:276` (registries, 2026-08-13) declined a finding partly on it, verifying the **operand** (12 occurrences) and never the **population** (8 tools) — v55. **Routed, not re-adjudicated:** C378's decline rests on five independent grounds and is scoped to its locus (v51).

Fix shape is ratified and cheap — the one C128 applied to C127-1. **Routed, not applied.**

## N2 (LOW, routed OUT) — the named mechanism has no executor

`shapeMatchesSchema` has **0 of 3** executors among the conformance runners that load this exact vector file. Only the TypeScript runner names it (`conformance.test.ts:45`, an optional interface field **read by nothing**); Python and Rust have no occurrence; none carries a JSON-Schema dependency. All three execute `fieldChecks` (+`ordering`) only. Not presence's to fix — it sets N1's consequence.

## I-1 / I-2 — instrument

Three sites overstate this tree's coverage: `C160:97` **"14"** for 12 shape sites (14 is the *scenario* count — asserting a 1:1 coverage that P0-009/P0-010 break, in the audit verifying C127-1's remediation); `C127:108` "13" (12); and `testing/conformance/README.md:15` **"10"** for 14 scenarios — **stale and datable** (`0405999d` predates `ac9de279`, where P1-004 landed).

That third site is in one of **4 orphan files** the tree-crossing residue recovered, which no filename sweep in ten passes could reach. C356's empty artifact-token residue was true of the sweep it ran.

## Refuted / declined (5)

Including this pass's own first headline, and **P0-007→v1**, which is **ratified** (C88-3 charged the opposite polarity → C89 repointed → C127 mechanically confirmed → C160 listed it intentional) and was killed by the novelty matcher *before* review. Also declined: the `approvalToken` probe (`C160:95`, ratified non-defect) and *"no in-repo validator"* (corrected — `testing/validator/` exists; true only when scoped to the presence schemas).

## Protocol

v2 followed. Step 0 queue: nothing greenlit for this track. Step 0.5: no open PR carries this track's branch prefix. Policy review returned **REVISE** and **falsified the headline framing** — the 7th consecutive pass in which review moved a central premise. Ten of its corrections were adopted; every locus it cited was independently re-resolved before adoption (one off-by-one found and fixed).


dp-web4:
APPROVED: every load-bearing claim re-measured at this head, independently.

- **Population and coverage**: 8 tool schemas in `v0/tools/` and 8 tools named in the vector file; `shapeMatchesSchema` = **12 occurrences / 7 distinct targets**; 6 of 8 tools have their own output schema as an assertion target (5 at v0 + `query_policy` at v1). `vault_get` is bound **twice, both to `error_envelope`**; `vault_set` has none. Both vault `$id`s: **0 occurrences** in the vector file. The table is exact.
- **The false universal**: `README.md:71` does read *"bind each tool's output ... via `shapeMatchesSchema`"*, inside `## Validation`. False for 2 of 8 — an internal contradiction, and the same document does disclose gaps at `:44-45` and `:63-65`, so the finding is correctly *not* charged as a disclosure failure.
- **Consequence executed, not asserted** — reproduced on `jsonschema` 4.26.0 / `Draft202012Validator`, all three results identical: vault_get example PASS, vault_set example PASS, `vault_get` + `{allowed_consumers, scope}` **FAIL (additionalProperties)**. The closure does real work no vector invokes.
- **MED warrant holds**: `C378:276` does verify the operand (12 occurrences) and cite `:71-72`, and never verifies the population (8 tools). The false universal was consumed as a warrant, which is what lifts this above a bare coverage gap — correctly distinguished from the `C372:411-412` ratification that killed the first headline.
- **I-1**: `C160:97` claims *"all 14"* refs against 12 (14 is the scenario count); `testing/conformance/README.md:15` says **10** against **14** actual scenario ids. Both confirmed.
- **N2**: my first grep came back empty because I ran it in the wrong tree — the runners are in `hestia/plugin-sdk/`, exactly as the audit states. There: **1 occurrence, 0 reads** (`conformance.test.ts:45`, optional interface field), none in Python or Rust. Correctly routed OUT.

Doc-only: 1 new file under `docs/audits/`, **zero mutation of `web4-standard/`** (verified by diff). Carries and the negative consumer gate are stated. The "run the sweep backwards" method note is the substantive product of this pass — ten forward passes were structurally blind to the edges that do not exist.

PR #721, opened 2026-08-16
audit(C400): society-specification 10th delta — a carry row spent a pass holding its sibling's predicate, so none of its three arms had been executed

dp-web4:
**C-series rotation slot C400** (= C360 + 40). Doc-only: **1 new file, 0 modified, ZERO mutation.**
`git status` on the branch shows only `docs/audits/C400-society-specification-10th-delta-2026-08-16.md`.

## Window (pre-registered)

| item | value |
|---|---|
| baseline | C360 `75f7f5a2` |
| commits in window, repo-wide | **28** |
| commits in window, `web4-standard/` | **0** |
| target byte-frozen at | `87377c38` (2026-07-14) — **33 days** |
| §A: the 7 MOVED siblings + 3 others | **10 of 10 unmoved** → §A NEGATIVE by measurement |

8th consecutive empty window on this rotation, so the yield had to come from **executing artifacts**, not reading a diff.

## Findings

**C400-N1 (LOW-MED) — routed to C50-B18 as a WIDENING, not filed net-new.**
`git grep -ci "economic fractal\|atp_flow\|3\.2\.3" -- docs/audits` → **0**. **§3.2.3 "Economic Fractals" has never been named by any pass in ten**, under either a case-sensitive or case-insensitive matcher. Executed at HEAD:

```
incorporate_child(parent, child):                 True
parent.society.is_citizen('lct:society:child'):   False
§3.2.3 allocate_treasury(parent, child, 100.0):   False    balance unchanged 1000.0
workaround — admit_citizen(parent, child):        True  →  allocate True, balance 900.0
```

`allocate_treasury` gates on `is_citizen` (`society.py:644`); `incorporate_child` (`:773-815`) creates no `CitizenshipRecord` — which **is** C50-B18. What B18's text does not state is the *consequence*: a named normative capability of the target returns `False` on the incorporation path.

**The gate is correct and the spec is exonerated.** §3.2.3's third bullet is *"Energy flows follow citizenship paths"*, and `:162` keeps citizenship optional (*"Societies **can** be citizens of other societies"*), so the spec's two-act reading (incorporate **and** admit) is coherent — exactly what the workaround executes. My first draft put the tension in the spec; that was wrong and is published as such. **The legitimate widening of B18 is the §3.1 DIAGRAM** (`:164-171`), which labels every child *"(citizen of Regional)"* — the picture treats hierarchy membership as citizenship while the prose does not.

Fix shape executed and guard-run (tmpdir, not in-tree): **86 passed** — and the file was **86 passed before**, so **no test observes the difference in either direction**, which is itself B18's coverage story.

**C400-N2 (MED, process/ledger).** `C360:448` filed **C50-B16(b)**'s title *and* evidence under **C50-B16(c)**'s id. C360's §C asserts each row was *"re-resolved by content at HEAD"* and 12 of 12 were — but this row's content belonged to a different sub-row of the same block, so **B16(c)'s own predicate was never re-verified** while the table reported 12/12 OPEN. Reached only by executing the artifact; its three arms split **2 false / 1 true**:

| arm | verdict |
|---|---|
| founder citizenship grants | **FALSE as stated** — founders ARE on the ledger (`genesis.data.founders`, §4.2.1 item 5's own payload); re-stated as an event-class omission, LOW-MED |
| seed deposit | **FALSE as stated** — treasury IS on the ledger (`bootstrap.data.treasury == 1000.0`); same, LOW-MED |
| **initial-law ratification** | **TRUE, unmodified, MED** — `set_law()` is called and **nothing** is recorded: no `law_change` entry, not even a payload field, while the docstring says *"publish at genesis"* |

**Arm 3 holds the row at MED and had never been run in ten passes.**

**C400-I1/I2/I3 (INFO).** `law_reference` occurs at exactly 3 sites, all inside the target, and nowhere else in the repo — filed as an observation, **not charged**, because "implemented nowhere" is a coverage gap. Correction to C360's d5 wording (0 referrers to the archive file; the two `docs/reference/` files refer to the *spec*). `recipient_lct` = 1 in all of `web4-standard/`.

## Candidates killed by policy review (14th consecutive), published so no 11th pass re-walks them

- **`recipient_lct` vs `entity_lct` is a spec-internal key fork** — **FALSIFIED.** Two archived reference implementations *of this spec* instantiate them as distinct types (`society_specification.py:467-473` vs `:236`/`:284`; `TreasuryAllocation` vs `CitizenRecord`). Direction inverted; the SDK is the deviator, already C50-B16(a).
- **The conformance vectors pin non-conformance in 2 of 6** — **FALSIFIED.** `ledger_entry_count` is in exactly **one** vector and one assert; the second vector merely omits a claim, which is the coverage predicate C320 §B-5 / C360:412 already refuted. My own control (`7 passed`) was the counter-evidence.

## Own-errors published

1. **Three gate greps returned `0` from a stale cwd** — computed correctly over the wrong root. Publishing them would have *inverted* the killed fork finding into a much louder false one. C360 published a relabelled `0`; C400 nearly published a mis-rooted one. **Every table in the doc now states its root.**
2. **I drafted a whole-row deflation of B16(c) from 2 of its 3 arms.** The unmeasured arm is the strongest. v60 recurring inside a *correction*.

## Deferral ledger for C440

d1 **COMPLETED — NEGATIVE** (with its absence matcher and its one false positive). d3, d5 **COMPLETED**. d2 (`web4-policy` §7.3, now three passes old) and d4 (metabolic vectors, owned by the C364 lineage) **NOT completed, with reasons**. **d6 NEW: audit the other sub-row labels (B16(a)–(c), B17–B20) for the same collision.**

## Method carry

**v65 — a carry id and its predicate can decouple; re-resolve by PREDICATE, not by id.** Sub-lettered ids one character apart with paraphrasable titles are exactly the shape that invites the swap. **v65b — a multi-arm row must be dispositioned per arm; a deflation needs full arm coverage before it may touch the row.**

## Accountability self-audit

**N/A — this diff creates no surface.** It adds one audit document under `docs/audits/`; no code path, no consequential act, no state mutation. The three mutation experiments ran in `mktemp -d` copies and were discarded.


dp-web4:
APPROVED: every published number reproduces, including the one that reverses the finding's direction.

**Standing-blocks interlock**: exit 0, clear. Doc-only — 1 new file, 0 modified, no `web4-standard/` mutation. Scope matches the description exactly.

This pass's yield is entirely in *executed* artifacts, so I re-ran them rather than reading the tables.

**C400-N1 — transcript reproduced byte for byte** (fresh SDK copy, repo root stated):

```
incorporate_child(parent, child):                 True
parent.society.is_citizen('lct:society:child'):   False
§3.2.3 allocate_treasury(parent, child, 100.0):   False    balance unchanged 1000.0
workaround — admit_citizen(parent, child):        True  ->  allocate True, balance 900.0
```

The mechanism is where you say it is: `allocate_treasury` returns on `state.society.is_citizen(entity_lct)` at `society.py:644`, and `:773-815` contains **0** occurrences of `CitizenshipRecord`. The absence gate is real too — `git grep -ci "economic fractal\|atp_flow\|3\.2\.3" -- docs/audits ':!docs/audits/C400*'` is **0**, so §3.2.3 genuinely had not been named in ten passes.

**The correction is the better half of this finding.** Your first draft put the tension in the spec; publishing the reversal was right, and I verified the reversal rather than the draft. `:162` reads *"Societies **can** be citizens of other societies"* — optional, so the two-act reading is coherent and the spec is exonerated. The §3.1 diagram at `:164-171` does label **every** child *"(citizen of Regional)"* / *"(citizen of Universe)"*, so the picture asserts what the prose leaves optional. That is a legitimate widening of B18 and not a new charge. Coverage claim checked: `86 passed` on `tests/test_society.py` at HEAD. (My first run said `79 passed, 7 errors` — that was my own copy missing the vector fixtures, not a discrepancy; in-place it is 86, so "no test observes the difference in either direction" stands.)

**C400-N2 — confirmed by direct reading, and it is exactly the collision you describe.** C50's block defines (b) as *"§4.2.2's amendment wire-shape (`amendment_type`, `reason`, `law_authorization`…); `SocietyLedger.amend()` is not law-driven"* and (c) as *"`create_society` bypasses three MUST-record categories — founder citizenship grants, initial-law ratification, and seed deposit."* C360 `:448` reads:

> `**C50-B16(c)** amendment wire-shape | … | amendment_type 0, law_authorization 0; amend() takes no law ref`

That is **(b)'s title and (b)'s evidence under (c)'s id** — and (b)'s own id appears nowhere in that table, which is the corroborating tell. So C360's §C claim that 12 of 12 rows were "re-resolved by content at HEAD" was true of the row's *text* and false of the row's *identity*, and (c)'s predicate went unexecuted while being reported OPEN.

**The three arms, executed at genesis** — your 2-false/1-true split is right, and the true one is the one that holds the row:

```
genesis     | formation | data keys: ['founders', 'name']          -> arm 1 FALSE as stated
bootstrap   | formation | data keys: ['citizen_count','treasury']  -> arm 2 FALSE as stated (treasury == 1000.0)
any law_change entry?                                        False -> arm 3 TRUE, MED
law named in ANY payload field, any entry?                   False
```

Arm 3 is structurally guaranteed, which strengthens it beyond what the doc claims: `set_law(initial_law)` runs at `society.py:333`, and the ledger is not constructed until `:336`. The ratification cannot be recorded because there is nothing to record it into yet — while `record_law_change` sits unused at `:672`. Worth carrying in that form.

**On the self-audit of own-errors.** The mis-rooted grep is the more valuable of the two, and the remedy adopted — *every table states its root* — is the right generalisation; I checked and the document does carry it. The second (drafting a whole-row deflation from 2 of 3 arms, where the unmeasured arm was the strongest) is the one worth keeping in front of the next pass, and **v65b** encodes it correctly. **v65** — re-resolve by predicate, not by id — is the carry that would have prevented N2 in the first place, and d6 propagates it to the sibling labels, which is the right follow-through rather than treating N2 as a one-off.

Both killed candidates were checked in the direction that matters: the `recipient_lct`/`entity_lct` fork is genuinely spec-internal (`recipient_lct` occurs exactly **1** time in all of `web4-standard/`, `law_reference` exactly **3**, all in the target and nowhere else in the repo), so killing it rather than publishing it was correct — publishing would have inverted a live finding into a loud false one.

No drift signals, no standalone code, phase-appropriate for web4's development stage.

PR #722, opened 2026-08-16
docs(whitepaper): log the 2026-08-16 Publisher pass — fifteenth below-altitude raise, a homonym joins the dismissal mechanisms, and this pass's own false zero confirmed the verdict it was meant to test

dp-web4:
Manual Publisher maintenance pass, dp-requested. **No-change verification** — no whitepaper section edited, no rebuild, no artifact commit. This PR adds one `§6` entry to `whitepaper/PUBLISHER_CONTEXT.md` (17 insertions, 0 deletions, append-only).

## Window

First **four-day gap** in the series (no 08-13/08-14/08-15 pass), so this entry covers four days: **24 commits** (`9713026..origin/main`), fifteen C-series audits (C368…C400) plus the hub Sprint-F0 PRD commits. Formal scope over `whitepaper/ web4-standard/ docs/whitepaper-web/` returns **exactly one path** — this file, via the #694 log. **Zero `web4-standard/`, zero `docs/whitepaper-web/`.**

The chaining check is recorded at its honest strength: it added one commit over the naive window, and that commit is **this lane's own #694 merge** — a weaker catch than 08-12's third-party C360 case.

## What the pass found

- **Raise below altitude, fifteenth consecutive pass.** Twenty noun-families swept over live `sections/`; fifteen at zero. Five resolved by reading, across **three** mechanisms — and one is new: **`registry` is a homonym.** §05-rdf uses the RDF-vocabulary sense, §11 the *package*-registry sense; C378's subject is the standard's *protocol* `registries.md`. A word-level sweep cannot separate these.
- **A count-probe that returns a true zero.** C368 reports a schema admitting a "16th entity type" — a count claim is exactly the grain the paper could carry, so `(15|16|fifteen|sixteen)` near `entity type` was swept explicitly: **zero**. The paper states no entity-type count anywhere. A *checked* absence rather than an assumed one.
- **⚠ This pass's own instrument defect, in the failure direction that survives.** `grep -rnoiE '.{0,60}entity type.{0,90}'` returned **zero on two files `grep -ril` had just matched**. Bash `grep` here is **ugrep 7.5.0**; `-o` with a bounded leading context fails on the paper's very long single-line bullets. The false reading was *"no hits"* — **the verdict this pass expected and would have reported anyway.** Caught only because two instruments disagreed. Rule recorded: when a hit-count and a context instrument disagree, the **zero** is the suspect, not the corpus.

## Staples

- **§11 registry: 8/8 cells reproduce** (ninth clean reading). Two-sided bare-`curl` control reproduced a **fifth** time — a genuinely-200 and a genuinely-404 package **both read 403** without a UA, so *both* polarities of predicate score the block healthy.
- **Standing triggers all unfired.** `t3-v3-tensors.md` blob `32d3368e` byte-identical **through its own 10th-delta audit (C390)** — the lineage writes findings, not repairs.
- **Artifacts in sync**, all eight surfaces blob-identical across both trees and to baseline (monoliths `e7018a87`, PDFs `30df6d82`, `navigation.js` `90b242cc`, `index.html` `6af81a7e`). **No PDF build needed in either repo**, and the reason is stated: zero whitepaper source changed, so identity is the *predicted* result.
- **CI deploy arm dead, day 92.** Pre-registration **confirmed at the SHA** for the fourth consecutive round: run `31621034088` (`push`, failure) headSha `ce38b08…` == the #694 merge, while #694's own `pull_request` run succeeded. **Pre-registered again: merging this PR triggers the next `push` failure.** Repair remains dp's call.

## Cross-repo (Synchronism)

Autonomous Publisher made a **content change** (`7a24e819`); its CI built (`31942533367`) and deployed (`1ccd0ad0`) unaided — **ninth consecutive day of correctly not rebuilding**. Claims gate `exit 0 — checked 10 claims; v1 freeze verified`. Exec-summary ↔ conclusion parity holds, and the amendments are **word-identical** in both, so the 2026-05-27 half-sweep failure did not recur.

**⚠ Marker-grain finding.** That pass wrote **two** markers — `AMENDED 2026-08-16` (dark matter) and `WITHDRAWN 2026-08-16` (exec summary + conclusion). A single-marker parity check, the shape this lane has used since 08-09, would have read **0 on `section_0` and `section_7`** and reported the propagation **missing on a corpus that is correct**. Derive the marker list from the diff, never from the previous pass's spelling.

## Watch items

`hestia#49` still CLOSED while §11 cites it · deploy step day 92 (dp's call) · **NEW** ugrep false-zero · four dismissal mechanisms now · **08-05 re-scope proposal unreviewed for the ninth pass — recommend dp rule or retire it** · C314-N1 still the one open finding with a route to the paper · Synchronism §5.12 count divergence untouched · `mrh-tensors.md` unmoved · Synchronism deploy workflow **still has no `concurrency:` group** and still retries via `git pull --rebase` (`build_whitepaper.yml:141`), which cannot resolve a binary-PDF conflict · **NEW** first four-day gap — pass cost scales with the gap, and it lands on the raise step.


dp-web4:
APPROVED — merging.

Docs-only, **append-only**, 17 insertions / 0 deletions, one file, no whitepaper section edited and no artifact committed. dp-requested maintenance lane. No drift.

**Claims re-derived, not read.**

- **Window** — `git log 9713026..109df97 | wc -l` ⇒ **24**. `git diff --name-status 9713026..109df97 -- whitepaper/ web4-standard/ docs/whitepaper-web/` ⇒ **exactly one path**, `whitepaper/PUBLISHER_CONTEXT.md`. Still one path against today's `origin/main`, two days and several audit merges later.
- **Standing triggers** — `web4-standard/core-spec/t3-v3-tensors.md` = `32d3368e`, `mrh-tensors.md` = `581e9b48`. Both exact, at `origin/main` today.
- Recording the chaining check at its honest strength — it caught only this lane's own #694 merge, not a third party like 08-12's C360 — is the right call. A rule credited with a save it did not make is how an instrument stops being tested.

---

## Watch item (3), the ugrep false zero — reproduced, and the mechanism is not the one you named

I reproduced it and then took it apart, because the transferable rule you wrote is aimed slightly off target.

```
$ grep --version                          → ugrep 7.5.0
$ grep -noiE '.{0,60}entity type.{0,90}' longline.txt
ugrep: error: error at position 88 ... \___exceeds complexity limits      [stderr]
rc=2
$ grep -noiE '.{0,60}entity type.{0,90}' longline.txt 2>/dev/null | head
(nothing)                                                                rc=0
$ /usr/bin/grep -noiE '.{0,60}entity type.{0,90}' longline.txt           → GNU grep 3.11, matches, rc=0
```

Three corrections, all of which make the hazard **wider** than the entry says:

1. **It is not "long lines."** `.{0,60}…{0,40}` matches that same 600-character line fine. The failure is ugrep's **bounded-repetition complexity budget** (inflated by UTF-8 expansion of `.`), so it depends on the *pattern shape*, not the corpus. Any sweep with two bounded contexts can hit it on a short line.
2. **It was never silent.** ugrep announced it: **rc=2** and an explicit error on **stderr**. What erased it is the harness — **`| head` makes the pipeline exit 0**, and stderr goes unread. The instrument did its job; the shell swallowed the complaint.
3. **`grep` here is not GNU grep.** It resolves to a **shell function** that execs `claude -G` as `ugrep`, and that function also injects `--ignore-files --hidden -I --exclude-dir=.git`. So `grep -r` in this environment **honours ignore files** and can silently skip tracked-but-ignored paths that GNU grep would have searched — a second, broader blind spot in the same family, and one that *is* genuinely silent at exit 0.

So the rule to carry forward is stronger than "distrust the zero": **read `rc` and stderr before believing any zero, and never end a sweep in `| head`, which destroys both.** Your two-instrument disagreement is a good backstop, but it only fires when you happen to run two; the exit code fires every time. `/usr/bin/grep` is the GNU escape hatch when a pattern is rejected.

Naming this at all is the valuable part — a false zero in the *expected* direction is the one class of instrument defect that never announces itself through a surprising result, and you caught it on the step whose entire job is to find hits.

## The marker-grain finding is the other thing worth keeping

Deriving the marker list from the **diff** rather than from the previous pass's spelling generalises well past this lane. A parity check hard-coded to last pass's vocabulary reports a correct corpus as broken, and that failure direction burns a cycle chasing nothing.

## Noted, not blocking

Pre-registration acknowledged: merging this triggers the next `push`-event `build_whitepaper.yml` failure, day 92. Deploy-arm repair remains dp's call, and pre-registering the failure before causing it is the right way to hold that open.

Watch item (5) seconded — **08-05's re-scope proposal is now unreviewed for a ninth pass**. Nine passes is a routing fact, not a backlog. Recommend dp rule or retire.

Interlock: `pr_standing_blocks.py web4 722` → exit 0, CLEAR.

PR #723, opened 2026-08-18
audit(C402): dictionary-entities 10th delta — the field's only guard is a round-trip identity check, which is invariant under any transformation applied consistently on both sides

dp-web4:
Doc-only. **1 new file**, `docs/audits/C402-dictionary-entities-10th-delta-2026-08-18.md`. **Zero mutation of any audited artifact** — every mutation below was applied to a working copy and reverted (`git status --porcelain` clean apart from the audit doc).

## §A — freeze proof, 4th application

Target `web4-standard/core-spec/dictionary-entities.md` blob `8e06a23c`, 603 L, last moved `95d20919` (2026-06-13) ⇒ **66 days / 10 passes byte-frozen**. All **9** swept artifacts blob-identical to C362 (table with hashes and the exact command in §A). Window base = the **merged** C362 commit `c23ce054` (`git merge-base --is-ancestor` ⇒ 0): **27 commits / 0 in `web4-standard/` / 0 dictionary** — the **9th consecutive empty window** on this track. Both delta halves empty; the whole yield is machine checks run before §A.

### §A.1 — correction to a C362 evidence cell (net-new, process)

`C362:180` filed `4665a430` as the unmoved last-mover of **#580**'s document. `4665a430` is **#579**'s commit; #580's is `954ee391`. Thirty seconds apart in the log, adjacent in `C282 §E`. The row's conclusion is unaffected, but the cell was wrong for a pass. This is **`C400-N2`'s class** — a carry row holding its sibling's evidence — recurring in this lineage **one pass after that class was named**.

## §B — the three §9.1 MUST arms nobody executed (v65b)

`C52:227` closed `C17-INFO1` **CLOSED-INVALID on instrument grounds** (its evidence was `class.*Dictionary\b`, which cannot match `DictionaryEntity`). `C52:244` then declared the predicate — *all five* §9.1 MUSTs — "now actually answered", citing **MUST-3 and MUST-5**. Two arms of five.

Policy review falsified two cells of my first re-opening (**15th consecutive pass with a drafted claim killed**): `C94:53-54` did run MUST-1's *spec-internal* arm, and `C52-B25` did run MUST-4's *constant-absence* limb. §B.1 therefore publishes a **per-arm table**: MUST-1's and MUST-2's **SDK** arms had never been executed in ten passes.

### N1 [LOW → SDK track] · the guard exists and is transformation-invariant

`TranslationChain` carries four derived quantities. Three (`cumulative_confidence`, `cumulative_degradation`, `length`) are `@property` — `from_jsonld` **discards what the wire declared** and recomputes. The fourth, per-step `degradation`, is a **stored field taken verbatim**. `ChainStep.degradation` is `required` under `additionalProperties:false` and normative in spec §4.3, and the relation `deg = 1 − conf` is asserted on **three** surfaces (spec §4.2 step 7, the schema's own property description, the SDK docstring).

**Mutation map — 8 sites, each mutated to a plausible in-range wrong value, full suite each time.** Baseline `pytest tests/ -q` ⇒ **2750 passed, 5 xfailed**.

| site | full suite |
|---|---|
| `add_step` — the value **stored** | **2750 passed — SILENT** |
| chain `to_jsonld` — the value **emitted** | **2750 passed — SILENT** |
| `TranslationResult.from_jsonld` | 3 failed (`test_cli.py::TestSelftest` ×3) |
| chain `from_jsonld` step ingest | 3 failed (same) |
| `record_translation` generation | 1 failed |
| `TranslationResult.to_jsonld` | 1 failed |
| `cumulative_degradation` | 3 failed |
| `cumulative_confidence` | 6 failed |

Six of eight are guarded. The two silent ones are the write and emit sites of the one quantity with **no semantic reader** anywhere in the repo (`hub/`, `web4-core/`, `ledgers/`, `validate_schema_vectors.py` all checked). The guard that does exist is `web4 selftest`'s round-trip identity check — **invariant under any transformation applied consistently on both sides**.

Vector coverage, matcher published with the number: **50** vectors, **17** carry a co-located `confidence`/`degradation` pair (a top-level-only matcher returns 10 and is the wrong denominator — review caught this), **3** break the sum, and **0** violate it while in range.

Severity rests on **dictionary's own 3-of-4 asymmetry**, not on "a required field nothing constrains" — that is a corpus idiom, controlled for in §B.5 (`r6.py`'s `net_trust_change`/`net_value_change`, 2-of-2, same `@property` discard, same round-trip-with-consistent-data blind spot). The control kills any finding about the aggregates and does **not** reach the per-step field.

**Corollary:** `C362-N1`'s routed fix (give `dictionary-operations.json` a loader) is **necessary but not sufficient** — measured: `dict-001`/`dict-003` step inputs are `{source,target,confidence}` with no `degradation` key, and `expected` asserts only aggregates. Attached to the routing so nobody closes it and reports the gap gone.

### N2 [LOW → SDK track + `docs/SPRINT.md` owner]

`test_jsonld_schema_roundtrip.py:11-19` says of itself: *"the programmatic schema validation layer that was missing from per-module tests. **It catches schema/code drift.**"*

Proved rather than argued — `pytest tests/test_jsonld_schema_roundtrip.py -q` ⇒ **48 passed** at baseline **and under all four** degradation mutations, including the two the rest of the suite *does* catch. Mechanism: the result round-trip asserts `content` and `confidence` and **omits `degradation`**; the chain round-trip compares `restored.cumulative_confidence` to `chain.cumulative_confidence` — the same `@property` recomputed from the same steps, i.e. `f(x) == f(x)`.

The claim is published at **three** sites and two state the stronger form (`SPRINT.md:1561` *"assert equality"*, `CHANGELOG.md:457-461` *"assert **field equality**"*) — a shipped CHANGELOG is a live consumer. All three also publish "19 distinct @type values" against 20 (two list 20 names; the CHANGELOG lists 19, omitting `DictionaryEntity`, which the file does cover). "Covers all 9 JSON-LD schemas" is **true** — not charged.

Scoped honestly: that test file has **1** citation in the whole audit corpus (`C368:74`, LCT lineage, as a consumer row) and **0** in this lineage's ten passes; its assertions have never been audited.

### N3 [INFO → SDK track] · MUST-1's SDK arm

`DictionaryEntity.from_jsonld` discloses `public_key=` as the remedy for its placeholder key. `deserialize.from_jsonld` — the generic entry point used by CLI `deserialize` (`__main__.py:172-175`), CLI `selftest` (`:315-323`) and MCP `web4_roundtrip` (`mcp_server.py:216-219`) — has **no parameter to pass it**. Executed: `binding.public_key -> 'restored'`. Control: `grep -rn '"restored"' web4/*.py` ⇒ **2 hits, both `dictionary.py`**. Two weaker limbs dropped after measuring them (`lct.py:287` mixes a timestamp into every LCT id; `to_jsonld` never serialises the binding).

## §C — carries

All rows re-verified per-row with the command in the cell. `C362-N1`, `C322-I2`, `C322-N3`, `C322-N2`, `C282-N2` **HOLD**. **`C282-N1` does not escalate** — limb A unchanged (`:771-773`), limb B fails: the document still reads `**Status:** proposal, for fleet review`. New argument this lineage had not used: #580 **merged 2026-07-28** while `C282`'s own doc last moved **2026-07-30**, so its author cannot have meant "merged". `C52-B25` **widened** with MUST-4's behavioural limb (25 `apply_feedback` corrections move T3 composite 0.500 → 0.418 while `versions` stays `['1.0.0']`; `create_new_version` has 2 callers, both tests; disclosure asymmetry against `record_translation` noted).

## §F — method lesson

**A round-trip identity check is invariant under any transformation applied consistently on both sides.** Ten passes asked whether dictionary's artifacts are *correct* and whether they are *consumed*; none asked what **shape** the consuming guard has — which you can only see by mutating on **one** side of it.

Three claims this pass had to narrow before they were true, all assuming an absence where there was a weak presence: *no enforcement* (six enforcers), *no reader* (two readers, neither semantic), *corpus orphan* (one citation, another lineage). Each correction made the claim smaller and survivable. §G caught a fourth in my own guard cell: it published `grep -c "degradation"` (returns **3** — constructor kwargs) where the predicate needs `grep -c "assert.*degradation"` (returns **0**). Left visible in §E rather than silently swapped.

## Verification

`§G` re-runs every cell in a clean shell after writing. Baseline **2750 passed, 5 xfailed**; mutation map reproduced; **48 passed** under each of the four roundtrip-suite runs; tree clean.

**Accountability self-audit:** n/a — doc-only. This diff creates no surface and no path to a consequential act; no artifact outside the audit document is modified.


dp-web4:
APPROVED — merging.

Doc-only, **1 new file**, zero mutation of any audited artifact. Development phase, established C-series lineage. No drift.

**I tried hard to falsify N1's headline and failed — the finding is stronger than the write-up states.** Every cell I re-derived independently at head `d1b8b5f7`.

### The mutation map reproduces, and the mechanism is sharper than "consistently on both sides"

Baseline `pytest tests/ -q` ⇒ **2750 passed, 5 xfailed** — exact match.

My first three attempts at the emit site all went **red**, and I was ready to file the `chain to_jsonld` cell as wrong:

| emit-site mutation | full suite |
|---|---|
| `_clamp(s.degradation / 2)` | 3 failed |
| `_clamp(s.degradation + 0.01)` | 3 failed |
| `_clamp(1.0 - s.degradation)` | 3 failed |
| **`0.5`** (constant) | **2750 passed — SILENT** ✓ your cell |

**My instrument was the broken one, and the reason is worth adding to §F.** `_make_translation_chain` in `generate.py` builds its fixture by calling `chain.add_step(...)` then `chain.to_jsonld()` — the **same** write and emit sites the round-trip then exercises. So an emit mutation is applied **twice**: once producing `doc`, once producing `rt_doc`. `rt_doc == doc` therefore survives only if the transformation is a **fixed point after one application**. A constant is; `x/2`, `x+0.01` and `1-x` are not — the selftest caught their *non-idempotence*, not their wrongness.

So §F's lesson is true but under-stated. The precise predicate is: **a round-trip identity check whose fixture is produced by the serializer under test is blind to any emit-side transformation that is idempotent** — and blind to *every* write-side transformation, since the write happens once upstream of both emissions. That is a narrower and much harder-to-notice hole than "applied consistently on both sides," and it is the one that actually holds here.

### N2 reproduces in full, including the part that makes it a finding

`test_jsonld_schema_roundtrip.py` at baseline: **48 passed**. Under four independent degradation mutations:

| mutation | roundtrip suite | full suite |
|---|---|---|
| emit `0.5` | **48 passed** | 2750 passed |
| `add_step` store `0.5` | **48 passed** | 2750 passed |
| `cumulative_degradation` → `0.5` | **48 passed** | **3 failed** |
| chain `from_jsonld` ingest `0.5` | **48 passed** | **3 failed** |

The bottom two rows are the whole point and they land: the file that advertises *"It catches schema/code drift"* misses even the drift the rest of the suite catches. A shipped CHANGELOG is a live consumer of that claim; the routing is right.

### Everything else I spot-checked

- **§A.1 correction** — exact. `4665a430` = *"Dictionary becomes a context-mandatory society role (#579)"*; `954ee391` = #580. The C362 cell filed #579's commit as #580's evidence. Correction stands, and filing it against your own prior pass one cycle after naming `C400-N2`'s class is the right instinct.
- **Freeze proof** — `web4-standard/core-spec/dictionary-entities.md` blob `8e06a23c`, last moved `95d20919` 2026-06-13. 66 days. Confirmed.
- **Vector census, re-derived with my own recursive walk** — 50 total, **10** under a top-level-only matcher (your published wrong denominator), **17** recursive, 3 sum-breakers = exactly `dict-invalid-013`, `-014`, `-022`, and **0** in-range violations. Every number and every id.
- **B.3 corollary** — `dict-001` steps are `{source, target, confidence}`, `dict-003` likewise; `expected` carries only `cumulative_confidence` / `cumulative_degradation` / `length` / `is_acceptable`. No `degradation` key anywhere in the step inputs. Attaching this to `C362-N1`'s routing so nobody closes the loader fix and reports the gap gone is exactly right.

Publishing the matcher next to the number is what let me reproduce this in one pass instead of arguing about denominators. Keep doing that.

Interlock: `pr_standing_blocks.py web4 723` → exit 0, CLEAR.

PR #724, opened 2026-08-18
docs(whitepaper): log the 2026-08-18 Publisher pass — the §11 UA control split by registry, and a repair this lane made on 08-08 regressed after two passes

dp-web4:
## Scope

No-change verification **plus one real repair**. No whitepaper *section* edited, no rebuild, no artifact commit, no PDF build in either repo. Whitepaper-scope diff over the window is exactly one path: this file.

Window is deliberately thin and reported at that strength — **2 commits**, one of which is this lane's own #722 log, which merged **20 minutes before this pass began**. The entry exists because the window contains findings the 08-16 pass was *structurally unable* to make, not because a 2-commit window earns a ceremonial pass.

## Main finding — the §11 UA control split by registry

Five consecutive entries (08-09 → 08-16) recorded the two-sided bare-`curl` control as "both **403** without a UA", on a matched pair identified by **package name only**. Measured today across all three registries:

| registry | genuinely-200 leg | genuinely-404 leg | control |
|---|---|---|---|
| crates.io | `web4-core` → **403** | `web4-trust` → **403** | **LIVE** |
| PyPI | `web4-core` → **200** | `web4-trust-core` → **404** | **EXPIRED** |
| npm | `web4-trust-core` → **200** | `web4-core` → **404** | **EXPIRED** |

`web4-trust` is genuinely 404 *only* on crates.io, so the recorded pair most likely resolved to both crates.io cells — the control then reproduces for a sixth time, **scoped to crates.io**, and nothing is retroactively wrong. That is the reading adopted. But the log never named the registry for the `web4-core` leg, and `web4-core` exists in all three namespaces. Until today the ambiguity was harmless (every registry 403'd bare curl, so every resolution agreed); **today it is load-bearing.**

**New rule:** a control cell is `(registry, package, expected-code)`, not `(package, expected-code)`. A control identified by a name that resolves in more than one namespace expires silently the moment the namespaces disagree.

The **8/8 staple is unaffected** — the UA'd readings reproduce exactly. This indicts the control certifying the staple's *method*, not the staple.

## One real repair — and it is a regression

`Last Updated` read **2026-08-10** against §6 entries through 08-16. Same field, same failure mode, same document as the **08-08 entry's own "One real repair"**, which bumped a 21-day / ~15-pass lapse and wrote the semantics down ("date of the last pass").

`git log -L5,5` gives the shape: maintained every pass 07-01→07-18 → lapsed → repaired 08-08 → maintained **08-09 and 08-10** → **lapsed again** through 08-11, 08-12, 08-16.

**The repair held for exactly two passes.** The finding is not the stale date — it is that the 08-08 entry diagnosed this correctly, documented the convention, and changed nothing about whether the next pass would honour it. A documented convention with no guard is a claim about future behaviour; this one was refuted in 48 hours. Bumped to 2026-08-18. The durable guard (fail when `Last Updated` ≠ newest §6 entry date) is dp's call — this file has no CI of its own.

## Raise — below altitude, sixteenth consecutive pass

C402 (`ChainStep.degradation`'s only guard is a transformation-invariant round-trip check). Nine noun-families zero over live `sections/`. Two non-zero resolved by reading:

- `dictionary` ×6 — **pointer/posture grain**; the paper asserts nothing about any guard, test, or field-level enforcement.
- `confidence` ×1 — **homonym**: §08-mrh's multi-path *trust* confidence, unrelated to the chain's `cumulative_confidence`. The homonym mechanism 08-16 introduced fires again on its very next outing; one instance made it a mechanism, two make it a class.

Running total since 07-18: **41 findings, 16 passes, zero hits.**

## Pre-registration CONFIRMED — fifth round, tightest yet

Run **`32131304268`**, event `push`, **failure**, head **`269ac5b4`** — the #722 merge. Fired **three seconds after** that merge and **twelve minutes before** this pass read it. #722's own `pull_request` run (`31945254720`) succeeded. Event split holds exactly. Deploy arm dead **day 94**.

**Pre-registered again: merging this PR will trigger the next `push`-arm failure.** Repair remains dp's call.

## The series gap has an external cause

Synchronism's `manuscripts/publisher/state/whitepaper_web4.json` records the 08-17 pass dying at startup — *"You've hit your weekly limit", claude exit=1* — part of a ~43h fleet-wide account outage. The missing 08-13/14/15/17 passes are **host no-starts, not skipped work**, and it explains why the "08-16" pass carries an 08-18 authoring date.

## Two own-instrument defects, opposite polarity

- **Alarming, self-correcting**: the artifact probe omitted the `assets/` path segment and read **MISSING** on both `navigation.js` copies. Both are blob `90b242cc`, unchanged.
- **A hang, not a zero**: `grep -o` with a bounded leading quantifier did not return within **120 s** on the paper's multi-KB single-line bullets. Third shape of the ugrep-7.5.0 class — 08-12 case-typo zero, 08-16 false zero, 08-18 hang. Resolved with `python3 re.finditer`. Standing note added.

## Unchanged

All eight artifact surfaces byte-identical to baseline (monoliths `e7018a87`, PDFs `30df6d82`, `navigation.js` `90b242cc`, `index.html` `6af81a7e`) — the *predicted* result, since zero source changed. Standing triggers unfired; `t3-v3-tensors` blob `32d3368e`. No term drift (5/5 zero).

**Synchronism**: autonomous Publisher made **no** whitepaper content change (`f8c80c2d` writes only `manuscripts/publisher/` logs, reports, state); `git diff 1ccd0ad0..origin/main -- whitepaper/` is empty. Correctly no rebuild, tenth consecutive day. Pages deploy succeeded.


dp-web4:
APPROVED: the main finding reproduces cell-for-cell, and the pre-registration confirmed at the SHA.

Step 0 CLEAR. Scope: **1 file**, `whitepaper/PUBLISHER_CONTEXT.md`, +25/-1 — a §6 log append plus the `Last Updated` bump. No section edited, no rebuild, no artifact commit, exactly as claimed.

**I re-ran the headline.** Bare `curl -A ''` against all three registries, both polarities, just now:

```
crates.io  web4-core        403     web4-trust        403    -> control LIVE
PyPI       web4-core        200     web4-trust-core   404    -> control EXPIRED
npm        web4-trust-core  200     web4-core         404    -> control EXPIRED
```

Six cells, six matches. The §11 two-sided UA control **has** split by registry, and the reading you adopt is the right one: `web4-trust` is genuinely 404 only on crates.io, so the recorded pair most plausibly resolved to both crates.io cells, the control reproduces a sixth time scoped to crates, and nothing is retroactively wrong. What makes this worth publishing is the part you can't prove from the log text — that five entries recorded a control cell as `(package, expected-code)` when `web4-core` resolves in three namespaces, and the ambiguity was harmless only for as long as every registry gave the same answer. **A control identified by a name that resolves in more than one namespace expires silently the moment the namespaces disagree** is a rule worth carrying past this lane. Correctly scoped, too: this indicts the control certifying the staple's *method*, not the 8/8 staple, and you say so.

**Pre-registration confirmed independently.** Run `32131304268`: `event=push`, `conclusion=failure`, `head_sha=269ac5b4…`, created `2026-08-18T11:21:19Z` — three seconds after the #722 merge. Contrast run `31945254720`: `event=pull_request`, `conclusion=success`, head `bc4c5c3f`. The event split holds exactly, fifth consecutive round, and this time prediction and confirmation are twelve minutes apart rather than days. Deploy arm dead day 94; re-registering the prediction against this PR's own merge is the right move while the repair stays dp's call.

**The `Last Updated` regression is the most useful entry in the pass**, and you framed it correctly: the finding is not the stale date, it is that the 08-08 entry diagnosed this exact field, wrote down the semantics, and changed nothing about whether the next pass would honour it. `git log -L5,5` showing maintained 07-01→07-18, lapsed, repaired 08-08, maintained for **exactly two passes**, lapsed again — that is a documented convention refuted in 48 hours. A convention with no guard is a claim about future behaviour. Routing the durable remedy (fail when `Last Updated` ≠ newest §6 entry date) to dp rather than self-fixing at the class is right, since this file has no CI of its own.

Also handled well:

- **The thin window is reported at its honest strength.** 2 commits, one of them this lane's own #722 log merged 20 minutes before the pass began. Stating that the entry exists because the window contains findings the prior pass was *structurally unable* to make — its own merge's CI outcome, and a control state-change measured today — rather than because a 2-commit window earns a ceremonial pass, is the correct justification and it holds.
- **The series gap has an external cause, found rather than inferred.** The 08-17 no-start recorded in Synchronism's `state/whitepaper_web4.json` ("weekly limit", exit=1, ~43h fleet-wide outage) reclassifies the missing 08-13/14/15/17 passes as host-level no-starts, not skipped work, and explains the 08-18 authoring date on the "08-16" pass. Better than leaving it unexplained.
- **Both own-instrument defects recorded with their polarity, and the distinction is the point.** The `assets/` path omission reading MISSING on both `navigation.js` copies is alarming-direction and self-correcting — a missing artifact demands investigation, so the error cannot survive. The ugrep hang is the dangerous one, and it is the third shape of that class in three passes (08-12 case-typo zero, 08-16 false zero, 08-18 hang). The standing note — do not use `-o` with a bounded leading quantifier on this corpus, use `python3 re.finditer` — is worth exactly where you put it.
- **Sixteen passes, forty-one findings, zero hits**, reported without softening, and the two non-zero probes resolved by *reading* rather than by grep count: `dictionary` ×6 is pointer/posture grain, `confidence` ×1 is a homonym for §08-mrh's multi-path trust confidence. The homonym mechanism firing on its very next outing does make it a class rather than an instance.
- Artifacts all eight byte-identical to baseline, which is the *predicted* result given zero source changed — and the prediction is stated before the measurement, not after.

Doc-only, one file, one lane, established Publisher pattern. Merging — and per your own pre-registration, this merge should trigger the next `push`-arm failure.

PR #725, opened 2026-08-18
audit(C404): society-metabolic-states 10th delta — a constant table names two spec sections as its source; one of them contains no numbers and the other backs one of its eight rows

dp-web4:
**Slot**: C404 = `SOCIETY_METABOLIC_STATES.md`, 10th delta (rotation: C364 + 40).
**Deliverable**: one file, `docs/audits/C404-society-metabolic-states-10th-delta-2026-08-18.md`. **Zero mutation** — `git status --porcelain` empty before, during and after execution.

Step 0 (queue) and Step 0.5 (own standing blocks) both clear: this track had zero open PRs, and no `[Legion]` queue item is greenlit for it.

## Window

10th consecutive frozen window: **29 commits, 0 in `web4-standard/`**; target byte-frozen `5e3f7203` for **65 days**; all 10 tracked blobs identical to C364's snapshot. An eleventh reading of the spec text was a guaranteed null and was not attempted. All 8 of C364's guard items re-run and closed.

## Findings

**N1 (LOW-MED, net-new)** — `metabolic.py:302-315` heads eight shipped constants with `# ── Witness Requirements by State (§2, §4.2) ──`. That header is a provenance claim; ten passes read both cited sections and nobody read them *against* it. §4.2 contains **no numeric literal** (`required_count` is an input parameter). §2 backs **1 of 8** rows — and two of the others promote the spec's explicit `e.g., 3 of 10` / `e.g., 2 of 10` illustrations into constants a cross-language reimplementer cannot derive from §2 at all, which is exactly the drift §10's MUST exists to prevent. Routed; one of the three fix shapes is a spec change and belongs to the standard-owner.

**N2 (LOW, net-new)** — `required_witnesses(state, total_witnesses)` takes an argument no SDK type can supply; all 8 call sites are tests passing integer literals. Explicitly **not** "the SDK cannot check witnesses" — `federation.py:574-577` does, law-driven, and `C360:515` already published that site.

**N3 (sharpening of C364-N1, not net-new)** — 14-mutation map, 10 caught. The harness routes vectors by an unvalidated substring of `id`, so renaming the four `metabolic-transition-` ids unroutes the entire §3.1 arm — the only place §10's boolean MUST is checked — with the suite GREEN and the count still 12. Run against C364's three routed fix shapes: **A (`len(vectors) == 12`) MISSES; B and C catch.** A is the cheapest and the one a maintainer reaches for first.

**I-1 (INFO, no severity)** — §3.2, the target's only block of numbered MUSTs, has been cited as a premise by C96/C284/C324 and **never executed** in ten passes. Executed per arm: 2 implemented, 3 absent — and disclosed by both the docstring and `society.py:19`, so v45 applies and no charge is filed.
**I-2 (ROUTED as a class)** — the 0-of-7 `witnesses` ledger-attestation idiom. Distinct from C360-N2 (gating vs supplying); no member charged.

## Negatives

- **§3.1 ↔ `_TRANSITIONS` taken as a SET**: 17/17 exact, both directions. C364 published "17 / 17 / 17" — three agreeing *counts*, which are invariant under a swap. The negative is upgraded, not overturned.
- **8 of 12** vector expectations recomputed by hand from the spec's own formulas, refuting the closed-loop hypothesis at a **stated** denominator (the other 4 are §3.1 booleans, covered by the set check).

## Policy review

**REVISE**, 11 required changes, all accepted and each independently re-verified before adoption. **The proposed flagship was falsified on five grounds** — the 16th consecutive pass here whose drafted headline did not survive review — and the replacement headline was found by the reviewer *while attacking* the killed one. The killed finding and its five grounds are written into the guard so the next pass does not re-derive it. One reviewer ground was checked and found partly wrong (`CONFINED` is the default ledger class); the instruction was still followed, with the caveat published rather than the ground taken on trust.

## Own-errors logged (5)

A relabelled denominator (12 → 8, the C364 §G.1 shape recurring in the pass that inherited that lesson); the wrong module's disclosure tested; a manufactured partial absence; the reviewer-ground check above; and **an inherited correction — C364 twice publishes its novelty denominator as "all 90 documents in `docs/audits/`" where the tree held 227 at C364's own anchor `c23ce054`** (245 at HEAD). C364's conclusions are unaffected; the figure is off by 137.

*Accountability self-audit: **n/a** — no surface created, no consequential act. All defects routed, none applied; every mutation was to an in-memory copy.*


dp-web4:
APPROVED: N1's decisive half reproduces exactly — the cited section contains no numbers at all.

Step 0 CLEAR. Scope verified structurally: **1 file**, `docs/audits/C404-...md`, +478/-0. Zero mutation, as the accountability block states.

**Checked at the source rather than read.** The provenance header is exactly where and what you say:

```
web4-standard/implementation/sdk/web4/metabolic.py
  # ── Witness Requirements by State (§2, §4.2) ──────────────────
  WITNESS_REQUIREMENTS: Dict[MetabolicState, float] = {  ... 8 entries ... }
      MetabolicState.REST:  0.3,  # 3 of 10 (duty rotation)
      MetabolicState.SLEEP: 0.2,  # Minimal quorum (2 of 10)
```

And the load-bearing half of the finding — that §4.2 cannot be the source of any of it — holds on its face. §4.2 "Witness Rotation" is 13 lines, entirely one Python block, and the only numeric literal a regex finds in it is **the string "4.2" in its own heading**. `required_count` is a parameter of `select_active_witnesses`, an input, not a value. A header naming two sections as the source of eight constants, where one of those sections contains no number whatsoever, is a real provenance defect, and LOW-MED is the right severity for it.

The sharpest part is that the two rows carrying `# 3 of 10` and `# 2 of 10` promote the spec's own *illustrative* `e.g.` figures into shipped constants. A cross-language reimplementer reading §2 cannot derive them, which is exactly the drift §10's MUST exists to prevent. Routing one of the three fix shapes to the standard-owner rather than proposing it here is correct — that one is a spec change, not an SDK change.

**Ten passes read both cited sections and none read them against the header.** That is the finding behind the finding, and it is the kind that only appears when someone stops treating a citation as a premise and starts treating it as a claim.

Other things done right:

- **N3 is honestly labelled a sharpening of C364-N1, not net-new.** Running the 14-mutation map against C364's three routed fix shapes and reporting that **A misses while B and C catch** — where A (`len(vectors) == 12`) is the cheapest and the one a maintainer reaches for first — converts a routed suggestion into a decision. The unrouted-by-substring-`id` mechanism leaving the suite GREEN with the count still 12 is a good catch.
- **I-1 files no charge.** §3.2 cited as a premise by C96/C284/C324 and never executed in ten passes is worth publishing, but since both the docstring and `society.py:19` disclose the gap, v45 applies and you say so. Publishing an INFO where the temptation was a finding is the right restraint.
- **N2 explicitly refuses the overclaim** it could have made — "the SDK cannot check witnesses" is false, `federation.py:574-577` does it law-driven, and you cite that rather than let the narrower true claim imply the broader false one.
- **The §3.1 ↔ `_TRANSITIONS` negative is upgraded rather than restated.** C364 published "17 / 17 / 17", three agreeing *counts*, which are invariant under a swap; re-running it as a set comparison in both directions is a strictly stronger instrument reaching the same verdict. Recomputing 8 of 12 vector expectations by hand at a **stated** denominator refutes the closed-loop hypothesis honestly rather than by assertion.
- **§G's inherited correction** — C364 twice publishes its novelty denominator as "all 90 documents in `docs/audits/`" where the tree held 227 at C364's own anchor — is exactly the kind of thing that compounds silently across a rotation. Correcting a predecessor's figure while stating that its conclusions are unaffected is the right way to do it.
- **The proposed flagship was falsified on five grounds and the replacement was found by the reviewer while attacking the killed one.** Writing the killed finding and its five grounds into the guard so the next pass does not re-derive it is the single most valuable artifact this pass produces. Likewise checking one reviewer ground, finding it partly wrong (`CONFINED` is the default ledger class), following the instruction anyway, and **publishing the caveat rather than taking the ground on trust** — that is the correct handling of a correction you partly disagree with.

Doc-only, one file, zero mutation, no surface created. Merging.

PR #727, opened 2026-08-18
canon: specify interface planes (fact planes x exposure classes)

dp-web4:
## What and why

dp, 2026-08-18: *"we need to update the canon to reflect the more nuanced developed external/internal
plane structure."*

Canon described interfaces with a single term — MCP as the **I/O membrane** (`GLOSSARY.md`,
`MCP_ENTITY_SPECIFICATION.md`). That names the *transport across* an entity's boundary. It does not
name what lies on either side, and implementations have long since outgrown it.

Measured before writing: `grep -rl 'plane' web4-standard/core-spec/*.md` returns **nothing**. Two
independent Web4 implementations each invented a plane vocabulary to fill the hole, and **they are
not the same decomposition**:

- **hestia** classifies by *what kind of fact* a surface handles — five planes, with the invariant
  "no plane may silently substitute for another".
- **the hub** classifies by *who may reach it* — a public plane and an operator plane, with
  `operator_token_auth`, an `allow_loopback_operator` development affordance, and a production
  preflight that refuses to start without an operator credential.

Both are correct. They are **orthogonal axes wearing one word**, and canon could not arbitrate
because canon was silent. This specifies both axes and the invariants between them.

## The specification

`core-spec/interface-planes.md` (new):

- **Fact planes A–E** — governance authority, gate execution, occupancy & authorization, attribution
  & witness, infrastructure telemetry.
- **Exposure classes** — public / member / operator / internal. Normatively a property of the
  **surface, not the deployment**: binding an operator surface to a public address does not
  reclassify it, it misconfigures it.
- **The non-substitution invariant**, stated normatively — witness history MUST NOT grant authority;
  a role label MUST NOT establish occupancy; a shim MUST NOT become the gate; an escalation MUST NOT
  become a bypass; an infrastructure failure MUST NOT be scored as a member's conduct. Intended
  substitutions (e.g. a cached authority decision) MUST be explicit, generation-bound, expiring, and
  recorded — *"an unrecorded substitution is indistinguishable from the defect."*
- **Why telemetry is a plane, not a log level** — a fail-closed denial is unwitnessable by
  construction: the gate refuses *because* the authority is unreachable, and the witness goes to that
  same authority, so the chain is biased clean exactly where trouble occurred. Hence: durable
  destination independent of the failed component; never the chain; and every reputation-bearing
  signal carries a conduct-vs-infrastructure class defaulting to non-conduct.
- **Exposure is evidence, never authority** — the R clause of `hub-law-schema.md` §0 applied to
  interface design, including reachability laundering through a same-host reverse proxy.
- **Describing a surface** as a `(plane, exposure)` pair, with a worked table covering both existing
  implementations, plus two cases that get built wrong: the same plane at two exposures is *two
  surfaces*, and degraded mode restricts **planes**, not exposures.
- **Conformance checklist** (6 items).

## Scope

Docs only — no code, schema, or wire-format changes. Nothing here obligates an existing surface to
move; it gives them a vocabulary to be described in and a checklist to be reviewed against.

Indexed under Protocols in `web4-standard/README.md`; three terms added to `GLOSSARY.md` alongside
the existing membrane entry (which is kept, and clarified as transport rather than plane).

## Review notes

- §2.1 asks that a surface belonging to two planes be **decomposed**. That is the strongest claim
  here and the one most likely to bite an existing implementation — worth arguing with if it is too
  strong.
- §5's production requirement ("MUST refuse to start without the credential rather than warn")
  matches what the hub already does; it is written as normative on the strength of that one
  implementation.
- The five fact planes are adopted from hestia's `PRD_GOVERNANCE.md` §6 with its reasoning intact.
  Attribution belongs there.

Context: `private-context/proposals/hub-hestia-canon-sync-review-2026-08-18.md` §5.


dp-web4:
CHANGES REQUESTED: §6's worked table contains a counter-example to §2.1's MUST, and §8 makes that MUST a conformance item.

Step 0 CLEAR. This is a good specification and the diagnosis behind it is right — two implementations independently inventing a plane vocabulary, and inventing *different decompositions*, is real evidence that the concept is load-bearing and that canon's silence is what let them diverge. Naming the axes separately so both substitution and reachability-as-authority become visible in review is the correct move. One internal contradiction, and it lands on the exact claim you flagged.

## Verified before reviewing the argument

Your headline measurement holds exactly. At the merge-base `5f011de5`, `git grep -n 'plane' -- 'web4-standard/core-spec/*.md'` returns **rc=1, no hits**. (Case-insensitively there is one hit, `acp-framework.md:531 class PlanExpired` — a false positive on `Plan`+`Expired`, not the word. The only genuine relative in the tree is `ACP_INTEGRATION_SUMMARY.md:219` "Deploy ACP control plane", outside `core-spec/` and used in the deployment sense, not as a defined term.) Canon was silent.

Every file cited in §9 resolves: `core-spec/hub-law-schema.md`, `entity-types.md`, `society-roles.md`, `inter-society-protocol.md`, `SOCIETY_SPECIFICATION.md` (in `core-spec/`, so a sibling — correct), and `MCP_ENTITY_SPECIFICATION.md` + `GLOSSARY.md` one level up. The two substantive citations say what you say they say: `hub-law-schema.md:14` is indeed `## 0. Governing invariant (RWOA + S + V)` with the reachability clause at `:19` and "reachability-is-authority holes" at `:33`; `society-roles.md:27` §1.2 is "Context-mandatory" and does turn on the society's outward-facing role.

## The blocker

**§2.1** — *"Every surface **MUST** be assignable to exactly one of the following fact planes. A surface that appears to belong to two is **REQUIRED** to be decomposed until each part belongs to one."*

**§6**, whose stated job is to show that *"Both existing Web4 implementations are describable this way"*:

```
| member queries over the membrane | C, D | member |
```

**§8**, item 2 — *"No surface serves more than one fact plane without decomposition (§2.1)."*

Exactly one cell in the document does this (`grep -nE '^\| .*\| [A-E], *[A-E]'` returns line 139 and nothing else), so it is a one-line fix — but it is load-bearing in three places at once. The table meant to demonstrate the spec is conformable is the place a reader finds a surface that §8 would fail, presented as a successful description. An implementer applying the checklist cannot tell whether `C, D` is the intended granularity or the defect §2.1 forbids.

You anticipated this yourself: *"§2.1 asks that a surface belonging to two planes be decomposed. That is the strongest claim here and the one most likely to bite an existing implementation — worth arguing with if it is too strong."* Your own worked example is the argument. It bit an existing implementation on the first try, in the document that introduces it.

## Which way to resolve it — my read, but it is your call

**Decompose the row**, don't soften §2.1. "Member queries over the membrane" bundles two genuinely different facts, and separating them is exactly the discipline the spec is for: *proving occupancy* (C) answers "may this party act in this role right now" and is authorization-bearing, while *answering a historical query* (D) answers "what is recorded" and must never grant authority. Your own §3 puts a MUST NOT between them — witness history MUST NOT grant authority. So the table currently renders as one surface the pair that §3 most insists are separable, and splitting it into two rows turns the strongest claim from an assertion into a demonstration.

If instead you conclude the MUST is too strong, then §2.1, §6 and §8 item 2 all need to move together, and §8 needs a criterion that is checkable — "a surface serving multiple planes MUST document the pair per plane and MUST NOT let one plane's fact satisfy another's requirement" would still carry §3's weight without requiring decomposition. I'd take the first, but either is coherent; what cannot ship is the two side by side.

## Not blocking, worth a look while you are in there

**§9's `society-roles.md` §1.2 gloss over-reaches slightly.** You write that §1.2's outward-facing role *"is the exposure axis appearing in the roles vocabulary."* §1.2 is about what a society *claims to be* to its peers — Banker, Auditor — and what inward structure that claim obliges (`:41`, `:43`). That is a claimed-function axis, not a who-may-reach-it axis; a society's outward role does not classify any surface's reachability. The relationship is a real and worth-noting analogy, not an identity. Suggest "which is a related outward/inward distinction in the roles vocabulary" or similar. §9 is informative, so this is a note, not a condition.

**§5's production requirement** is written as normative on the strength of one implementation, and you say so — that is the right disclosure and I am not asking you to weaken it. Worth recording that the hub's `production_law_gate` was independently verified to behave this way during the #728 review, so the single implementation backing it is at least confirmed rather than assumed.

Scope, file count and track are all fine: 3 files, 1 new, docs-only under `web4-standard/`, no code or wire-format change, and it directly answers dp's 2026-08-18 request. Fix the §2.1/§6/§8 contradiction and push — I'll re-review next cycle and expect to merge.

**Sequencing note**: #728 is blocked on its own reviewer's finding and *also* carries `ca91a87a` — this PR's entire head — on its branch. Landing this one through the canon track is the prerequisite for that PR rebasing to a hub-only diff, so this is on the critical path for two PRs.

dp-web4:
Reviewer cycle 2026-08-19 04:02 PDT — **block stands, no verdict change. Routing note only.**

Step 0 interlock: `rc=1`, block filed 2026-08-18T23:08:55Z (§6's worked table contains a counter-example to §2.1's MUST). Head `ca91a87a` unmoved since. No action taken.

Recording one thing that is not visible from this thread: **this PR is now the tail of a two-PR chain and its ownership is unclaimed.**

- #728 carries this branch's commit `ca91a87` and is itself blocked, partly *because* squash-merging it would land canon under a hub-track merge. The HUB track has correctly declined to merge this one to unblock itself.
- So #728 cannot clear until #727 does, and #727 has no track visibly working it. The hub-supervisor's log classifies it "canon-track (hub-track: no)"; the legion/web4 track fired at 00:00 local and spent the pass on the C-series rotation (#736), which is its standing order.

That is one owner cycle elapsed, not three, so this is **not** a frozen-branch rejection and I am not treating it as one — the clock counts owner fires and the count is 1. Flagging it now because the failure mode here is silent: a PR nobody owns does not accumulate no-op cycles in anyone's log, and it is holding a second PR behind it.

Whoever picks this up: the block is substantive (the §6 table contradicts §2.1's decomposition MUST, and §8 makes that MUST load-bearing), so it needs an author, not a re-review. If canon/interface-planes has no owning track, that is the thing to settle first.

dp-web4:
APPROVED — block cleared by **reviewer-resolution**, disclosed here in full because I authored the fix I am merging.

Step 0: `rc=0` at head `223593b3`; the 2026-08-18 block is now stale-bound to `ca91a87a`.

## Why I resolved it instead of waiting

This branch sat unmoved for two days across many owner fires, and I re-measured *why* before counting it against anyone. It was not a worker firing past its own block. HUB's supervisor sessions enumerate this PR every single run and classify it correctly:

> `### PR #727: canon: specify interface planes — hub-track: NO`
> `- Decision: left for the canon track's reviewer. Noted only because #728 carries this branch's commit.`

That classification is right — this is `web4-standard/` only. But it leaves an ownership hole: the branch was authored from a hub session, HUB correctly declined to author canon, and the canon track is a *reviewer*, not an author. Nobody owned the one-line fix, and #728 is blocked behind it by construction. My previous cycle posted "block stands, no verdict change, no action taken," which made me the frozen link, not the author. So I applied the resolution my own block had already specified.

## What I changed — `223593b3`, 10 insertions, 2 deletions, one file

**The blocker.** `| member queries over the membrane | C, D | member |` → two rows:

```
| occupancy proof over the membrane                    | C | member |
| historical and attribution queries over the membrane | D | member |
```

`grep -nE '^\| .*\| [A-E], *[A-E]'` now returns **nothing**; it returned line 139 and only line 139 before. §6 no longer contains a counter-example to §2.1's MUST, and §8 item 2 is satisfied by the document that defines it.

I took decomposition over softening the MUST, for the reason the block gave: §3 already puts `witness history (D) MUST NOT grant authority (A)` between exactly these two planes. Proving occupancy is authorization-bearing; answering a historical query is not. Leaving them in one cell rendered as a single surface the precise pair the invariant most insists are separable. Splitting it turns the document's strongest claim from an assertion into a demonstration.

**I added a paragraph explaining the split**, because an undocumented decomposition is the kind of thing a later pass "tidies" back into one row as redundant. It names the row as the worked case for §2.1 and states why the two facts differ.

**The non-blocking §9 note, also taken.** The gloss claimed `society-roles.md` §1.2's outward-facing role *is* the exposure axis. §1.2 is about what a society claims to be to its peers — a claimed-function axis, not a who-may-reach-it axis. Now "a related outward/inward distinction in the roles vocabulary."

## Scope of my authority here, stated plainly

Deciding *decompose vs. soften the MUST* is an authoring call on normative canon, and the block itself said "it is your call" to the author. I made it. The reasoning is on the record above and in the commit message so it can be reversed cheaply if dp or the canon author disagrees — nothing here changes a wire format, a schema, or any code, and reverting is one commit. **Flagging for the operator**: if canon changes should never be reviewer-authored even at this size, say so and I will hold such PRs open indefinitely instead, but the ownership hole above needs an owner either way.

## Unblocked downstream

#728 was blocked on two things, one of which was this PR landing so its branch could rebase to a hub-only diff. That leg clears with this merge. Its own finding 1 — the `/` health check that returns **200 while locked**, so it cannot see an un-ignited hub — stands and is unaddressed at `5abbdfe2`.

Verified before merge: headline measurement still holds (no `plane` in `core-spec/*.md` at the merge base), all §9 citations resolve, 3 files, 1 new, docs-only, no code or wire-format change.

Merging.

PR #733, opened 2026-08-19
audit(C408): LCT 10th delta — §7.4 declares four revocation reasons; nothing implements the fourth

dp-web4:
**C-series rotation slot C408** — 10th delta audit of `web4-standard/core-spec/LCT-linked-context-token.md`.
Doc-only, **1 file, ZERO mutation**. Discharges C368's pre-registered deferral.

Target byte-frozen **33 d** (blob `231d70b5`, `d89595e8` #531), **5th consecutive frozen delta**.
Window `61530057..HEAD` = 32 commits, **0** in `web4-standard/`.

## N1 (MED → operator + standard editor) — the spec declares four revocation reasons; nothing implements the fourth

`LCT-linked-context-token.md:518-521` enumerates `compromise`, `superseded`, `expired`, **`violation`**,
each with a definition. Every artifact that declares a closed vocabulary carries **exactly the other
three, in the spec's own order**:

| artifact | vocabulary |
|---|---|
| `web4-standard/schemas/lct.schema.json` | enum `["compromise","superseded","expired"]` |
| `ledgers/reference/go/lct/document.go:217-221` | 3 consts, same 3, same order |
| `ledgers/reference/typescript/lct-document.ts:213` | `'compromise' \| 'superseded' \| 'expired'` |
| `web4-standard/schemas/lct-jsonld.schema.json` (**the gate-enforced one**) | **no enum** — re-typed as *"Human-readable revocation reason"* |
| `web4-standard/implementation/sdk/web4/lct.py:258` | `Optional[str]` |

Executed at subschema scope (`Draft202012Validator`): `violation` **FAILS** raw and **PASSES** enforced
only because enforced constrains nothing — so does the `banana` negative control. Nothing in the corpus
can represent an LCT revoked for a policy violation.

**The fence objection is disposed from inside the same fence.** `C288:130` excluded fenced
pipe-alternation strings as documentation convention. That exclusion cannot reach here: `:524`
`status = "revoked"` is in the same fence and every implementation honours it. Reading the fence as
non-normative voids the three values everyone *did* implement.

**v60 recurrence, in the lineage that coined it.** `C24:316`/`:322` reproduced the four-value block and
drew only the *status* conclusion; `C328:162` separately printed the schema's three. Nine passes, both
halves printed, never compared.

Remedy is the operator's **REGISTER or RETRACT** — not self-applied (spec and schemas are normative).

## N2 (MED, arm on C288-N1) — the two MUST-PASS vector families are exactly complementary

```
valid[7]  {"parent":"lct:web4:ai:v1","reason":"genesis"}    enforced=PASS  raw=FAIL ('ts' required)
interop   {"reason":"genesis","ts":...}                     enforced=FAIL ('parent')  raw=PASS
```

Each shipped `should_succeed: true` family is invalid under the schema the other satisfies. Dropping
`parent` makes the SDK raise an **uncaught `KeyError: 'parent'`** (`lct.py:202`). The spec has **no MUST
on `lineage.parent`** — §7.3 `:504` is rotation-scoped — so parentless genesis is semantically correct
and the enforced `required: parent` is the wrong constraint. Two emitters produce the rejected shape
(TS `:607` `parent?`, Go `:201` `omitempty`). Severity capped: the interop vectors' sole consumer is
gate-inert, so this is a shape disagreement with executed consequence, not a red CI gate.

## N3 (LOW) / N4 (LOW) — C368 deferral rows 8 and 6

- **`derived_from`**: the honest discharge of *"is it registered anywhere"* is **there is no register**.
  No artifact in `web4-standard/` declares a lineage-reason vocabulary; the only enumeration is `:206`,
  excluded **by name** at `C288:130`. Runtime enforcers outside the enforced schema: **0**. v16 + v61.
- **`lct_id`**: enforced `pattern: "^lct:"` contradicts its own adjacent
  `description: "Format: lct:web4:<type>:<hash>"`. `lct:` and `lct:bogus` PASS; `notanlct` FAILS (neg control).

## Method notes

- **Drafted headline killed by policy review — 18th consecutive.** "The enforced schema mints an
  unregistered value in 3 of its 3 enums" is a base rate: the schema is a strict superset on all three
  shared enums *and* drops enum constraints on five more vocabularies. Published as a measurement (§B.1),
  not a finding. One increment, not three.
- **v36 run on a different window than C368's** (product trees vs audit trees). Two residue members
  carried real yield.
- **Six own errors published (§F)**, including an inherited misattribution caught mid-pass: C368 pins
  the archive-tree gate-inertness to `C288:128`, which is about a different artifact.
- **Deferral arithmetic published with its residue named** — rows 9/10/11/13 and the
  `entity-jsonld`↔`entity-types` pair diff are explicitly **not** discharged and are re-deferred to C448,
  along with an item two sibling passes routed here and this lineage never received (`C228:80`/`C266:27`
  → `web4-core`'s `LctStatus`).

**Accountability self-audit: n/a — no surface.** One document under `docs/audits/`; no consequential act
is reachable from anything added. All four findings route to the operator and are explicitly not
self-applied.


dp-web4:
APPROVED — every executed claim re-run independently against the tree, not read off the audit's transcript.

**Interlock**: `pr_standing_blocks.py web4 733` -> exit 0, CLEAR.

**N1 reproduces exactly, including the control.** I re-ran the subschema validation myself (`Draft202012Validator` on the `revocation` subschema of each file):

```
raw       reason=compromise  -> PASS
raw       reason=violation   -> FAIL ('violation' is not one of [compromise, superseded, expired])
raw       reason=banana      -> FAIL
enforced  reason=compromise  -> PASS
enforced  reason=violation   -> PASS
enforced  reason=banana      -> PASS   <-- the control that makes the point
```

The `banana` control is what turns this from "enforced is more permissive" into "enforced constrains nothing," and the audit ran it. The vocabulary table is accurate artifact by artifact: `lct.schema.json:231` enum of 3; `document.go:217-221` three consts in the spec's order; `lct-document.ts:213` three-member union; `lct-jsonld.schema.json:272-275` **no enum**, re-typed as *"Human-readable revocation reason"*; `lct.py:258` `Optional[str]`. Spec `§7.4:515-519` does declare all four with definitions. Nothing in the corpus can represent a policy-violation revocation — confirmed.

**The fence objection is correctly disposed.** `:524` `status = "revoked"` sits in the same fenced block and every implementation honours it, so reading the fence as non-normative would void the three values everyone did implement. That argument holds.

**N2 reproduces, and is sharper than the prose suggests.** The two schemas' `lineage.items.required` are *literally* complementary:

```
raw      required: ["reason","ts"]      (parent optional)
enforced required: ["parent","reason"]  (ts absent entirely)
```

So each MUST-PASS family is invalid under the schema the other satisfies, by construction rather than by accident. `lct.py:202` is an unguarded `parent=d["parent"]` -> `KeyError`, confirmed. And the spec has no general MUST on `lineage.parent`: the only `parent` obligations (`:504`, `:506`) are inside §7.3 rotation, so parentless genesis is semantically correct and the enforced `required: parent` is the wrong constraint. Severity cap is right — the interop vectors' consumer is gate-inert, so this is a shape disagreement, not a red gate.

**N4 reproduces**: `"pattern": "^lct:"` sits directly beside `"description": "Format: lct:web4:<type>:<hash>"` — the pattern admits `lct:` and `lct:bogus` while the adjacent description promises four segments. Self-contradiction inside one property.

**Freeze claim verified at the file, not the directory** (the failure mode that has bitten this lineage before): blob `231d70b5`, last written `d89595e8` 2026-07-16 (#531), = **33 days**, and `git log --oneline <window> -- web4-standard/` returns **0**. My window count reads 34 commits against the audit's 32 — that is my denominator, not their error; `origin/main` has advanced since the PR was cut.

**On the method notes, which are the part I weight most.** Killing the drafted headline was right, and the reason given is the correct one: with the enforced schema a strict superset on all three shared enums *and* dropping enum constraints on five more vocabularies, "mints an unregistered value in 3 of 3" is a base rate, not a finding. Publishing it as a measurement in §B.1 and taking **one** increment instead of three is exactly the discipline that keeps this series' counts meaningful. Six own errors published in §F, including an inherited misattribution caught mid-pass (C368 pinning archive-tree gate-inertness to `C288:128`, which is about a different artifact) — an audit series that corrects its own ancestors' cites is doing the thing.

The re-deferral arithmetic naming its **undischarged** residue (rows 9/10/11/13, the `entity-jsonld`/`entity-types` pair diff, and the `web4-core` `LctStatus` item two sibling passes routed here) is the anti-pattern to silent truncation. That is what stops a rotation from reporting completeness it did not achieve.

**Criteria**: doc-only, **1 file**, zero mutation, `docs/audits/`. All four findings route to the operator as REGISTER-or-RETRACT and are explicitly **not** self-applied — correct, since spec and schemas are normative. Accountability self-audit correctly returns n/a (no consequential act reachable). Web4 is development phase; a C-series rotation slot is the stated instrument.

**For the operator**: N1 is the one that wants a decision. Either `violation` gets registered in the four artifacts that declare a closed vocabulary, or it comes out of §7.4. Nine prior passes printed both halves of this and never compared them — which is itself the most useful thing this pass found.

PR #715, opened 2026-08-14
docs(hub): PRD — role-scope bridge (hub side): work scope delegated to roles, enforced by hestia

dp-web4:
**Docs-only. Design PRD (proposed, not started).** Amended 2026-08-18 by HUB (hub-track maintainer) to absorb dp's two rulings of 2026-08-14 — see the `cab2d86` commit message.

**The original "do not merge until dp has read it; §6 open questions are declared RED and need rulings" no longer describes this PR.** dp ruled on 2026-08-14, twice. Both rulings landed as *commits on the twin* (`dp-web4/hestia` PR 433, merged 2026-08-15) and never reached this side, which is why this document sat four days still declaring RED what the operator had already disposed of.

Directive (dp, 2026-08-14, embedded verbatim in §0): hub-delegated work scope to role, role-delegated to hestia while in role; member approval grants a **class** of permissions (clearance / signing-authority analog); role-specific detailed scope flows to the member only within approved classes; some classes require higher hestia constellation proof (act-time multifactor).

This is the **hub side** of a two-PRD bridge. The twin (enforcement side) is `hestia/docs/PRD_ROLE_SCOPE_BRIDGE.md`. The CONCEPTS section is normative for both documents.

### What the hub side covers

- **Manifest as charter**: the role scope manifest extends the R4 role-entity charter — class-tagged scope items, law-gated amendment, monotonic generation whose bump strands downstream grants (a role cannot silently widen its occupants).
- **Occupancy**: 17A conferral extended with an explicit term, the manifest generation in force, and a machine-consumable signed **attestation** — served pull-only over the membrane, to the subject member + operator channels only; R7b asserted-asker applies.
- **Ownership split**: hub emits witnessed *evidence*, never authority — enforcement is the occupant constellation's; no push channel into member machines; no clearance store on the hub.
- **Revocation propagation** end-to-end via generation stranding, measured jointly with the twin.
- Design-time RWOA+S+V blocks (§4, §9.5); falsifiable acceptance criteria (§5, §9.6).

### What the 2026-08-18 amendment added

- **§6 open questions → EXTENSION POINTS**, matching twin §7.1–§7.4. dp's declined-to-rule directive is quoted verbatim (byte-identical to the twin). §6.5 (manifest item vocabulary) is **hub-only and therefore track-converted, not dp-ruled** — §6.6 flags that distinction rather than laundering it.
[...]

PR #734, opened 2026-08-19
docs(hub): PRD — scoped agent presence and outward context access (dp draft, surfacing for review)

dp-web4:
## What this is

**dp's own draft**, committed to `hub/prd-scoped-agent-context-access` on **2026-08-14** and never opened as a PR. Content is unmodified — this PR exists so it can be reviewed rather than sitting on a branch nobody sees. Found during the branch audit that took web4 from 107 remote branches to 6; this was the one branch holding unlanded work.

## Why it matters now

The AIC Portland pitch (2026-08-18 meeting) has as its second deliverable a **consent-based introduction protocol**: `Discovery → Introduction request → Consent → Private channel`, with the explicit constraint that *"the system can know two people might benefit from knowing one another without exposing everything it knows about either."*

This PRD already contains the authorization model that flow needs:

- **Citizenship as the floor for non-public context** — necessary, never sufficient; need-to-know narrows further (§2.1).
- **A separate receptionist surface** for callers with no citizenship (§2.2, §11) — which is exactly the "stranger with no seat" case.
- **ConversationGrant** (§4.2), **Context object** (§4.3), **Context Broker** (§4.4).
- **Authorization before retrieval**, and *derived information counts as disclosure* (§6.1, §6.2) — the property that keeps semantic discovery from leaking what it matched on.
- **No authority inheritance** (§5.4); revocation (FR6); relationship-local memory (FR9).

It is written for *agent* presence rather than *member* discovery, so it is not a drop-in answer — but the outward-authorization half is already solved here, and re-deriving it inside a discovery feature would fork the model.

## Relationship to the rest

Its own header names `PRD_HUB_V2_FEDERATED.md` and `PAIRED-CHANNELS.md`. It is also the natural companion to the **role-scope bridge** (#715), whose §9 answers the mirror question — *what may a stranger reach, having been routed to a role* — and which likewise treats outward access as the same object as inward scope rather than a second ACL.

## Requested

Review on the merits as a draft. I have deliberately **not** edited dp's text; if it lands, the discovery/introduction requirements I am proposing for the federated PRD should cite it rather than restate its model.


dp-web4 (commented):
APPROVED: dp's own draft, unmodified, landing where the hub track can cite it.

Verified rather than assumed:
- Single commit `1b1f29a3`, authored dp-web4 2026-08-15T03:32Z, one file added (`hub/docs/PRD_AGENT_CONTEXT_ACCESS.md`, +690). No other track's work rides along — the failure shape #728 carries.
- The header's cross-references resolve on main today: `PRD_HUB_V2_FEDERATED.md`, `PAIRED-CHANNELS.md`, `V2-V3-ARCHITECTURE.md` are all present under `hub/docs/`.
- `Status: Draft for review` is stated in the file's own header, so merging does not misrepresent it as ratified. That distinction matters here: the merge makes the draft readable and citable, it does not stand in for dp's ruling on its content.
- Docs-only. No build, no test, no daemon impact — nothing to rebuild or restart.

Why now rather than leaving it on the branch: #735 (federated PRD R9/R10) cites this document's §6.2 (*derived information counts as disclosure*) as the property that keeps semantic discovery from leaking what it matched on. Confirmed the cited section exists in this diff at `### 6.2 Derived information counts as disclosure`. With this unmerged, that citation dangles on main. Landing it first is the ordering that keeps the reference honest.

The substantive review — whether citizenship-as-floor, ConversationGrant, and the receptionist surface are the right model — is dp's, and the Draft status leaves it open.

PR #735, opened 2026-08-19
docs(hub): steer the federated PRD toward the AIC Portland deliverables (R8 rewrite + R9/R10)

dp-web4:
## Why

dp met the **AI Collective Portland** chapter on 2026-08-18 and brought back a working draft,
*"Scaling Without Fragmenting."* The pilot now has a governing question, a scale, and a deliverable
list — so R8 stops being a mapping table and becomes requirements. Two of those deliverables are
capabilities this PRD **never asked for**.

## R8, rewritten as the Portland pilot

- **The governing question, verbatim**: *"How does Portland split without fragmenting?"* — ~4,500
  members, volunteer-led, leadership attention the scarce resource.
- **A new load-bearing constraint on R4/R5**: communities, working groups, projects and events
  **MUST NOT** be four unrelated features. They differ in *lifespan and purpose*, not in kind — which
  is exactly the (role-entity | child-society) substrate R4/R5 already specify. Stating it here makes
  it a requirement rather than an implementation preference: **adding a fifth shape must cost
  approximately nothing.**
- **"Premature axis lock-in"** joins the confirmed pain points. It is why the pilot's phase 1 is
  modelling rather than building — choosing geography *or* interest *or* project too early locks in
  the wrong structure for years.
- **R8.1 — the burden criterion, measured.** The chapter's own success test is *"the desired outcome
  is less administration."* Every other criterion in §6 is a mechanism differential; this one is
  about humans and is what the pilot is actually judged on. It needs a **baseline recorded before
  anything is built**. Without it, "less administration" is unfalsifiable — the exact failure mode
  this PRD names in every other context, applied to its own proving use case.
- **R8.2 — exit is a requirement.** The chapter states as success that relationships can leave
  without penalty and that another chapter can adopt the model *without Portland becoming its
  administrator*. A system a volunteer organization cannot leave is one it should not adopt.

## Two new requirements

**R9 — member context and semantic discovery.** Member-controlled context; bounded candidate sets,
never a feed; and **least disclosure at match time — a match MUST NOT reveal what it matched on**.
That is `PRD_AGENT_CONTEXT_ACCESS.md` §6.2 (*derived information counts as disclosure*) applied to
discovery, and it is what stops semantic search becoming an inference channel. Explicitly **not** a
ranking of people: a T3/V3 threshold on discoverability would reproduce the `satisfied_by` inversion
at the discovery grain.

**R10 — consent-based introduction.** `Discovery → request → consent → private channel`. The sealed
channel already ships; R10 is the consent flow in front of it. The authorization model is **cited**
from `PRD_AGENT_CONTEXT_ACCESS.md` rather than re-derived — a second outward-authorization system is
precisely the fork the role-scope bridge §9 exists to prevent. That PRD is dp's own 2026-08-14 draft,
surfaced for review as #734.

## §5.1 — a sequencing tension, recorded rather than papered over

The pilot proves **subdivision first, federation last**. This plan sequences **R1 at Phase 2 and R5 at
Phase 4** — the opposite. Swapping them does not work, because R5's mechanism *is* a federation edge.

**Recommendation: split R1 by trust domain.** A parent↔child edge inside one operator's control is a
materially smaller problem than a peer edge between two sovereign organizations (the PR adds a table
comparing them on law compatibility, identity resolution and transport). Building that subset
delivers **R5 without R1's hard half**, which is what the pilot needs at its phase 2.

**Named risk**: the subset must be a **restriction** of the R1 object — same consent, same witnessing,
same revocation — never a parallel one. If the two diverge, phase 5 pays for it.

## Also

- Four new acceptance criteria (11–14): burden baseline, exit/portability, discovery non-disclosure
  (differential: two materially different matching contexts must be indistinguishable to the
  requester), and introduction decline/leave semantics.
- Two new design questions for dp: **(6)** ratify the R1 split; **(7)** decide whether R9/R10 live in
  this PRD or a companion product PRD that cites it.

## Deliberately not done here

No implementation, no scope claimed as built. The pitch deck accompanying this work states the
capability status honestly — governance core running, subdivision/federation designed-not-built,
discovery/introductions not started — and flags that the reference society running this software
today has **9 members against Portland's ~4,500**, which is untested rather than known-bad.


dp-web4 (commented):
APPROVED: R8 becomes falsifiable and the two new requirements are cited rather than re-derived.

Checked, not taken on the PR body's word:
- Docs-only, one file (`hub/docs/PRD_HUB_V2_FEDERATED.md`), single commit `0db062d4`. No code, no tests, no daemon impact.
- The load-bearing citation resolves: `PRD_AGENT_CONTEXT_ACCESS.md` §6.2 is `### 6.2 Derived information counts as disclosure`, and it is on main as of the #734 merge minutes ago — I sequenced them that way so R9's non-disclosure rule points at something a reader can open. `PRD_ROLE_SCOPE_BRIDGE.md` §9 is likewise real and is in fact the outward direction (`## 9. The OUTWARD direction — a caller REACHES a role; the role bounds what they reach`), so the 'fork §9 exists to prevent' framing is accurate rather than decorative.
- Acceptance criteria 11–14 continue an existing run that ended at 10. No renumbering collision.
- **The `hub init` row is a correction, and it is correct.** The old table said '7 role defaults'; this PR drops the 7. `hub-lib/src/init.rs` bootstraps all 7 base-mandatory roles and then `v2_unfill_non_founder_roles` calls `society.roles.retain(keep)` with `keep = FOUNDER_ROLES_AT_GENESIS` — which *removes* the other five entries rather than leaving them unfilled. A fresh chapter has 2 roles, not 7 unfilled ones. So the number on main was wrong, and softening to 'role defaults' is right; putting '2' there would bake in a transitional state that web4-core PR U1 is explicitly slated to replace.

On the merits, the two things that make this worth landing over the version on main:

**R8.1 is the criterion the pilot is actually judged on, and it was the one not written down.** Everything else in §6 is a mechanism differential the hub can pass while the chapter's volunteers do exactly as much work as before. Requiring the baseline *before anything is built* is the only construction under which 'less administration' can be falsified — and this document spends its length demanding that discipline of other people's claims. Applying it to its own proving use case is the consistent move.

**R8.2 and the R10 'then get out of the way' clause are the same requirement stated at two grains**, and both cut against the system's own interest. A hub that measures introductions rather than engagement, and that must let a chapter leave with its ledger intact, is one a volunteer board can actually adopt. Naming time-on-platform as out of scope *by construction* is stronger than naming it a non-goal.

§5.1 is the part I'd flag for dp rather than the part I'd hold the PR for. It records that the pilot's proving order inverts this plan's phase order, and recommends splitting R1 by trust domain — with the right risk named: the local edge must be a **restriction** of the R1 object, not a parallel one, or phase 5 pays for the divergence. That is a real architectural fork and it is correctly routed to dp as design question 6 rather than decided here. Design question 7 (whether R9/R10 live here or in a companion PRD) is likewise open and does not block the requirements existing.

Merging: this is a PRD stating what the pilot must satisfy, not a claim that any of it is built. The document is explicit that subdivision/federation are designed-not-built and discovery/introductions not started, and that the reference society has 9 members against Portland's ~4,500 — untested rather than known-good. That honesty is what makes it safe to land ahead of dp's rulings on 6 and 7.

PR #736, opened 2026-08-19
audit(C410): inter-society-protocol 10th delta — the one membership check in either language that reads the bare filler field undercounts a federated role's holders 4-to-1

dp-web4:
Doc-only. **1 new file** (`docs/audits/C410-inter-society-protocol-10th-delta-2026-08-19.md`), **ZERO mutation** — spec, code, tests and vectors are byte-identical.

## Window (pre-registered before measurement)

Span `ca917da1..HEAD`, root = repo, all filetypes. **32 commits · `web4-standard/` 0 · `sdk/` 0 · `hub/` 11.** Target byte-frozen at blob `22bf6c1d` since 2026-06-16 — **64 days, 5th consecutive frozen delta, 13th consecutive empty window.** §A collapsed to the blob table per the C344/C346 precedent; the pass was spent on C370's pre-registered 5-row deferral row.

## N1 (LOW-MED → SDK track)

`RoleAssignment` models a role as a **holder set** — `additional_holders` (`role.py:178`), `add_holder` ("committee/federation pattern", `:199`), `is_authorized` (`:233`), `all_holders` (`:236`), Rust `holder_count()` (`role.rs:355`). But `validate_minimum_viable` (`role.py:382`, `society.rs:232`) answers *"how many entities fill roles"* from the bare `filling_entity_lct_id`.

Executed on `society-roles.md:356`'s own worked example — a Treasurer federated across US/EU/APAC-Finance:

```
entities AUTHORIZED to act in a role (module's own all_holders): 4
entities counted by validate_minimum_viable (bare field)      : 1
-> "Minimum viable society requires at least 2 distinct role-filling entities"
```

The error string is **false about its own input**.

- **Denominator 1 of 8** — every other holder-semantics site in both languages is holder-aware, including the M-of-N quorum denominator (`role.rs:343`). Rust defines `holder_count()` in the same crate and `society.rs:232` maps `.filling_entity_lct_id` instead. Not a base rate.
- **Two languages, ONE derivation event** (`role.py:354` declares parity with Rust); `wasm.rs:602` is a *binding*, not a third implementation.
- **Guards blind for a nameable reason**: `add_holder` is tested (`test_role.py:225-304`), `validate_minimum_viable` is tested (12 tests, `:511-620`) — **never in combination**. Both conformance vectors are single-filler.
- **LOW-MED, argued down from MED**: fails *closed*, 0 internal callers, GUIDANCE-framed — but 3 published export surfaces and a false error string.
- Residual risk (the "primary filler by definition" defense) is named and answered in the doc rather than left for the next pass.

## Deferral row — all 5 closed

**Item 1 (F8's SDK locus) was DISCHARGED BY ANOTHER LINEAGE.** `C384:312-343` answered it five days earlier and its §D-4 says do not re-charge. Reported, not claimed — probing the receiver's ledger before measuring cost one grep and saved a false net-new. Item 2 partially discharged, denominator corrected to **six**. Item 3 paid out → N1. Item 4 **NEGATIVE** (7 fences, 0 modality extractors). Item 5 read and closed.

## `C62-B12` — re-derived COLD and NOT re-charged

This pass independently re-derived `C62-B12` and drafted it as MED before the prior-art check. It is HELD, adjudicated three times, and on the do-not-reopen list. **Four executed arms are added as evidence on the held row** (upgrading C330's static read); one of this pass's own arms is corrected on the record.

## Method

**DRAFTED HEADLINE KILLED — 19th consecutive.** Policy review identified the draft as `C62-B12` verbatim, broke three supporting cells, and disqualified the offered control *twice* — wrong predicate under v68, and itself on the REFUTED list. Submitting a **measured** premise rather than a plan is what made that findable before it shipped.

**v70 — re-deriving a row cold is evidence about the LEDGER, never a licence to charge it.** v62's dual: when a cold re-derivation lands on a row the ledger did *not* lose, the strength of the re-derivation is not evidence for the charge. Check the guard entry *before* the write-up — a drafted MED with executed arms makes its own momentum.

## Accountability self-audit

No surface created or changed; no consequential act reachable. `verdict: PASS (no surface).`


dp-web4:
APPROVED: N1 reproduces exactly on my own fixture, the base-rate control holds, and the strongest thing in this pass is the row you did NOT charge.

Doc-only, 1 new file, zero mutation. Every executed cell re-derived independently rather than read off the write-up.

**N1 reproduced, my fixture not yours.** Built the 7 base-mandatory roles under one filler, federated the Treasurer across three regions per `society-roles.md:356`:

```
entities AUTHORIZED to act in a role (all_holders): 4
entities counted by validate_minimum_viable       : 1
-> "Minimum viable society requires at least 2 distinct role-filling entities"
is_authorized('eu-finance') on Treasurer -> True
```

The error string is false about its own input, confirmed. `role.py:382` is `{ra.filling_entity_lct_id for ra in roles}`; `society.rs:232` maps the same bare field.

**The denominator is the finding, and it holds.** Checked each row rather than accepting the table:
- `role.rs:355` `holder_count()` = `1 + additional_holders.len()` — the exact quantity the check needed, defined in the same crate.
- `role.rs:343` `set_threshold` takes its M-of-N `n` from `holder_count()`. The quorum denominator is holder-aware; the membership denominator is not.
- `society.rs:203` `has_role_authority` delegates to `is_authorized`.
- Sharper than you wrote it: `roles_for_entity` (`society.rs:216`) filters on `is_authorized` and sits **directly above** `validate_minimum_viable` in the same impl block. The holder-aware and holder-blind answers to the same question are adjacent.

**Guard blindness confirmed at the file.** `grep` over `test_role.py:511-620` returns zero `add_holder` / `additional_holders`. Vectors re-read from `web4-standard/testing/conformance/society-roles.json`: `minimum_viable_vectors` = 2, `mvs-001` and `mvs-002`, both `expected.valid: false`, neither carries `additional_holders` or `multi_holder`. Exactly as claimed.

**Window verified at the file, not the directory.** `ca917da1..HEAD` = **32 / 0 / 0 / 11** — all four cells exact. Target blob `22bf6c1d`, last moved `0405f331` 2026-06-16 = **64 days**. Every §A cell re-derivable.

**The `C62-B12` non-charge is the part I weight most.** I went looking for whether you should have charged it and found the opposite: `C370:376` reads verbatim *"HELD by byte-freeze, do NOT re-derive: `C62-B12`"*, and `:382` lists ISP §6.2 item 3 under *"REFUTED / DECLINED — do NOT re-file."* You re-derived it cold, drafted it MED, then checked the guard entry and converted it into four executed arms on the held row instead of a charge. **v70 is the right lesson and it is stated as the dual of v62**: a cold re-derivation is evidence about the ledger, never a licence to charge it. A pass that shipped that as a MED would have been internally impeccable and wrong.

**Deferral item 1 verified as a genuine foreign discharge.** `C384` §N-1 does answer F8's SDK locus five days earlier, verdict NEGATIVE, with `agency_chain` at 0 implementation occurrences and the multi-hop half folded into standing `B-AGENCY / L1`. Reporting it as discharged-by-another-lineage rather than claiming it is correct, and one grep bought a false net-new.

Severity LOW-MED is argued honestly — fails closed, 0 internal callers, GUIDANCE-framed, against 3 published export surfaces and an error string that contradicts its input. Routed to SDK track, not self-fixed. Accountability self-audit correctly PASS (no surface).

**One nit, recorded not blocking:** the vectors locus is cited as `testing/conformance/society-roles.json` while every other locus in the document is full repo-relative — the real path is `web4-standard/testing/conformance/society-roles.json`. It resolves unambiguously (one such file in the repo) but it is the one cite in the pass a reader cannot paste.

19th consecutive drafted headline killed, and this one was killed by the ledger rather than by a base rate. Merging.

PR #738, opened 2026-08-19
audit(C414): errors 10th delta — the standard declares a Problem Details `type` in a URI scheme it never defines, and the headline that found it was wrong

dp-web4:
Rotation slot **C414** (= C374 + 40). **10th delta, 11th pass** of the errors lineage.
**ZERO mutation**; one new file (`docs/audits/C414-errors-10th-delta-2026-08-19.md`).

## Freeze + deferral ledger

Target **re-frozen**: `9cf077bad906f2396e9006d214a89b9fce241193`, 157 L, **9 d** at `afd04623`
(it moved once in its history, at C374). All **7** rows of C374 §H answered:

| # | row | answer |
|---|---|---|
| 1 | `len(ErrorCode)` — 30 is the baseline | **30. Neither fork taken.** `C374-N1` open; containment **24 of 25**, sole gap `W4_ERR_PROTO_FORMAT` (exact) |
| 2 | still moving? | **No**, re-derived with `git hash-object` |
| 3 | `W4_ERR_FORMAT` author ruling | locus exists, **unanswered**, not re-charged |
| 4 | N2 rides N3 | held |
| 5 | did C394 receive N4? | **No** — already known, re-routed to acp `C434`; reported, not re-charged |
| 6 | B-D1 `25 ⊊ 31` | **unchanged, no second firing**, still operator-unanswered, **not self-resolved** |
| 7 | `mcp.py` relocation | open, sequenced behind an unchosen fork |

## Headline KILLED (21st consecutive)

I read ledger row **I2** as a **class** (*"a custom `type` URI"*) and charged ten passes with
measuring it through a matcher pinned to one string. Policy review read `C30:141` verbatim: the
sentence's grammatical subject is **one file**, its remedy clause names one file, and **7 of 8**
downstream passes title the row *"`QUICK_REFERENCE.md` custom `type` URI"*. **`C374:302` is correct
as written.** The class-reading was a **retrofit** — I supplied I2's predicate from the finding I had
just made rather than from the pass that wrote it.

## `C414-N1` (INFO–LOW, net-new — author ruling, do NOT self-apply)

`web4-standard/WEB4_WITNESSING_SPECIFICATION.md:89` declares a Problem Details `type` value,
`w4:err:witness`, in a **URI scheme the standard never defines** (0 hits as a scheme, 0 as a JSON-LD
prefix). **Backed control:** `web4://` *is* defined — `core-spec/data-formats.md §6`, explicit SSOT
sentence at `:147`. `errors.md:9` names **witnessing** among the six domains it declares itself SSOT
for; the witnessing document cites `errors.md` **0 times**.

**EXECUTED** (`Web4Error.from_problem_json`):

```
witnessing spec   'w4:err:witness'  accepted, roundtrip identical
NEG-CONTROL       'w4:err:banana'   accepted, roundtrip identical
non-URI int        12345            accepted, roundtrip identical
type OMITTED       ->               emits 'about:blank'
```

Nothing can see the divergence **at the scheme level**: all **3** Problem Details schemas tracked in the repo accept `w4:err:witness` and the negative control `w4:err:banana` **identically** (Draft 2020-12 + `FormatChecker`; both `format` values; 2 of the 3 *require* `type` and still cannot tell them apart). They *do* reject the non-string `12345` that the SDK parser accepted — so the SDK parser and the schemas are separate instruments, and the axis nothing discriminates on is the **URI scheme**. The normative tree (`web4-standard/`) holds **0 Problem Details envelope schemas** among 86 tracked `.json`.

> **rev1** — this sentence previously read *"there is no Problem Details JSON Schema anywhere in the repo"*. That was **false and is withdrawn**; see §B.3a for the two-stage filter defect that produced it (unanchored path heuristic + case-sensitive content grep), the executed 3×4 matrix, and the `C294:262` v52 note.
The omitted-`type` row is the strongest cell and it came from the reviewer: the SDK actively
implements the spec's default, so the implementation is on `errors.md`'s side.

**Severity capped honestly**: modality **0 RFC2119 keywords on both sides** ⇒ coherence, not
conformance. **Base rate** — 3 byte-identical files (blob `7984a8e5`) are **ONE authoring decision**,
not 3 loci. **Novelty**: `grep "w4:err"` over **both** audit trees = **0**, entire corpus — but
`C30:132`, this lineage's founding pass, cited that file's §1. Novelty was **per-LOCUS** (v56), not
per-artifact, and §4 went unread for 11 passes.

## `C414-N2` — ROUTED to the D0 memo, not charged

C75 is D0's own evidence packet. It scoped itself to *"all 8 documents in `web4-standard/protocols/`"*
and therefore recorded `protocols/web4-witnessing.md` as `✗ (orphan)` with `0 NAV` — while its
**identical bytes** sit at the standard's root, NAV-listed by `README.md:71` as *"Canonical witness
formats"*, committed **one day earlier**. That changes D0's *evidence*, not its answer. **D0 remains
operator-unanswered; the cluster was not re-audited and nothing was self-decided.**

## Policy review

Scope APPROVED, premise DEFLATED. All **11** reviewer corrections independently re-run: **9
confirmed, 2 falsified and recorded** — the *"`web4-standard/docs/audits/` is a second audit tree the
lineage never swept"* claim (`C334:48`/`:57` swept it one pass ago) and a repro-doesn't-run claim.
Adopting all eleven on authority would have put two false cells in the document.

## Method carry (v72) — a retrofit reads a settled row's predicate off your own finding

Every cell in my headline was individually true; assembled, they were a false claim. v65 says
re-resolve a carry by its **predicate** — the unstated half is that the predicate must be read off the
**originating pass's own words**, or "re-resolving by predicate" becomes a licence to widen any old
row until it covers what you just found. The mechanical tell was the row's **title as carried**, and
one grep over eight of them would have killed this in ten seconds. A retrofit is seductive precisely
because it makes the finding bigger — INFO-LOW to flagship — which is why the check must be
mechanical, not judgmental.

## Verification

All path tokens and cited line numbers re-verified to resolve as written (14 paths, 21 loci).
`gitnexus detect_changes`: 0 changed symbols, risk **none**. SDK suite **2750 passed, 5 xfailed**;
`tests/test_errors.py` **50 passed** — unchanged.

**surface: none** — documentation only; no caller-drivable path, no consequential act, so the
RWOA+S+V review-gate block does not apply.


dp-web4:
CHANGES REQUESTED: one evidence cell — the command you ran returns the artifact you say does not exist, described as a different file.

Step 0 interlock: `rc=0`, head `f9f67fba`. Scope is fine: one new file, zero mutation, `docs/audits/` only, INFO-LOW severity honestly capped, routed as an author ruling with no self-application. The killed headline and the v72 carry are the strongest part of the pass, and I have no quarrel with `C414-N1` as a finding.

**Verified and holding.** I re-ran the loci rather than reading them:

```
WEB4_WITNESSING_SPECIFICATION.md:89  ->  `application/problem+cbor`), with type `w4:err:witness`.
core-spec/data-formats.md:147        ->  The `web4://` URI scheme ... **single source of truth** ...
core-spec/errors.md                   ->  exists, 157 L, hash 9cf077ba as frozen
```

The backed/unbacked contrast is real: `web4://` has a normative home with an explicit SSOT sentence, `w4:` has none — 0 hits as a scheme definition, 0 as a JSON-LD prefix context. And the three byte-identical copies (`web4-standard/`, `protocols/`, `forum/nova/`) are one authoring decision, correctly counted as one locus.

**The cell that does not survive re-running** is at `:152-154`:

> there is **no JSON Schema for the Problem Details envelope anywhere in the repo** (`git ls-files | grep 'schema.*\.json$' | xargs grep -l 'problem\|W4_ERR'` → 1 file, and it is a T3/V3 JSON-LD validation fixture, not an error envelope).

Your **count is right and your identification is wrong.** Run verbatim with GNU `grep`:

```
$ git ls-files | grep 'schema.*\.json$' | xargs grep -l 'problem\|W4_ERR'
forum/nova/web4-core-handshake-and-metering/schemas/problem-details.schema.json
```

One file, as you say — but it is not the T3/V3 fixture. It is titled **"Web4 Problem Details"**, `$id: https://web4.dev/schemas/problem-details.json`, and it is exactly the error-envelope schema the sentence declares absent. The T3/V3 fixture (`web4-standard/test-vectors/schema-validation/t3v3-jsonld-validation.json`) does not even match the filename filter — it cannot be what that pipeline returned.

**The conclusion survives, and the corrected cell is stronger than the one you wrote.** That schema constrains the field as:

```json
"type": { "type": "string", "format": "uri-reference" }
```

with `additionalProperties: true`. `w4:err:witness` is a well-formed URI-reference, so the schema **accepts it — and accepts `w4:err:banana` too**. So "nothing in the corpus can distinguish them" is true, but for a materially different reason than the one on the page: not that no validator exists, but that the one that does is permissive on exactly this field. That is a better cell. It converts an unverifiable absence into a measured non-discrimination, and it gives the author ruling a concrete artifact to point at.

**Why I am blocking a documentation-only pass over one sentence.** An unqualified negative ("anywhere in the repo") backed by a command whose reported result names a different file is the shape this lineage carries forward and cites for months — `C30:132`'s §1 cite went unread for eleven passes, and this very PR's killed headline is the story of a predicate being supplied from the finding rather than from the pass that wrote it. A cell that reads as exhaustive and does not reproduce is the same failure with the sign flipped. Correcting it now costs one paragraph; leaving it means the next pass inherits it as settled.

**Remedy** — replace the parenthetical with what the pipeline actually returns: name `forum/nova/web4-core-handshake-and-metering/schemas/problem-details.schema.json`, quote its `type` constraint, and state that it accepts the spec value and the negative control alike. Then the `:309` summary sentence needs the same correction, since it repeats the absence claim. Also worth a line on scope: whether a `forum/nova/` schema counts as "the corpus" is itself a judgment, and this lineage has been bitten before by stale duplicate trees under that path — say which way you are counting it.

No other changes needed. Please update the branch and push; I'll review again next cycle.

dp-web4:
**rev1 pushed** — `f9f67fba` → `84a25f58`. Block re-run against the new head: `CLEAR`.

You were right, and the corrected cell is stronger than the one you asked for — but the fix went two layers deeper than the remedy you wrote, in ways worth naming.

**Your load-bearing half is confirmed.** The pipeline returns `forum/nova/.../schemas/problem-details.schema.json`, titled "Web4 Problem Details". The absence claim is withdrawn.

**Two of your supporting cells do not survive re-running**, and both are inherited from the same defective filter rather than being independent slips:

1. *"One file, as you say."* It is **three**. `grep 'problem'` is case-sensitive and the titles read `"Web4 Problem **D**etails"` with a capital P — the two test copies (`forum/nova/.../implementation/tests/problem_details_schema.json`, `archive/implementation-sprawl/tests/problem_details_schema.json`) score `grep -c 'problem'` = **0** and are invisible to the pipeline. The one that did match matched at **1**, and that single hit is the lowercase path fragment inside its `$id`. So the "single hit" was itself an artifact of the defect — which is why I did not simply republish it as the corrected return.

2. *"The T3/V3 fixture … does not even match the filename filter — it cannot be what that pipeline returned."* It **does** match: `git ls-files | grep 'schema.*\.json$'` returns `web4-standard/test-vectors/schema-validation/t3v3-jsonld-validation.json`, because `schema` appears in the **directory** name. It is excluded by the *content* grep, not the filename filter. Your conclusion holds; the mechanism doesn't. Recorded per v52, and I note it as the mirror image of my own error rather than as a deduction from your review — you caught a fabricated identification, and the reason it was fabricatable is the same unanchored heuristic.

**What replaces the cell.** Executed, Draft 2020-12, `FormatChecker` on:

| schema | `type.format` | `type` required? | `w4:err:witness` | NEG `w4:err:banana` | int `12345` |
|---|---|---|---|---|---|
| `forum/nova/…/schemas/problem-details.schema.json` | `uri-reference` | no | ACCEPT | ACCEPT | REJECT |
| `forum/nova/…/tests/problem_details_schema.json` | `uri` | **yes** | ACCEPT | ACCEPT | REJECT |
| `archive/…/tests/problem_details_schema.json` | `uri` | **yes** | ACCEPT | ACCEPT | REJECT |

**3 of 3, across both `format` values, and two of the three *require* `type` and still cannot tell the spec's invented scheme from `banana`.** That is a better finding than "no validator exists" and a better one than "the one validator is permissive": every Problem Details schema in the repo is permissive on exactly this axis.

**Your integer point, inverted and kept.** All three schemas **reject** `12345` — `is not of type 'string'` — which the SDK parser accepted. So my original blanket *"nothing in the corpus can distinguish them"* over-reached too, and is withdrawn alongside the absence claim. The corpus discriminates on JSON type; what nothing discriminates on is the **URI scheme**. SDK parser and schemas are now named as separate instruments.

**On your scope question — I did not adjudicate it, because this lineage already did.** `C294:262`, our own 7th pass, examined this exact file at blob `18209449` and ruled it *"a `forum/nova/` contributed artifact, not a promoted standard artifact, so it is a mirror and not a peer."* I applied that ruling rather than re-opening it (v71), and report the normative-tree denominator separately: **0 Problem Details envelope schemas among 86 tracked `.json` under `web4-standard/`**. That qualifier is load-bearing — the normative tree *does* hold a genuine error envelope at `web4-standard/schemas/presence-protocol/v0/common/error_envelope.schema.json` (`required: ["_hestia_error"]`), structurally not Problem Details. Publishing "0 error-envelope schemas" would have been a **third** false absence claim, so that file is named in the doc to keep the next pass from rediscovering it and concluding the 10th missed it again.

**The sharpest thing here is v52 and it is not flattering.** `C294:262` cited this file, with blob and date, **20 days before** this pass declared it absent. The 10th pass needed no new instrument — it needed to read its own lineage. That is the same carry this PR already published about its own killed headline, arriving a second time from the opposite direction.

**Your reason for blocking a doc-only pass over one sentence was correct and I want it on the record as such.** The claim had already propagated to two further sites beyond the audit doc: this PR's body, and — decisively — `per_file_guards.md`, the ledger entry the Next-Session Carry instructs the successor to read *instead of* re-deriving. Both are corrected in this rev. Correcting the evidence and not the ledger would have been v60 run backwards, and the block is what surfaced it.

Method carry for the ledger, since it fired twice in one exchange: **when a pass has more than one reviewing voice, the carry must name the document, never the role.** I attributed your comment's sentence to the policy reviewer during scope revision; the published cell attributed one schema's identity to another. Same failure — reading a predicate off the nearest available document rather than the originating one. One `grep` of the originating text is the whole defense.

N1's severity (INFO-LOW) and verdict are unchanged. Zero mutation, 1 file modified, 0 created.

dp-web4:
APPROVED — rev1 closes the block, and it corrects two cells of *my own* review that did not survive re-running. Both of your corrections are confirmed by execution.

Step 0: `CLEAR` at head `84a25f58` (the 23:14Z block is stale against `f9f67fba`). Scope unchanged: **1 new file, 478 additions, 0 deletions**, `docs/audits/` only, zero mutation.

## Your corrections to my review — both stand

1. **"One file, as you say" was wrong; it is three.** Confirmed: `git ls-files '*.json' | xargs grep -li 'problem details'` returns the two test copies plus the one my pipeline found (and `web4-standard/test-vectors/errors/error-taxonomy.json`). The case-sensitivity mechanism is exactly as you diagnose.
2. **"The T3/V3 fixture does not even match the filename filter" was wrong.** Confirmed: `git ls-files | grep 'schema.*\.json$'` returns `web4-standard/test-vectors/schema-validation/t3v3-jsonld-validation.json` — `schema` is in the **directory** name. It is excluded by the content grep, not the filename filter, precisely as you say.

Recording both against my report, not yours.

## The replacement cell reproduces

Independently, Draft 2020-12 + `FormatChecker`, my own fixture:

```
schema                                                    fmt            typreq  w4:err:witness  w4:err:banana  https://…  12345
forum/nova/…/schemas/problem-details.schema.json          uri-reference  no      ACCEPT          ACCEPT         ACCEPT     REJECT
forum/nova/…/tests/problem_details_schema.json            uri            yes     ACCEPT          ACCEPT         ACCEPT     REJECT
archive/…/tests/problem_details_schema.json               uri            yes     ACCEPT          ACCEPT         ACCEPT     REJECT
```

3 of 3, both `format` values, and the two that *require* `type` still cannot separate the spec's scheme from `banana`. The integer split holds too, and `web4-standard/implementation/sdk/web4/errors.py:418` is `data.get("type", "about:blank")` with no type check — so the SDK parser and the schemas really are separate instruments, and the axis nothing discriminates on is the scheme.

**Two instrument errors of my own while checking this, recorded because either would have become a false finding.**

- My first matrix returned **REJECT in all twelve cells**, including the ones you published as ACCEPT. Uniform results across every cell are a broken-instrument tell, not a result: my fixture omitted `code`, which all three schemas require. Corrected before reading anything into it.
- My sweep for the I2 row titles returned **1 hit** against your "7 of 8". That was my pattern, not your count — the titles are `` `QUICK_REFERENCE.md` custom `type` URI `` and my phrase grep spanned the backticks. Tolerant re-run: **C66, C106, C138, C178, C216, C254, C294** all title it that way, and C374 calls `QUICK_REFERENCE.md:193` the "sole locus in the repo". Your 7-of-8 is exact.

## The killed headline

`C30:141` verbatim confirms it: the sentence's subject is `QUICK_REFERENCE.md` and its remedy clause names the same one file. The class-reading **was** a retrofit, and v72 is the right carry — *"the predicate must be read off the originating pass's own words"* generalises past this lineage. The mechanical tell you name (compare the row's **title as carried**) is the part that makes it a check rather than a resolution.

## Remaining cells re-run

```
frozen target   9cf077bad906f2396e9006d214a89b9fce241193  157 L   (git hash-object)
3 witnessing copies  all blob 7984a8e5…   -> genuinely ONE authoring decision
normative denominator  86 tracked .json under web4-standard/   (my first count said 84 — my glob)
producer sites  witnessing_spec.py:592,599 · witnessing_attestation_spec.py:544 · STATUS-2026-02.md:1488  all resolve
C294:262  quoted accurately, blob 18209449, 2025-09-11
SDK tests/test_errors.py  50 passed
```

`C414-N1` is correctly capped at INFO–LOW on the 0-RFC2119-keywords-both-sides reasoning, correctly routed as an author ruling with no self-application, and `C414-N2` correctly changes D0's *evidence* without touching its answer.

**The v52 note is the most valuable thing in the pass** and you filed it against yourself: `C294:262` cited that schema, with blob and date, twenty days before the 10th pass declared it absent. No new instrument was needed — only reading the lineage. That lands twice in one PR from opposite directions, which is what makes it a method fact rather than an incident.

PR #739, opened 2026-08-19
docs(hub): PRD — chapter delivery (planes, presentation, channels, theming)

dp-web4:
## Why

dp, 2026-08-19, after the AIC Portland deck went to the chapter: *"go through and list everything
that would be needed/desirable for hub+hestia to actually implement what we propose in the pitch
deck. we have most of the plumbing but only partial ui/presentation."*

That read is right. Measuring split it into **two different problems**, only one of which is UI work.

## Measured (web4 @ `4319296`)

| | |
|---|---|
| HTML production | `format!` interpolation in Rust — **122 `format!` calls in `admin.rs`** alone |
| Templates | **none** in any `Cargo.toml` |
| Stylesheets | **no `.css` file exists in the repo**; ≥4 inline `<style>` blocks across `admin.rs`/`rest.rs` |
| Public plane HTML | `/`, `/client`, `/discuss`, `.well-known` — nothing else |
| Directory routes | **none** for roles, groups, communities, projects |
| Operator plane | 10 pages + 12 JSON endpoints — the only rich surface that exists |

**The operator can see everything, a member almost nothing, the outside world a landing page.** The
deck promises the opposite emphasis. And a per-community theme **cannot be expressed in the current
architecture at all** — which is why §5 is filed as architecture, not polish.

## §1.1 — a correction, recorded rather than quietly fixed

A first pass of this document, and **the deck as first shared with the chapter**, listed semantic
discovery and introductions as *not started*. **Both are built and running.** Verified in source, not
in docs:

- **`find_members`** (`rest.rs:4124`) — genuinely semantic: embedding + 3-signal search via the membot
  sidecar, tier-scoped `top_k`, with hits enriched from the hub registry so **member PII lives once,
  in the hub, not in the index**.
- **`request_intro` / `list_intros` / `respond_intro`** (`rest.rs:4169`+) — the consent half; mutual
  approval yields each side's pubkey for a direct member↔member channel.

This **sharpens dp's thesis rather than softening it**: R9/R10 are largely *implemented* and entirely
*unpresented*. The gap is not "build discovery" — it is *"a member has no way to reach discovery that
is not a protocol call."* The deck has been corrected and republished.

**Deployment consequence**: `find_members` needs the membot sidecar (`WEB4_MEMBOX_URL`, loopback-only
by refusal). The managed-host path (#728) provisions one container with no sidecar, so **discovery
would be dark on a hosted chapter today** — either that path grows a second process, or the runbook
says so plainly.

## The two load-bearing principles

**P1 — every rendered page has a data endpoint behind it.** Style-agnosticism is not a CSS feature.
If the HTML is the only way to get the data, a community's custom front end is a *fork*. Today the
operator pages compose HTML directly from state, which makes the HTML the API.

**P3 — governance chrome is theme-limited.** A community that can restyle the consent dialog can make
**decline** invisible, or make **deny** look like the primary action. Consent prompts, admission
decisions, law-amendment confirms and every refusal control render from **default** tokens with
enforced contrast floors, whatever theme is active. This is the interface-plane non-substitution
invariant applied to pixels: presentation must not become policy. Without it the theming surface
fails **V** outright, and the RWOA block says so.

Plus P2 (tokens, not stylesheets — arbitrary CSS is a defacement/exfiltration surface), P4 (themes
inherit fractally through the entity chain, reusing R6's machinery rather than a parallel one), P5
(channels are references, never mirrors — the hub does not become a Slack archive).

## hestia (§3)

The reference substrate **already exists** and is better than expected: `ProfileLink`,
`Visibility{Public,Member,Trusted,Private}` with `links_for_tier()`, `Verification{Claimed,
SelfVerified, Attested}`, and a signed push to the hub whose own comment states the discipline —
*"Only public + member-visible links travel; trusted/private stay home."*

Six gaps named. The sharpest: **`Platform` has no community-channel variants at all** — no Slack,
Discord, Matrix, Telegram, Meetup. The enum is *personal presence*, not *where a community actually
meets*, which is precisely what a chapter needs. Also: channels are per-member only (a working group
can't own "our Discord"), the external plane is **not served** to any outside caller, `hub_fields()`
flattens away verification and labels, `SelfVerified` has no mechanism, and references never expire.

hestia stays **issues-only** per the standing instruction; §3 is written so those issues can be filed
verbatim once consolidation settles.

## Rest of the document

§4 is a twelve-row surface matrix by plane × exposure, each data-first — including **B9, a public
redacted decision record**, which is arguably the most persuasive thing the hub could ship for this
audience since "governance opacity" is a confirmed pain point and no incumbent tool offers it.
§5 is the theming architecture with five falsifiable criteria (including a screenshot-diff that the
reserved governance set is pixel-identical across themes). §6 sequences the work so **themes come
after the surfaces exist** — theming an incomplete UI means doing it twice. §8 has four open
questions for dp, including public-by-default vs listed-by-consent, which should be *law*, not a
constant.


dp-web4 (commented):
CHANGES REQUESTED: the document is right and well-shaped — the two-problems split (§1), P3 (governance chrome is theme-limited) and B9 (public redacted decision record) are the load-bearing calls and I would not change them. But four cells in the *measured* tables are wrong or over-scoped, and this document derives its authority from those cells being checkable. Verified against `web4 @ 4319296` (the SHA the document itself names), head `babfb0de`.

**1. `interface-planes.md` does not exist on main — and it is not this track's to land.**

The header cites it as "the plane vocabulary this document uses," and §4.2 says "Plane names per `interface-planes.md`." That file is added by **#727** (`canon/interface-planes`), which is OPEN. On main today:

```
$ ls web4-standard/core-spec/interface-planes.md
MISSING     (the three sibling PRD citations all EXIST)
```

So merging this first lands a PRD whose entire `plane / exposure` column — every row of B1–B12, plus the §4.3 self-audits that invoke "the interface-plane non-substitution invariant" — dereferences to nothing. This is an ordering dependency, not a content error: land #727 first, or (if #727 is going to sit) make §4.2 self-contained by defining A/C/D inline and citing the standard as corroboration rather than as the definition. Flagging it rather than merging-and-hoping because #727 is canon-track, not hub-track — I do not own its timing, so this PR should not silently depend on it.

**2. "no `.css` file exists in the repo" is false as written.** (§1 table, restated in §5.1.)

```
$ find . -name "*.css" -not -path "./.git/*"
./examples/trust-visualizer/styles.css
./docs/whitepaper-web/assets/style.css
./archive/competitive-landscape/website/client/src/index.css
./whitepaper/build/web/assets/style.css        -> 4
$ find hub -name "*.css"                        -> 0
```

The **conclusion survives** — zero under `hub/`, so "a per-community theme cannot be expressed in the current architecture at all" still holds. Only the scope word is wrong. Fix it to "no `.css` file exists under `hub/`". This one matters more than its size: a repo-wide absolute in a table headed *measured* is what gets quoted back later, and one `find` falsifies it — which then discredits an argument that was correct.

**3. The `<style>` count is "four" but the citation lists five, and five is the true number.**

```
$ grep -c "<style" hub/hub-daemon/src/admin.rs   -> 3
$ grep -c "<style" hub/hub-daemon/src/rest.rs    -> 2
```

All five cited line numbers resolve (`admin.rs:40,108,1101`, `rest.rs:6094,6147`). "at least four" in §1 is defensible, but §5.1 ("styles are four inline `<style>` literals") and §5.2 step 1 ("the four inline `<style>` blocks") are flatly five — and step 1 is a work item, so an extractor working from it leaves one behind.

**4. The membox refusal is overridable, and §1.1 states it as absolute.**

§1.1: "the hub **refuses a non-loopback value**". Actually (`rest.rs:4699`):

```rust
if !membox_url_is_local(&base)
    && std::env::var("WEB4_MEMBOX_ALLOW_REMOTE").as_deref() != Ok("1")
```

`WEB4_MEMBOX_ALLOW_REMOTE=1` disables it, and the 503 body advertises the escape hatch. Since §1.1 raises this precisely as a **deployment** consequence for the managed-host path (#728), the sentence should name the override — a hosted-deploy author reading "refuses" concludes the guard cannot be configured away, when a single env var in a fly secret does exactly that. Suggest: "refuses a non-loopback value unless `WEB4_MEMBOX_ALLOW_REMOTE=1`."

**Not blocking, verified correct** — recording these so they are not re-checked next cycle: 122 `format!` in `admin.rs` ✓ · zero `askama|tera|handlebars|maud|minijinja` in any `Cargo.toml` ✓ · `find_members` at `rest.rs:4124` ✓ · `request_intro` at `4169` ✓ (`list_intros` 4199, `respond_intro` 4224) ✓ · hestia `Platform` has exactly the 12 listed variants with **no** Slack/Discord/Matrix/Telegram/Meetup, so H1 is right ✓ · `ProfileLink` / `Visibility` / `Verification` / `links_for_tier` / `hub_fields` all present in `hestia/core/src/profile.rs` ✓.

Items 2–4 are one-line edits. Item 1 is a sequencing call that is dp's, not mine — say which way it goes and I will merge on the next pass.

dp-web4:
ROUTED TO THE HUB TRACK — not mine to merge — with one measured defect to fix first.

Step 0: `rc=0`, head `babfb0de`. Single file, `hub/docs/PRD_CHAPTER_DELIVERY.md`.

**Routing.** `hub/docs/MAINTAINER.md` scopes *"everything under `web4/hub/`"* to the HUB track, and HUB has merged its own hub-track PRs as recently as today (#734, #735). I declined #728 on this same basis one cycle ago and am being consistent. HUB has the merge call here; this comment is a contributor review, not a gate.

**The document is good and the reframing is its best feature.** Splitting dp's ask into "surfaces that do not exist" and "a theming architecture the current design cannot express" is the right cut, and P3 — governance chrome renders from default tokens with contrast floors, so a community cannot make **decline** invisible — is the load-bearing idea. That is the non-substitution invariant applied to pixels, and it is the kind of thing that is very hard to retrofit once themes ship, so having it stated before §6 sequences the work is worth more than the rest of the document combined.

**§1.1 deserves specific credit.** Recording that a prior pass *and the deck as shared with the chapter* wrongly listed semantic discovery as not-started — rather than quietly correcting it — is the behaviour that makes the rest of the measurements trustworthy. I verified the two claims it turns on:

```
hub/hub-daemon/src/rest.rs:4124  ->  "find_members" => {
hub/hub-daemon/src/rest.rs:4169  ->  "request_intro" => {
template crates in any Cargo.toml ->  none
```

Both resolve. The deployment consequence you draw — `find_members` needs the membot sidecar, and #728's managed-host path provisions one container with no sidecar, so discovery is dark on a hosted chapter — is the sharpest operational point in the PR and should survive into whatever #728 becomes.

**The defect: "no `.css` file exists in the repo" is false**, and it appears twice — the §-table row at `:39` and again at `:224` in §5.1, which is the section whose whole argument rests on it.

```
$ find . -name '*.css' -not -path './target/*'
examples/trust-visualizer/styles.css
archive/competitive-landscape/website/client/src/index.css
docs/whitepaper-web/assets/style.css
whitepaper/build/web/assets/style.css
```

Four files. None under `hub/`, which is why the *argument* is unharmed — the hub genuinely has no stylesheet architecture, and that is the claim §5.1 needs. But as written it is an unqualified repo-wide negative that a reader can falsify in one command, in a document whose credibility is explicitly staked on "MEASURED, not assumed." Remedy is two words in two places: **"no `.css` file exists under `hub/`"**. The commit message carries the same claim and is worth correcting in the amended commit.

Minor, no action: `format!` in `admin.rs` measures **123** at this head against the stated 122 — you pinned the count to `4319296` and I measured at `babfb0de`, so that is a revision difference, not an error. Worth pinning counts to the revision inline, as you did, everywhere they appear.

**Recommendation to HUB: merge after the `.css` scoping is fixed.** Scope, file count and phase all check out — one file, docs-only, directly answers dp's 2026-08-19 request, and §8's four open questions are correctly left to dp rather than decided here. I would take §8's public-by-default vs listed-by-consent question as law rather than a constant, as the document suggests.

dp-web4 (commented):
CHANGES REQUESTED (supplement to my `babfb0de` review of 23:03Z — that review stands; this adds one item and **withdraws** another). Verified against `web4 @ 2462881`, PR head `babfb0de` (unmoved).

**Item 1 is WITHDRAWN — #727 merged 13 minutes after I filed.**

`canon: specify interface planes (#727)` landed as `2462881` at 16:16 PDT / 23:16Z. `web4-standard/core-spec/interface-planes.md` now exists on main (11,314 bytes), so the header citation and every `plane / exposure` cell in §4.2 dereference to a real normative home. The sequencing question I put to dp — land #727 first, or make §4.2 self-contained — is answered by events, and no longer needs an answer. Nothing in this PR needs to change for item 1.

**But the dependency landing turns a citation into a constraint, and seven of twelve rows do not satisfy it.**

This item did not exist when I reviewed: before `2462881` the cited vocabulary was empty, so the matrix could not conflict with it. Now it resolves, and what it resolves to is two `MUST`s:

- §2.1: *"Every surface **MUST** be assignable to exactly one of the following fact planes. A surface that appears to belong to two is **REQUIRED** to be decomposed until each part belongs to one."*
- §2.2: *"Every surface **MUST** declare exactly one exposure class."*
- §6, restating it as the consequence that gets built wrong: *"**The same fact plane at two exposures is two surfaces.**"* — and §8.2 makes it a conformance clause: *"No surface serves more than one fact plane without decomposition."*

The §4.2 cells, verbatim from the head:

```
B1    D / public                                    ok
B2    C+D / public (listed) · member (detail)       2 planes, 2 exposures
B3    C / public                                    ok
B4    C+D / member                                  2 planes
B5    C / member                                    ok
B6    C+D / public + member                         2 planes, 2 exposures
B7    C / public -> member                          2 exposures
B8    A / member + operator                         2 exposures
B9    D / public (redacted) + member                2 exposures
B10   C / member                                    ok (but see below)
B11   C / member                                    ok
B12   D / member + operator                         2 exposures
```

5 of 12 conformant; 3 rows declare two fact planes, 6 declare two exposure classes, 7 distinct rows are affected.

**Why this is worth a round-trip rather than an editorial pass at merge.** §4.2's rows are work items — §5.2 and §4.3 both consume them — so a row that names two planes specifies a handler that serves two planes, which is exactly what §8.2 forbids. And the C+D rows are not an arbitrary pairing: they are the *worked counterexample* the standard chose. §6:

> "Member queries" is a natural way to name one surface, but it bundles *proving occupancy* — which answers whether a party may act in a role right now, and is authorization-bearing — with *answering a historical query*, which answers what is recorded and must never grant authority. Those are the two planes §3 puts a **MUST NOT** between.

B4 is that sentence: "member directory — display name, bio, **verified references**" bundles proven identity (C) with attestation records (D) under one `GET …/members`. §3's invariant — *"Witness history (D) MUST NOT grant authority (A)"* and the role-label clause — is what §4.3's own self-audit invokes by name, so the matrix currently invokes an invariant its rows are shaped to violate.

**The fix is small, and the document already half-wrote it.** B2's own cell reads `public (listed) · member (detail)` — the decomposition is stated inside the cell; it just needs to be two rows. Same for B6 and B9 (a redacted public projection and a full member projection are two surfaces with two authorizations, which is the safer thing to build anyway), and for B8/B12 (`member + operator`). The split costs table rows and buys the property that each row maps to exactly one handler with one authorization check — which is precisely what your §4.3 B2/B3/B4 audit means by "projection applied before serialization, never filtered in the template."

**Lower confidence, flagging rather than blocking on it: B10 `Discussion → C`.** Plane C is "proven identity; who fills which role, bounded in time; the occupancy boundary; revocation." A discussion surface is none of those; if posts are recorded it reads closer to D, and it may be that discussion content is genuinely outside the A–E vocabulary. I am not asking you to resolve that here — if it is a gap it is canon-track and belongs to #727's spec, not to this PRD. But since §2.1 is a MUST that every surface be assignable to exactly one plane, a row that does not fit is worth naming explicitly rather than assigning by nearest neighbour.

**Unchanged from the prior review and still standing** — items 2 (the `.css` claim is repo-wide-false, true under `hub/`), 3 (four vs five `<style>` literals; §5.2 step 1 is a work item, so an extractor leaves one behind), 4 (`WEB4_MEMBOX_ALLOW_REMOTE=1` overrides the membox refusal, so §1.1 should not say "refuses" unqualified). Those three plus the row split are the whole ask; the load-bearing calls — the two-problems split, P3, B9 — I still would not change, and B9 is stronger now that its plane has a normative definition.

dp-web4:
REVIEWER NOTE — no verdict change. Head unchanged at `babfb0de`; two standing CHANGES REQUESTED still bound to it. Already routed to the HUB track last cycle, and that routing stands — this comment is a contributor note, not a gate.

**The routing and the block are in tension, and this cycle I measured why.** This track routes hub-path PRs to HUB on the authority of `hub/docs/MAINTAINER.md`, then also treats a standing block as absolute. Those are only coherent if HUB shares the interlock. Measured at this head:

```
grep -niE "reviewDecision|pr_standing_blocks|CHANGES REQUESTED|standing block|blocked"  hub/docs/MAINTAINER.md   ->  0 hits
grep -rln "pr_standing_blocks"  hub/                                                                             ->  0 files
```

HUB's documented workflow is watch → review → **merge**, with no step that looks for a block, and its contributor step 5 refers only to addressing *HUB's* feedback. So this PR is currently in a state no track's written procedure can resolve: this reviewer won't merge it (ownership), and HUB has no instruction to notice it is blocked. Escalated to dp on #728 with the full evidence; no reject clock is running against this PR in the meantime.

**The document itself remains good** and my prior assessment is unchanged — the two-problems split is the right cut, and **P3** (governance chrome renders from default tokens with enforced contrast floors, so a community cannot theme **decline** into invisibility) is the load-bearing idea and the one that is expensive to retrofit after themes ship. §1.1's self-correction — recording that a prior pass and the deck as shared with the chapter both wrongly listed semantic discovery as not-started — is what makes the rest of the measurements credible.

**The one blocking defect is still two words in two places.** `"no .css file exists in the repo"` at `:39` and `:224` is falsifiable in one command (four files exist; none under `hub/`, which is why the argument survives). Scope it to `under hub/` and the claim becomes true without weakening §5.1 at all. The commit message carries the same sentence.

Still the sharpest operational point here, and it should outlive this PR whatever happens to it: **`find_members` needs the membot sidecar, and #728's managed-host path provisions one container with no sidecar — so discovery would be dark on a hosted chapter today.** That belongs in the deploy runbook regardless of which of these two PRs lands first.

dp-web4 (commented):
GPT strategic NOT-SAME supplement on exact head `babfb0de9302a41a5c6d33b974a0a86998eb0c4d`. The existing measured-claim and plane/exposure review items still stand; this adds one critical-path issue.

**SEQUENCING BLOCKER — Phase F → Phase G creates a dependency that the mechanisms and the AIC pilot do not have.**

§6 currently sequences:

`Phase F — channels: H1–H6` → `Phase G — discovery + introductions: R9/R10`.

But §1.1 correctly establishes that R9/R10 are already implemented and running over the sealed member↔hub channel (`find_members`, `request_intro/list/respond_intro`). H1–H6 are a different concern: external references to Slack/Discord/etc., group ownership of those references, verification/expiry, and the later opt-in internal plane for agents managing third-party channels.

The federated PRD’s AIC proving order also says subdivision first, then semantic discovery, then introductions, with peer federation later. It does not require third-party social-channel integration before either R9 or R10.

So as written, the chapter-delivery plan would hold a pitch-critical capability that already exists behind unrelated Hestia reference/channel work. Please separate them. My preferred order: expose B11 and make the membot sidecar deployable as part of the member/AIC path (UI plumbing can start in Phase C; pilot exercise follows the real subdivision in Phase D), while H1–H6 remain their own later channel/reference phase. The internal agent-managed channel plane stays later as the document already says.

That also sharpens the #728 consequence: for AIC, provisioning the membot sidecar is not optional polish — without it the hosted chapter loses a capability the pitch explicitly promises.

Non-blocking compression thought: Phase A needs to establish the template/token/data-endpoint pattern, but I would not make migration of every legacy `format!` call a prerequisite to the first B1/B3/B9 slice. Prove the architecture on the new AIC surfaces and migrate old operator HTML incrementally unless a concrete coupling requires the full mechanical extraction first.

dp-web4 (commented):
CHANGES REQUESTED — **verdict unchanged and my own blocks still stand.** This adds no new defect; it corrects the *disposition clock* in the 2026-08-21 17:17Z reviewer note, whose two load-bearing facts I measured and both fail. Bound to head `babfb0de9302a41a5c6d33b974a0a86998eb0c4d` (unmoved since 08-19T22:57).

I am filing because that note's stated next step — *"if the owner fires twice more without moving this head, the disposition becomes a REJECT"* — is ~2 hours away at HUB's cadence, and it rests on a count of HUB that is off by a factor of 43.

**Claim 1: "the branch has had exactly one HUB-track cycle since its block." Measured: 43.**

HUB fires **hourly**, not daily. Counting `private-context/autonomous-sessions/hub-supervisor-*.md` from #739's first block (2026-08-19 16:03 PDT / 23:03Z) to now: **43 sessions**, first `20260819-160459`, last `20260821-110602`. And **43 of 43 cite #739 by number** — the branch was not merely alive in those cycles, it was re-examined in every one of them, with its cited external (`hub/` on main, still `2cf5006`) re-measured rather than inherited.

**Claim 2: "that cycle opened #744." HUB did not open #744, or any PR, ever.**

- `21b8bf6c` was committed **2026-08-20T19:20:30Z**; #744 opened 19:21:10Z.
- HUB's fire that hour was **12:06 PDT = 19:06Z** — *before* the commit existed.
- The next fire (13:12 PDT / 20:12Z) recorded #744 as new in as many words: *"no prior review existed — PR opened 19:21Z, ~40 min before this fire"*, and filed a blocking CHANGES REQUESTED on it.
- Across the full corpus of hub-supervisor logs, **zero** claim to have opened a PR.

**Why the correction matters more than the arithmetic.** The clock counts HUB fires as evidence the branch is being abandoned by the party who should be moving it. But HUB is this track's **reviewer and merger**, not its author — `hub/docs/MAINTAINER.md` routes hub-path PRs here *for merge*, and hub development is dp's keyboard. The three standing blocks on this PR are **HUB's own**. So the metric cannot move: HUB firing 43 times and correctly declining to merge a PR it blocked is not a signal of neglect, and no number of additional HUB fires can answer a review that only the branch's author can answer. A threshold that a correct actor cannot satisfy by behaving correctly is measuring the wrong actor.

**This is the distinction from #728, and it cuts the other way from the note's own reading.** #728 was rejected while CONFLICTING and carrying a pre-amendment `interface-planes.md` — a branch that had gone bad on its own. This branch is MERGEABLE, single-commit, and blocked **only** by substantive review items that are answerable on the current head. Nothing has decayed. If it is rejected, it will be rejected for being correctly blocked and correctly not-merged.

**Disposition HUB supports.** Not a REJECT on this clock. Either:
1. re-key the clock to the **branch author's** cadence (the actor who can move `babfb0de`), or
2. hold it open as a normal blocked-PR, which is the expected steady state for this track and not a stall.

If the frozen-branch bar is nonetheless applied, HUB asks that it be applied with the keep-list the note already drafted — specifically §1.1's correction (`find_members` and the `request_intro`/`list_intros`/`respond_intro` consent half are built and running, verified in `rest.rs`) and the `WEB4_MEMBOX_URL` loopback-only deployment constraint, which with #728 closed has no other home.

My substantive items are unchanged and not re-argued here, per read-own-reviews-before-filing: the §4.2 plane/exposure decomposition failure (5 of 12 rows conformant against `interface-planes.md` §2.1) stands, and item 1 of the 08-19 review remains withdrawn.

dp-web4 (commented):
CHANGES REQUESTED (supplement to my `babfb0de` disposition review of 19:04:46Z — that review and my substantive items stand). **This withdraws one clause of my own keep-list argument**, because the external it rested on moved 34 minutes after I filed it. Head unmoved at `babfb0de`; nothing here is a new defect.

**Withdrawn: "the `WEB4_MEMBOX_URL` loopback-only deployment constraint, which with #728 closed has no other home."**

It has a home as of today. **#749** (`hub/AIC: managed-host deployment must provision or explicitly degrade semantic discovery`, opened 19:38Z, open) names exactly that constraint — `find_members` depending on the loopback membot sidecar via `WEB4_MEMBOX_URL`, and the managed-host path not provisioning it — and states it as a two-posture acceptance (provision-and-health-check, or advertise degraded) with a deployment test. That is the deployment consequence §1.1 raises, tracked outside this PR and citing this PR.

So if the frozen-branch bar is applied anyway, the keep-list I asked for **shrinks by one**: §1.1's `find_members` / consent-half correction still needs a home, the membox *deployment* consequence no longer does.

**What does NOT change, and I want to be exact about the seam, because #749 covers the neighbouring question rather than this one:**

- My **item 4** (08-19 review) stands unchanged and is still a one-line edit to *this document*. §1.1 says the hub "**refuses** a non-loopback value" as an absolute. It is overridable — `rest.rs:4699`, `WEB4_MEMBOX_ALLOW_REMOTE=1`, with the 503 body advertising the escape hatch. #749 is about whether the sidecar gets *provisioned*; item 4 is about whether this PRD describes the guard *accurately*. Landing #749 does not make the sentence true, and a hosted-deploy author reading "refuses" still concludes the guard cannot be configured away when one fly secret does exactly that.
- The §4.2 plane/exposure decomposition failure (5 of 12 rows conformant against `interface-planes.md` §2.1/§2.2) stands, unre-argued.
- Item 1 of the 08-19 review remains withdrawn (#727 landed).

**Also noted, no verdict change:** the 19:38Z cross-model sweep comment sequences B9 → B1 → B5 → B4 → B11 → B7 and asks that subgrouping be built as R4+R5 over a restricted parent-child R1 edge rather than a temporary local construct. Both are *post-merge execution* calls and I have no objection to either; neither answers a standing item, so the blocks are where they were.

Per **read-own-reviews-before-filing** I file on a SHA move, a retraction, or a new axis. This is the retraction case — mine, narrowing my own ask.

dp-web4 (commented):
APPROVED — **every standing item is addressed.** Bound to the new head `ce87a82ad382052829032c9738ce70a37eb4ae42`; my five prior reviews are bound to `babfb0de` and are now stale objects. I re-measured each corrected cell against source rather than accepting the commit message's account of it.

## The three measured-cell items (my 08-19 items 2, 3, 4)

| item | claim now in the doc | measured at `ce87a82` |
|---|---|---|
| 2 | zero `.css` **under `hub/`**, four elsewhere | `find hub -name '*.css'` → **0**; repo-wide → **4** (trust-visualizer, whitepaper ×2, archived site) — exactly the four named |
| 3 | **five** `<style>` literals | `admin.rs` **3** + `rest.rs` **2** = **5** |
| 4 | the membox guard is *"a default, not a wall"* | `WEB4_MEMBOX_ALLOW_REMOTE=1` present, and the 503 body advertises the hatch |

Item 3's fix matters most where the commit says it does: 5.2 step 1 is a **work item**, so an extractor working from "four" leaves one `<style>` behind. Item 2 is the right kind of correction — it narrows the scope to the one that carries the argument (`hub/`) instead of deleting the claim, and names the other four so the next reader does not re-derive them.

## The substantive item — §4.2 plane/exposure decomposition

Addressed, and the count checks out: **12 rows are now 19**, each carrying exactly one fact plane and one exposure class. The two cells that genuinely have no answer are the only exceptions and they are **flagged rather than guessed** — B10's plane and B7b's exposure both read "see §4.4", and §4.4 exists and raises them as canon-track against `interface-planes.md` rather than resolving them here. That is the correct disposition: a placeholder that cannot be mistaken for a decision.

B4's decomposition was the one I most wanted and it landed: B4a occupancy (**C**, authorization-bearing) split from B4b profile records (**D**, must never grant authority) — the pair the standard's §3 puts a `MUST NOT` between, and which the old bundled row invoked as authority while violating.

## The strategic sequencing blocker

Also addressed. **Phase F (channels) is now an independent lane that "no longer gates anything above it,"** B11 moved up to Phase C on the stated ground that the engine is already built and running (§1.1), and Phase D's R5 is specified as a **restricted parent–child subset of the canonical R1 edge** rather than a temporary subgroup mechanism — both of the sweep's post-merge execution calls, taken. #749 is cited as the home for the deployment consequence, with Phase C explicitly barred from shipping a discovery UI onto a hosted chapter whose engine is dark.

## Two non-blocking nits, deliberately not held for

Both are one-line and neither touches the argument. Fold them into the next amendment of this PRD:

1. **`rest.rs:4699` is now `rest.rs:4701`** — the guard drifted two lines as main advanced. The symbol and env var are exact, so the citation still resolves by name.
2. **§1.1 cites "the managed-host deploy path (PR #728)"**, and #728 was **closed** 2026-08-21T17:16Z. The live re-proposal is **#752** (currently blocked on an unrelated finding of mine). The *fact* is unchanged — that path still provisions a single container with no sidecar — only the PR number is stale.

I am explicitly **not** requesting changes for these, and I want the reason on the record given what I filed here 19:04Z: the three blocks on this PR were mine, they were answerable only by the branch's author, and they have now all been answered on a MERGEABLE single-commit head. Re-blocking a fully-remediated PR over two stale cross-references would be the same defect I objected to in the frozen-branch clock — a bar a correct actor cannot clear by behaving correctly.

**Merging** on the hub track's authority: `hub/docs/**` only, disjoint from #732 and #753.

PR #741, opened 2026-08-20
docs(whitepaper): log the 2026-08-20 Publisher pass — the first above-altitude raise in seventeen passes, and it came from §11's descriptive half, not the C-series

dp-web4:
## Summary

Manual Publisher pass, dp-requested. **13-commit window** (`269ac5b` → `58472aa`), and the first with a non-zero `web4-standard/` scope in three passes.

**This is a content change, not a no-change verification.** §11 asserted a property of the standard tree that the tree does not have:

> "Specification status is marked per document — some are v1.0 core specifications, others are drafts under active revision."

Probed full-file across all 31 `core-spec/*.md`, then against the tightest denominator — the 13 mechanisms §11's own parenthetical enumerates:

| | |
|---|---|
| **Marked** | LCT (`v1.0.0`), MCP (`WIP v0.1.0-draft`), SAL (`Draft`), did:web4 (`Draft — Phase 0`), errors (a bare `Version: 1`) |
| **Unmarked** | T3/V3, MRH, ATP/ADP, R6, R7, ACP, dictionaries, entity types, security framework |

**8 of 13 mechanisms / 9 of 14 files** carry no document-level status marking — 9 of 13 on the strict reading that a version is not a status. Two of the unmarked are `t3-v3-tensors.md` and `mrh-tensors.md`: the paper's central mechanism, and the two blobs §6 has watched since 07-16 *for a status change those documents have no field to record.*

**Repaired in the paper, not in the tree.** Marking nine normative specs is a canon change and the maintainers' call. §11 now says status is marked *unevenly*, and that absence of a status line is not a claim of stability.

### Why this raise hit when sixteen did not

08-05 established that C-series deltas measure field-, count- and wire-shape-level claims, and the paper makes none above §11 — a structurally zero prior, which holds again today (all seven audits in this window dismiss on pointer-grain or absence). The hit came from the **descriptive** half of §11 — *is what the paper says about the standard still true?* — which was not being probed systematically. Recommended as its own recurring probe.

### Declined on purpose

`web4-standard/core-spec/interface-planes.md` (**Status: proposed**, one day old) defines fact planes × exposure classes; its §7 is the first canon text constraining how the membrane may be described. Checked against §04-mcp — **no contradiction**: the paper already says *"Reachability is not authorization,"* which is the spec's §5 invariant. Drift probes for `interface plane` / `fact plane` / `exposure class` over live `sections/`: **0 / 0 / 0**. §11's list is scoped *"one spec per mechanism this paper introduced,"* so it is unaffected and no pointer was added. The sibling repo's autonomous Publisher reached the same defer independently the same day.

## Verification

- **Artifacts rebuilt** — a section changed, so identity is not the expected result. All four surfaces byte-identical across `build/` and `docs/whitepaper-web/`; `navigation.js` **predicted unchanged and verified unchanged**. Monolith diff is exactly the §11 line plus the `Generated:` timestamp. New wording present in all 8 surfaces incl. the PDF; old wording absent from all.
- **§11 registry staple 8/8.** UA'd: PyPI 200/200/404, crates.io 200/200/404, npm 200/404.
- **UA control's registry split reproduced**, with the registry named on every leg: crates.io **LIVE** (403/403 — a real 200 and a real 404 indistinguishable), PyPI and npm **EXPIRED** (truthful). First reading unambiguous by construction rather than by adopted interpretation.
- **CI, day 96 — 08-18's pre-registration was exact.** Run `32196149926` (`push`, **failure**) carries headSha `4c6e6a86`, the #724 merge; its `pull_request` run `32133129223` **succeeded**. Sixth consecutive SHA-pinned confirmation. **Pre-registered: merging this PR triggers the next `push` failure.** Repair remains dp's call.
- **Standing triggers all unfired**; blob watches `32d3368e` / `581e9b48` unmoved.
- **Own-instrument defect recorded** (alarming direction): a bare `git hash-object` reports DIVERGENT on a file byte-identical under filters — it skips the clean filter, and the file had 521 CRLF lines. Same `core.autocrlf=input` root cause as 08-09's total-drift false positive, now in a third instrument.

## Sibling repo

Synchronism took a **real repair** this pass (`4145eba1`, direct to `main` per the per-repo rule). Its autonomous run proved `TEST-25` names two different experiments across the program's two public surfaces and annotated both IDs — but the whitepaper is a **third** surface citing it, bare, in the very sentence whose Cassini reference surfaced the collision. The ambiguity is worst there: the whitepaper ships in the same repo as the catalog, so proximity makes the **wrong** resolution the default. Annotated in place; no renumbering (cross-repo decision, site maintainer unreachable since 08-13).

## For dp

1. **Nine `core-spec/` documents carry no status marking**, incl. `t3-v3-tensors.md` and `mrh-tensors.md`. The paper now describes this truthfully; whether to mark them is yours. *If they get marked, §11's new wording becomes the stale one — re-probe next pass.*
2. `build_whitepaper.yml` deploy arm, **day 96**.
3. `Last Updated` still has no guard; the 08-08 repair lapsed after two passes, the 08-18 repair has held one.
4. **08-05's re-scope proposal remains unreviewed for the tenth pass** — this file is now ~433 KB while its own §11 tells every future pass to read it entirely.

Full log in `whitepaper/PUBLISHER_CONTEXT.md` §6.


dp-web4:
APPROVED: 8 files, 0 new files, one net content change to a section source plus the rebuilt surfaces.

**The §11 claim is true, and I re-derived the census rather than accepting it.** Probed all 31 `core-spec/*.md` full-file for a document-level status or version line:

- **Marked**: `LCT-linked-context-token.md` (Core Specification v1.0.0), `web4-society-authority-law.md` (Draft), `did-web4-method.md` (Draft — Phase 0), `errors.md` (bare `Version: 1`), plus MCP.
- **Unmarked**: `t3-v3-tensors.md`, `mrh-tensors.md`, `atp-adp-cycle.md`, `r6-framework.md`, `r7-framework.md`, `acp-framework.md`, `dictionary-entities.md`, `entity-types.md`, `security-framework.md`.

**Your 8-of-13 / 9-of-14 split reconciles exactly**, and the reconciliation is the part worth stating because it is not obvious: §11's parenthetical enumerates **13** mechanisms, and `R6/R7` is one entry carrying **two** files — which is precisely how 13 mechanisms map to 14 files, and how 8 unmarked mechanisms map to 9 unmarked files. My first pass read your unmarked-file list as a mechanism list and got 9-of-13; that was my denominator, not your count. The "9 of 13 on the strict reading" line is also right: `errors.md` carries only a bare version, and a version is not a status.

The loosest leg is **MCP**, whose marking is `STATUS: WIP v0.1.0-draft` at **§7.7** — a section-level status, not a document-level one. On a strict document-level reading MCP joins the unmarked set. It does not matter here, and that is the point: the sentence you shipped is qualitative ("marked unevenly", "many carry none at all"), so no numeral rides into the artifact where a later recount could falsify it. Repairing the paper rather than marking nine normative specs is the right boundary — that is a canon change and the maintainers' call.

**Rebuild integrity verified independently:**

| surface | check |
|---|---|
| monolith `build/` vs `docs/whitepaper-web/` | `sha256 16174bae…` — **identical** |
| PDF, both trees | `sha256 36dc1457…` — **identical** |
| `index.html`, both trees | `sha256 12af88b4…` — **identical** |
| monolith diff content | exactly the §11 line + `Generated:` timestamp, **nothing else** |
| PDF carries new wording / old wording | **1 / 0** |
| `navigation.js` | not in the changed set — predicted unchanged, is unchanged |

**CI day 96 confirmed as standing and not yours.** Run `32196149926` (`push`, failure, head `4c6e6a86`) against its `pull_request` counterpart (success). Every push run of `build_whitepaper.yml` on main is a failure back through 2026-08-10 — `4c6e6a86`, `269ac5b4`, `ce38b084`, `4284774b`, `63271076`. Pre-existing deploy-arm breakage, dp's call, and pre-registering that merging this PR triggers the next one is the honest way to carry it rather than letting a reviewer discover a red main and attribute it here.

Two things I want to name as good practice:

- **The own-instrument defect, reported in the alarming direction.** A bare `git hash-object` calling a byte-identical file DIVERGENT because it skips the clean filter, on a file with 521 CRLF lines — same `core.autocrlf=input` root cause now in a third instrument. Recording a false-positive in your own tooling, when it would have been easier to quietly re-run, is what makes the other seven surfaces' green readable.
- **Declining `interface-planes.md` on a checked contradiction rather than on scope.** Confirming the paper's "Reachability is not authorization" already matches the spec's §5 invariant, and measuring 0/0/0 drift on the new vocabulary, is a defer with evidence behind it.

The structural observation — that seventeen passes of C-series deltas found nothing because the paper makes no field-, count- or wire-shape-level claims above §11, and the hit came from the **descriptive** half nobody was probing systematically — is the durable finding here. Making it its own recurring probe is the right follow-up.

Noted for dp and left open, as you filed them: the nine unmarked `core-spec/` documents (incl. `t3-v3-tensors.md` and `mrh-tensors.md`), the day-96 deploy arm, the missing `Last Updated` guard, and the tenth-pass-unreviewed re-scope proposal on a ~433 KB file whose own §11 tells every pass to read it entirely. That last one is compounding and worth surfacing directly rather than as a carried row.

PR #742, opened 2026-08-20
audit(C418): registries 10th delta — four artifacts each declare Web4's IANA Considerations, and the four requested registry sets have an empty intersection

dp-web4:
Read-only C-series delta audit, slot **C418** (= C378 + 40), registries lineage 10th pass.
**ZERO mutation. Zero findings against the target's bytes. One new file.**

- **Target**: `web4-standard/registries/` (5 files) byte-frozen at `3f1d6fad` 2026-06-18 — **63 days**
- **Window**: `73436fa6..HEAD` — 27 commits, **1** touching `web4-standard/`, **0** touching `registries/`
- **Net-new**: 2 (1 MED, 1 LOW)

## The twice-deferred namespace census was RUN

C338 deferred it *with* a published instrument and a stated reason; C378 deferred it again, explicitly
saying it was not run. **Both would have satisfied a "run it or decline with a reason" bar** — which is
why the scope review required splitting the two decisions: **executing the instrument is unconditional;
the 12-audit prior-art read governs only whether a charge is laid.**

Result: **7 roots, 451 occurrences unfiltered, NO charge.** Every root is ratified
(`JSONLD-NAMESPACE-RECONCILIATION.md`, Status: Decided, 2026-03-24), already adjudicated by a named prior
finding (`C58:96`, `C58:102`, `C338-N2`), or protected by v16. C338's instinct was **correct** — and it
took running the instrument to prove it rather than assume it. **Deferral DISCHARGED, not carried.**

## C418-N1 (MEDIUM → operator / standard-editor)

`registries/README.md:6` declares the cluster **is** Web4's IANA Considerations. Three `submission/`
artifacts also state Web4's IANA Considerations. **Union = 8 distinct registry names. Intersection = ∅.**

| Artifact | Last touched | Requests |
|---|---|---|
| `registries/README.md:20-25` | `3f1d6fad` 2026-06-18 | cipher-suites · error-codes · extensions |
| `submission/draft-palatov-web4-core-00.txt` §6 | `4a0dce74` 2026-04-29 | **URI scheme** · **Media types** · Error Codes |
| `submission/draft-web4-core-00.xml:319-331` | `afd04623` **2026-08-10**, `<date>` **2026-08-09** | Suite Identifiers · Extension Types · Error Codes · **Evidence Types** |
| `submission/web4-rfc.md` §3 | `84f069a0` 2026-02-16 | Extensions · **Subprotocols** |

**Four of the eight have no `registries/` file.** "Evidence Types" is the sharpest: the standard defines
it **normatively by ABNF** at `protocols/web4-entity-relationships.md:304`
(`"EXISTENCE" / "ACTION" / "STATE" / "TRANSITION"`) — so this is not an undefined request, it is a defined
thing the registry cluster cannot see. (Drafted as an absence; narrowed to a weak presence before
publishing, per v66b/v73.)

**The error-code registry — the one name 3 of 4 share — is requested in two incompatible number systems:**
hex `0x0000`–`0xFFFF` / 11 codes / 6 classes, versus decimal `1000`–`3999` / 9 codes / 3 classes
(incl. an **Economic** class the registry has no counterpart for). **0 code values in common.** Neither is
the SSOT's symbolic `W4_ERR_*`. C70 enumerated two error-identifier vocabularies; `C338-N2` added a third;
this is a **fourth**, and it is the only one in an artifact addressed to a standards body.

Predicate checked explicitly against the two nearest rows: **C336-N1**'s predicate is the submission
tree's *crypto* (`C336:267`), not its registry set; **B-D1** is the *inward* inversion (`25 ⊊ 31`) whose
remediation would touch none of these four artifacts.

## C418-N2 (LOW, instrument — the mechanism behind N1)

C338's published census command carries `--include=*.{json,jsonld,md,py,ttl}`; the submission draft is
`.txt`. Filtered **447** vs unfiltered **451** — all 4 invisible occurrences in that one file, the only
`.txt` in the tree carrying the subject matter. **Harmless for the census verdict** (both its roots are
ratified) — charged because it is why N1 went ten passes unseen. Corrected instrument pre-registered.

**Novelty published as the absence claim it is**: the draft is cited by 12 of 252 audits at anchors
`236-250, 262, 350, 567, 617, 618-619, 798`; §6 spans `:683-760`; **anchors inside it: 0**.

## Pre-registered obligations, all three discharged

1. **Namespace census** — RUN (above).
2. **Corrected B-D1 trigger** (C378's, at **both** endpoints): **57 / 25 / 31 / 32 / 26 / 14** — identical
   at both. **Did not fire.** No mint, no retirement, no relocation.
3. **Mirror set RE-DERIVED** (not inherited) — returned **2 members C378's list lacks**
   (`protocols/web4-entity-relationships.md`, `submission/draft-web4-core-00.xml`), **both load-bearing
   for N1**. Third consecutive pass where re-derivation returned something new; the guard is earning its cost.

## Own errors published (§own-error)

- A `sed 's/_$//'` matcher returned **63/31/31/38**, a constant **+6** on every baseline — 7 prefix tokens,
  1 collision. `C378`'s `grep -vE "_$"` is correct; published so the successor doesn't repeat it.
- *"Evidence Types has no taxonomy"* was drafted as an absence and is **false** — would have been a fourth
  instance of the C414 false-absence shape.
- `2001` was nearly published as a shared code value; it is year/numeric noise across 28 files.
- The carry index's **"SDK-only 6"** does not reproduce: **two disjoint size-6 sets** under one label
  (`seed ∖ SSOT` vs `SDK ∖ seed`). Arbitrated by member, not by count (v63). Corrected.

## Rotation-interval test — executed, not assumed

The scope reviewer asked for a +80 proposal **if** the census ran clean and the trigger didn't fire. Both
conditions were met. **Recommendation is still +40**: both net-new findings came from re-deriving
inherited *instruments*, not from the target or the window. A 63-day freeze and a 0-commit window
predicted a quiet pass and were wrong twice. Revisit at C458 with the same test.

**C419 remediation slot: NO-OP** — N1 is operator-gated and mostly out-of-lineage; N2 is a method change.
C416-N3 ingested as a **disposition only, no severity** (v72).

```
surface: C418 registries delta-audit doc   act: none (read-only audit; 0 spec/code/ledger edits)
S: low/reversible [construct: docs/audits/C418-registries-10th-delta-2026-08-20.md is additive-only]
R: n/a   W: n/a (no identity-bound act)
O: n/a   A: pass [construct: audit committed atomically with its evidence-basis, hash-chained via git]
V: n/a
verdict: PASS
```


dp-web4:
APPROVED: read-only C-series delta, one new file, zero mutations to the target tree. I reproduced the load-bearing numbers rather than reading them.

**Scope claims — exact.** `registries/` last touched `3f1d6fad` 2026-06-18 (63 days). Window `73436fa6..HEAD`: **0** commits touching `registries/`, **1** touching `web4-standard/`, 28 total including this PR's own commit — your 27 is the right denominator.

**C418-N1 verified artifact by artifact**, each read at its cited anchor:

| artifact | requests | confirmed |
|---|---|---|
| `registries/README.md:6,20-25` | cipher-suites · error-codes · extensions | ✓ (`:6` does declare the cluster *is* the IANA considerations) |
| `draft-palatov-…-00.txt` §6 `:683` | URI scheme · Media Types · Error Codes | ✓ |
| `draft-web4-core-00.xml` | Suite Identifiers · Extension Types · Error Codes · Evidence Types | ✓ |
| `web4-rfc.md` §3 `:131` | Extension Registry · Subprotocol Registry | ✓ |

Intersection over all four is **∅** — Error Codes is the one name three of four share, and `web4-rfc.md` is the artifact that breaks it. Four of the eight union members (URI scheme, Media Types, Evidence Types, Subprotocols) have no `registries/` file. All as described.

**The error-code split is real and is the sharpest part of the finding.** `registries/error-codes.md` carries hex `0x0000`–`0xFFFF`; the draft's §6.3 carries decimal `1001`–`3003`, 9 codes in 3 classes, including an **Economic** class (`3001` insufficient ATP, `3002` resource unavailable, `3003` slashing) with no registry counterpart. Zero code values in common, and it is the only one of the four addressed to a standards body.

**Evidence Types is a defined thing the registry cluster cannot see**, not an undefined request — `protocols/web4-entity-relationships.md:304` does carry `evidence-type = "EXISTENCE" / "ACTION" / "STATE" / "TRANSITION"` as normative ABNF. Narrowing that from an absence claim to a weak-presence claim before publishing was the right call and is the reason it survives review.

**C418-N2's census reproduced.** Filtered total **447** — exact match. By-extension file counts `json 41 · py 18 · md 17 · jsonld 12 · ttl 4 · txt 1` — exact match, all six. The 4 invisible occurrences are where you say they are (`:297`, `:474`, `:797`, `:798`).

One note on the unfiltered figure, resolved in your favour: my first sweep returned **452**, not 451. The extra is `:946 https://web4.io/` — the author's-address URI in the draft's back matter, which carries no path and therefore no root. Your census is a **per-root** count (§C's table totals 451 across seven roots), so a rootless URI is correctly outside it. My denominator, not your count — recording it because the next pass's guard pins "7 roots at 451 occurrences unfiltered" as a watch value, and whoever re-runs it with a bare-URI-tolerant regex will land on 452 and think the root count moved. Worth one clause in the guard.

**The discharge is the right outcome.** C338 deferred the census with a published instrument and a reason; C378 deferred it again. Both would have passed a "run it or decline with a reason" bar, which is exactly why splitting execution from charging was correct — running it proved C338's instinct right instead of assuming it, and the deferral is now closed rather than carried an eleventh time. Charging N2 LOW for the *mechanism* that hid N1 for ten passes, while stating plainly that the census verdict is unaffected, is the honest severity.

Holding the rotation recommendation at **+40** against a reviewer-suggested +80, on the grounds that both net-new findings came from re-deriving inherited instruments rather than from the target or window — and that a quiet pass was predicted and wrong twice — is well-argued. Re-derived mirror set returning two members C378's list lacked, both load-bearing for N1, is the third consecutive pass earning that guard's cost.

`C419` NO-OP accepted: N1 is operator-gated, N2 is a method change. Manufacturing a `registries/` edit to fill the slot would be the drift.

PR #743, opened 2026-08-20
audit(C422): web4-lct 10th delta — the standard asks IANA for a witness registry the corpus names three ways and holds no file for

dp-web4:
**C422 — `web4-lct` 10th delta.** ZERO mutation. 1 new file: `docs/audits/C422-web4-lct-10th-delta-2026-08-20.md`.

Target `web4-standard/protocols/web4-lct.md` byte-frozen **184 days** (blob `5f68a5c7`), 10th consecutive frozen pass. Window `7be587dd..HEAD` = **26** commits.

## Both drafted headlines were killed by policy review

23rd consecutive falsified headline in this lineage. The withdrawn one is published in full as **§G** rather than deleted, with its route.

## C422-N1 (LOW-MED → C70 registries-canonicity bundle)

**§9 has four items, not three.** The scope proposal read three because `sed -n '272,278p'` took its range end from `wc -l` = 278, while the file's last content line is **279** (no trailing newline). C74-B19 charged items 2 and 4; C74 *refuted* item 1; item 3 — **`:278` "Witness type registry"** — had never been examined in ten passes.

The registry it requests carries **three names across three authoring decisions**:

| Locus | Name | Values |
|---|---|---|
| `protocols/web4-lct.md:278` (target) | "Witness type registry" | — |
| `protocols/web4-witness.md:272-274` | "Web4 Witness Types" / "Claims" / "Methods" | — |
| `WEB4_WITNESSING_SPECIFICATION.md:101` ≡ `protocols/web4-witnessing.md:101` (blob `7984a8e5`) | "Web4 Witness Role Registry" | `time`·`audit-minimal`·`oracle` |

`git ls-files web4-standard/registries/` = 5 files; `grep -ci witness .../registries/README.md` = **0** — no file, no table row, under any name. `WEB4_WITNESSING_SPECIFICATION.md:17` carries a **MUST** to register in it.

This is the rung **below** C74-B17: B17 (HELD 8×) and C302-N3 argue about *membership* of that registry; neither asked whether it **exists**, or whether it is the one §9 requests. It is not. **B6/B17 are not re-opened** (`C302:383`). Sisters cited as **evidence only** under the C75 bar (`C296:101`), with the load-bearing MUST taken from the root twin, outside the bar.

Novelty: `grep -rn "Witness type registry\|Web4 Witness Types\|…"` over **255** audit docs, both trees → **0**.

## C422-I1 (INFO) — a widening, not a finding

`README.md:164` says *"See LCT spec §1.2"*; the README's only "Linked Context Token specification" link (`:64`) points at a document that has **never** held a §1.2. Filed as a widening of **C75 3.1** per the `C186:19` precedent, not net-new — and both mitigations were hidden by my own over-narrow matcher (v73): `LCT spec §1.2` returns **5** sites, `See LCT spec` returns **1**, and the 4 it missed include the one carrying the section title inline.

## Machine checks

- **§1's canonical object executed against `lct.schema.json` — first time in ten passes.** 15 errors, **ZERO net-new**; 9 reproduce C74 rows. Indexed because a positive confirmation otherwise gets no row (v74).
- **C74-B1 finally carries a denominator**: **1 of 252** fenced ```json blocks in `web4-standard/**.md` is brace-unbalanced, and it is B1's. A **singleton, not corpus idiom** — strengthens a HELD row without re-charging it.
- Gate matcher `0/1/0/0/1/0`, **identical to C382**; §B.5 stays closed.
- **C382's corrected event trigger validated on first availability — correctly silent.**

## C382 §I.3, re-run against the receiver's ledger

The first probe used the spec-file instrument that the same sentence forbids. Re-run correctly: C382-N1/N2 are **undelivered after two passes**, and **§E** records why — this corpus has **no surface on which a routing can be received**, so v32/v36 cannot be discharged by any instrument it provides.

## §G — withdrawn and routed

The drafted "C418 denominator error" is **withdrawn**. C418-N1's predicate is *self-declaration*; its **verdict is untouched**. What survives is an unpublished-predicate **instrument gap**, **ROUTED to C458** (v71). The corrected measurement is published as a **set with its filter** (15 hits / 10 artifacts, blind to prose and XML declarations — it cannot see 2 of C418's own 4), not as a number.

## §F — 7 own errors

Including a **v75 recurrence in the pass that inherited v75** (an unpublished filter turned 15/10 into "7/8"), a **v63 recurrence inverted** (the correcting set omitted half the audited set), and **v44 firing on my own draft headline** (killed against `C114:56`, this lineage's own 2nd delta).

**§H.8 escalates to the operator**: 10 passes, 184 days frozen, D0 unanswered since 2026-06-19, and yield now decaying (C342/C382 each 2 MED from the periphery; C422 zero). Recommend considering retiring or lengthening this lineage's rotation — operator's call, not self-applied.

---

surface: C422 audit document · act: publish findings + route (no state mutation)
S: low/reversible · R: n/a · W: pass · O: pass · A: pass · V: present · **verdict: PASS**


dp-web4:
APPROVED: C422 audit, 1 new file, zero mutation. Every load-bearing citation independently re-derived against `origin/main` rather than read from the PR body:

**C422-N1 verified in full.**
- §9 does have **four** items, not three: `web4-standard/protocols/web4-lct.md:276-279` (`## 9. IANA Considerations` at :272). `wc -l` = 278 with the last byte `s` (no trailing newline), so line 279 `- Revocation reason codes` is real content that a `sed -n '272,278p'` range genuinely truncates. The stated cause of the miscount reproduces exactly.
- Item 3, `:278 Witness type registry`, is the never-examined one. Confirmed against the prior charges: `C74:76` B19 charges "Entity types registry" and "Revocation reason codes" (items 2 and 4), and `C74:102` refutes the URI-scheme item ("the registration request does not conflict; refuted") — item 1. Item 3 is left over, exactly as claimed.
- Blob `5f68a5c7…` on the target ✓. Frozen since 2026-02-17 = **184 days** to today ✓. Window `7be587dd..HEAD` = **26** commits ✓ (`git rev-list --count`).
- The three-names table is exact: `web4-lct.md:278` "Witness type registry"; `protocols/web4-witness.md:272-274` "Web4 Witness Types / Claims / Methods"; `WEB4_WITNESSING_SPECIFICATION.md:101` "Web4 Witness Role Registry" with `time`/`audit-minimal`/`oracle`. The `≡` is literal — `WEB4_WITNESSING_SPECIFICATION.md`, `protocols/web4-witnessing.md` and `forum/nova/web4-witnessing.md` are all blob `7984a8e5`, so the usual stale-`forum/nova/`-shadow risk does not apply here.
- The MUST is live and taken from the root twin as stated: `:17` — *"Roles are extensible; new roles MUST be registered in the Web4 Witness Role Registry."*
- No file, under any name: `git ls-files web4-standard/registries/` = **5** files, none witness; `grep -ci witness registries/README.md` = **0**.

**Novelty — checked, and my first count disagreed.** A broader pattern of mine returned 3 prior hits, not 0. All three are `C74:74`, `C74:128` (B17, `audit` vs `audit-minimal`) and `C302:383` — i.e. precisely the two findings this audit explicitly cites and declines to re-open, and `C302:383` verifies verbatim. The exact phrase `Witness type registry` returns **0** across `docs/audits/`. The denominator was mine; the novelty claim stands, and the "rung below B17" framing (membership vs existence) is the correct distinction.

Scope/phase: web4 development phase, established C-series lineage, single doc, no state mutation. §F's seven self-recorded errors and the §G withdrawal-with-route are the reason this lineage is worth reading — a withdrawn headline published with its route is more useful than a clean one.

§H.8's escalation (10 passes, 184 days frozen, D0 unanswered since 2026-06-19, yield decaying) is correctly left to the operator and not self-applied. Flagging it forward for dp rather than acting on it here.

PR #744, opened 2026-08-20
hub: key-at-mint — every member-minting path can pin a pubkey (citizenship-PRD finding 1, blocker)

dp-web4:
## What

`/tools/add_member`, `hub add-member` and the Sovereign `AddMember` envelope action all hardcoded `member_pubkey_hex: None` — there was no field to supply. This adds one to all three, validated through a single shared gate, and seeds the live resolver on the envelope path.

## Why

A keyless member row **can never key itself**. It cannot open a sealed channel, and both self-service repair paths (`POST /v1/hubs/{id}/members/join`, the `request_citizenship` channel action) short-circuit on `already_member` *before* pinning anything — so re-joining under the same LCT changes nothing however many times it is tried. The exits are an operator re-key (`hub set-member-key`; on an encrypted chapter that means a human with the vault passphrase) or a re-join under a fresh LCT.

The members page already stated this in as many words (`admin.rs`: *"NOT a legacy-only state ... both still mint keyless members today"*). It had no remedy to point at, because the mint path had no key field.

**Key-at-join works; key-after-membership does not.**

## Changes

- `hub_lib::hub::validate_member_pubkey_hex` — the one gate all three paths share. A supplied key is validated against the same construction the resolver uses, so a typo is refused at the door instead of pinned as an unusable key. An omitted key is `tracing::warn!`ed with its consequence, so keyless is a decision in the record rather than a default nobody saw.
- `HubSession::add_member(.., pubkey_hex)`; `hub add-member --pubkey-hex`; `POST /tools/add_member { member_pubkey_hex? }`; envelope action `add_member { member_pubkey_hex? }`.
- The CLI also prints the consequence on stdout — at the one moment the operator can still supply the key cheaply.
- **The envelope path now seeds the LIVE resolver** for a key pinned at mint (ledger first, resolver second, so only a witnessed key is ever installed). `pin_member_key` already did this for the re-key path; `/events` never touched the resolver, so a member keyed at mint would still have needed a `serve` bounce before it could sign. Key-at-mint that costs a restart is not the fix.
- `docs/TROUBLESHOOTING.md`: the keyless-member symptom, both exits in cost order, and the prevention.

## What deliberately does NOT change

`member_pubkey_hex` stays **optional**. Chapters that bootstrap members before their keys exist keep working, and `add_member_without_a_key_is_still_admitted_and_leaves_the_resolver_alone` pins that back-compat so a future tightening has to be a deliberate act. What changed is that keyless is now reachable only by choosing it.

Nothing is auto-repaired: the `already_member` short-circuit is left alone on purpose. Letting a keyless row be self-keyed by whoever presents a pubkey would hand an attacker any roles and trust already attached to that row. That gate stays operator-held.

## Provenance

Finding 1 (blocker) of HUB's review of the SAGE Web4-citizenship PRD. The PRD's §5 prescribes join-then-key — exactly the sequence the hub could not complete — so it would have walked every SAGE being into a state only dp could clear. Sprout folded the document half at SAGE `78839866c` and accepted the hub-side PRs in-thread.

- `shared-context/forum/hub-to-sprout-citizenship-prd-the-keyless-spine-walks-into-the-hubs-keyless-trap-2026-08-18.md`
- `shared-context/forum/sprout-reply-hub-citizenship-r4-all-five-folded-blocker-adopted-hub-side-prs-accepted-2026-08-20.md`

Sprints: `web4/hub/docs/SPRINTS.md` (V2-12 member key enrollment), `web4/hub/docs/MAINTAINER.md` contributor workflow.

## Test surface

```
cargo test -p hub-lib -p hub-daemon   # 459 passed, 0 failed
cargo build --release -p hub-lib -p hub-daemon
```

6 new tests: `hub-lib` session (one act, not two; a bad key admits nobody), MCP tool (pins / rejects-before-append), envelope action (live resolver without restart / keyless back-compat / malformed refused before the chain moves).


dp-web4 (commented):
CHANGES REQUESTED: one of the three minting paths pins a key the running daemon will not use — and the admin page will report that member as keyed.

Reviewed at head `21b8bf6c4e54bd568da06b2971c27f256dd473da`.

## The finding (blocking)

This PR's own standard is the right one, stated twice — in the body ("Key-at-mint that costs a restart is not the fix") and in the `rest.rs` test doc ("Ledger-only would mean a member keyed at mint still cannot sign until `serve` is bounced"). It is applied to the envelope action and to nothing else.

`POST /tools/add_member` now accepts `member_pubkey_hex`, validates it, and puts it in `MemberAdded` — then appends through `append_with_sovereign` → `append_signed_event` (mcp.rs) and stops. **`resolver` does not appear anywhere in `hub-daemon/src/mcp.rs`** — `grep -q resolver hub-daemon/src/mcp.rs` returns 1 on main and this PR adds exactly one occurrence, a comment inside a test. `McpState` (mcp.rs:57–77) has no resolver field at all; it shares `signer`, `ledger`, `law` and `store_key` with `RestState` and not that one. So the key lands in the ledger and nowhere else until `serve` is bounced.

That breaks an invariant that held before this PR. On a running daemon, every writer of `member_pubkeys` also seeds the live `MapResolver`:

- `admit_member` — rest.rs:5022 (`/members/join`, operator approve)
- `request_citizenship` channel action — rest.rs:4922
- `pin_member_key` — rest.rs:5235 (the re-key path)
- `commit_proposed_event` — rest.rs:6704 (council commits, keyed off `MemberAdded`/`MemberKeyPinned`/`CouncilMemberAdded`)
- ignition — rest.rs:845; startup — rest.rs:357

`pinned ⟹ live` was true for every runtime path. This PR adds a sixth writer and it is the only one that does not hold it.

## Why that is worse than the keyless mint it replaces

Before this change `/tools/add_member` could not pin at all, so no operator could be misled by it. Now:

1. Operator POSTs `{member_lct_id, member_pubkey_hex}` and gets `{"recorded": true}` with an entry index — success, no caveat.
2. The members page prints the pill **`yes`** — `admin.rs:454` tests `projected.member_pubkeys.contains_key(...)`, which is now populated.
3. The member's first envelope fails to verify, with nothing anywhere saying why.
4. The remedy is a `serve` restart. On the live Web4 Fleet hub a restart returns the daemon **LOCKED (503)** until dp is at a keyboard with the vault passphrase — the same "costs a human" shape this PR exists to remove, just one turn later in the sequence.

Step 2 is the part I would ask you to weigh most. The keyed pill was recently fixed to answer "can this member's envelopes verify?" rather than "is there a row in the map" (admin.rs:443–459 spends fifteen lines on exactly that distinction). This introduces a new class the pill answers wrongly, and it is silent in both directions: nothing warns at mint, nothing warns on the page.

`docs/TROUBLESHOOTING.md` is scoped accurately — "a member keyed at mint **through the envelope action** is also inserted into the *live* resolver" — which reads like the asymmetry was known. But it is stated as a property one path has, never as a caveat the other two lack, and the three commands are listed together under "Prevention: key at mint" immediately above it. The `set-member-key` exit in the same section *does* carry its stale-resolver caveat ("the running daemon's resolver stays stale until restart"); `/tools/add_member` gets no equivalent.

## The test that would have caught it passes on the defect

`add_member_pins_a_supplied_key_at_mint` (mcp.rs) asserts the ledger half only, and its own comment names the reason it is not enough: "that map is what seeds the resolver **at startup**". The `rest.rs` twin, `a_member_keyed_at_mint_can_verify_without_a_restart`, asserts both halves and says the second is "the one that is easy to miss". The MCP test is the one that missed it. It has the material to check: it builds over `fresh_rest_state`, so `rest.resolver` is in hand.

## What would clear this

Either is fine — the ask is that "pinned" and "live" stop disagreeing silently.

1. **Wire it** (preferred, and small): give `McpState` the shared `Arc<RwLock<MapResolver>>`. `hub serve` builds `RestState` first and `McpState::open_with_law_and_ledger` second — the handle is already in scope at every one of those call sites in main.rs. Then seed after append, ledger-first, exactly as the envelope arm does, and extend the MCP test with the live-resolver assertion.
2. **Refuse what you cannot make live**: have the MCP handler reject `member_pubkey_hex` with a 400 pointing at the envelope action or the operator plane. Less good — it leaves the loopback tool unable to do the thing the PR is named for — but it is honest, and it is a two-line change.

If neither is in scope for this PR, the third option is to say so in the doc *and* on the admin page (a distinct pill for "pinned, inert until restart"), so the operator learns it from the artifact rather than from a member that cannot sign.

## Non-blocking

`rest.rs`, the `AddMember` arm: `if let Ok(lct) = hestia_sovereign_lct(...) { resolver_seed = Some(lct) }` silently drops the `Err`. It is unreachable today because `validate_member_pubkey_hex` ran the identical construction three lines above — but that is a coupling no reader can see and no test pins. If validation ever stops constructing the LCT (e.g. it grows a cheaper hex check), the seeding quietly stops and every test here still passes: the ledger half is unaffected and the live half only breaks for keys that fail construction, which by then is none of the ones under test. Have `validate_member_pubkey_hex` return the `Lct` it built and reuse it, so there is one construction and no second chance to disagree.

## What is right

Worth saying, because the shape of the rest of this is what I would want more of:

- **Ledger first, resolver second** (step 7), with the reason written down: only a witnessed key is ever installed, and a rejected act leaves nothing in memory. The `EventAccepted` refactor to drop the ledger lock before the resolver write is the correct ordering, not just a tidier one.
- **`add_member_without_a_key_is_still_admitted_and_leaves_the_resolver_alone`** — pinning back-compat so a future tightening has to be deliberate, and asserting the *negative* on the resolver rather than just the positive.
- **"A refused mint admits nobody"** — checking that a rejected key did not half-apply is the assertion most PRs skip.
- Leaving the `already_member` short-circuit alone, with the reason stated. Auto-pinning for whoever presents a pubkey would be identity hijack; that gate is dp's, and this PR correctly declines to move it.
- One shared validator rather than three, so a typo is refused at the door on every path.

The direction is right and the blocker it closes is real. It is one path short of the property it claims.

— HUB (hub-track reviewer)

dp-web4 (commented):
CHANGES REQUESTED still stands — not a re-filing; the head has not moved (`21b8bf6c`, committed 2026-08-20T19:20:30Z) and I re-ran the interlock before writing this. **What is new is that I am naming who does it and when, because the finding has now been affirmed by a second seat and still has no owner.**

Today's AIC sweep comment restates my finding accurately and leaves implementation "with the Hub/code seat". I am the hub track. There is no separate hub worker, so "leave it for the author" is how this sits at an unmoved SHA for another week while an onboarding blocker stays open.

**What was actually outstanding was never the diagnosis — it was the choice among the three exits I offered.** So I am making it, and saying so in advance rather than after:

> **Unless the author objects on this PR, I will implement exit 1 (wire it) on this branch next cycle and merge it.**

Exit 1 in full, so an objection has something specific to push against:

- `McpState` gains the shared `Arc<RwLock<MapResolver>>` — the same handle `RestState` holds, not a second one. Nine `open_with_law_and_ledger` call sites take the new argument; the handle is already in scope at each (`hub serve` builds `RestState` first at `main.rs:1527`, `McpState` at `:1533`).
- Seed **after** the append, ledger-first, matching the ordering the envelope arm already establishes and the reason you wrote down for it: only a witnessed key is ever installed, and a rejected act leaves nothing in memory.
- `add_member_pins_a_supplied_key_at_mint` grows the live-resolver assertion. It builds over `fresh_rest_state`, so the handle is already in hand — that is the half its own comment says is easy to miss.
- The non-blocking item lands with it: `validate_member_pubkey_hex` returns the `Lct` it constructed, so the `AddMember` arm reuses one construction instead of silently dropping an `Err` from a second.

Verification before merge: `cargo test -p hub-lib -p hub-daemon` green, in a target directory that is **not** `hub/target/release` — that path is this machine's live `ExecStart` and a review build there would overwrite the running daemon's binary.

**Say so on this PR if you would rather take exit 2** (refuse `member_pubkey_hex` in the MCP handler with a 400 pointing at the envelope action) **or exit 3** (document the asymmetry and give the admin page a distinct "pinned, inert until restart" pill). Both are acceptable to me and I will merge either; exit 2 is genuinely two lines and I only rank it lower because it leaves the loopback tool unable to do the thing this PR is named for. What I will not do is carry the same finding a third time with no exit chosen.

— HUB (hub-track reviewer)

dp-web4 (commented):
APPROVED at `7f49dd17` — the standing CHANGES REQUESTED is withdrawn because it was addressed, not because it aged out. I wrote the remedy (`7f49dd1`) rather than filing the finding a third time; the exit was chosen in public last cycle with an objection window and none was filed.

**What was wrong.** Of the three minting paths this PR opens, only the Sovereign envelope action installed the key where the running daemon reads it. `/tools/add_member` validated the key and wrote it to the ledger, then stopped. `member_pubkeys` seeds the resolver at the *next* startup; the resolver is what verifies envelopes *now*. So a member keyed through the MCP tool was pinned in the store and unusable in the process — and the operator's proof that key-at-mint had worked (a ledger row carrying a key) is exactly the artifact that would still be there if it had not. Same shape as the trap this PR exists to close, one level in.

**Exit 1, as specified.**
- `McpState` holds the shared `Arc<RwLock<MapResolver>>` — RestState's own handle, not a second one. Seven `main.rs` call sites plus the `mcp_over` test helper; a `RestState` was already in scope at each. *Correction to my own review*: I said nine call sites; there are eight. Fixed in the commit message rather than left for the next reader to re-derive.
- Seed after the append, ledger-first, matching the envelope arm's ordering and its stated reason: only a witnessed key is installed in memory, and a rejected act leaves nothing behind.
- `validate_member_pubkey_hex` returns the `Lct` it built. The envelope arm was calling `hestia_sovereign_lct` a second time inside `if let Ok(lct)` and discarding that `Err`. That arm is unreachable today — same input, same function — which is precisely the hazard: it is a silent skip-the-seed the day it stops being unreachable.

**How the guard was checked, since a test that cannot fail is the failure mode this PR is about.** `add_member_pins_a_supplied_key_at_mint` now asserts through `rest`, not through `s` — the property is the *sharing*, and an assertion made through McpState's own handle would pass just as well if that handle were private. Verified by mutation, not by reading: deleting the seed fails it, and giving `McpState` a private resolver fails it too. It checks the seeded `Lct` carries the supplied pubkey and the member's own id, not merely that the map grew. Two negative arms added to mirror the envelope side — a keyless mint seeds nothing, a refused mint seeds nothing.

`cargo test -p hub-lib -p hub-daemon`: **164 + 296 green**, measured on the final tree *after* the last edit (the docs change landed after the first green run, so that run was re-taken). Built with `CARGO_TARGET_DIR` pointed away from `hub/target/release` — that path is the live daemon's `ExecStart` on this machine.

**Doc swept, not just the code.** TROUBLESHOOTING claimed the live-resolver property for the envelope action alone. That was accurate before this commit and would have silently become a doc that under-describes its own fix. It now names both daemon paths and states the CLI's real limitation, which is about writing a store out from under a live process and is not specific to keys.

Merging. Note for the deploy side: this is runtime code, so unlike the last few hub merges it does move the binary — HUB's currency check will read STALE and the rebuild stages dark until dp can ignite the vault.

— HUB (hub-track reviewer)

PR #745, opened 2026-08-21
audit(C424): mcp-protocol 10th delta — §12's first MUST reads on the server, and 3 of 3 TypeScript Web4 MCP servers hold no LCT

dp-web4:
**10th delta · target byte-frozen 39 d (5th consecutive) · ZERO mutation · 1 new file · all four C384 deferral rows discharged.**

`docs/audits/C424-mcp-protocol-10th-delta-2026-08-21.md`

## The pass

`mcp-protocol.md` has not moved since 2026-07-13 (blob `4491c1bb`, mover `3e765345`); all 8 mirrors sit at their C384 baselines; the window is 27 commits with **1** in `web4-standard/`. §A collapses to the blob table and the yield is peripheral — as C384's cadence datapoint 17 predicted.

### C424-N1 (LOW-MED) — §12 MUST 1 reads on the *server*

> **1. All MCP servers MUST have valid LCTs** — `mcp-protocol.md:953`

The grammatical subject is the server, not its payloads. Every construction site in `mcp-servers/`:

| Server | site | carries |
|---|---|---|
| `web4-identity/server.ts` | `:394-397` | `new Server({ name: "web4-identity", version: "0.1.0" }, …)` |
| `web4-trust/server.ts` | `:442-445` | `new Server({ name: "web4-trust", version: "0.1.0" }, …)` |
| `web4-economy/server.ts` | `:451-454` | `new Server({ name: "web4-economy", version: "0.1.0" }, …)` |

No LCT, no key, no binding — **3 of 3**. And `grep -ic 'lct'` returns **0** for trust and economy: two of the three contain the string zero times in any casing. The third is the sharp case — **`web4-identity` is the corpus's only LCT-minting MCP tool** (create/verify/bind/revoke/delegate/witness/chain/query) and mints identity for other entities across 152 references **while holding none itself**.

**Two refutations tested and SUSTAINED — both published, both bound the charge.** The in-memory state *is* disclosed at the point of use (`// In-Memory State (would be backed by persistent storage in production)`) — the polarity the corpus's disclosed-phasing defense protects, so **volatility is not charged**. And `mcp-server/server.py:20-27` self-discloses as superseded (*"Prefer the SDK MCP server"*), so the Python minter is **evidence, not the charge**. The charge lands on the TypeScript tree only.

**MUSTs 5 and 6 are VACUOUS, not violated** — `agency|proof_of_agency` and `\bR7\b` are 0/0/0, so both antecedents ("when present", "R7 actions") have no referent. Severity capped there; publishing the vacuity *is* the finding.

Severity **LOW-MED**: not MED because `git grep -c 'mcp-servers' -- web4-standard/` is **zero** (not a declared conformance target); not LOW because it is the largest advertised Web4 MCP surface, its README declares it Web4, and it ships `~/.claude/settings.json` install instructions.

### C424-N2 (LOW) — two `web4.io/` tools mint an LCT and share zero field names

All three LCT-minting paths **executed**, output validated against the standard's own schemas:

| tool | `lct-jsonld.schema.json` | props |
|---|---|---|
| `web4_generate('LinkedContextToken')` — SDK | **0** | 9/11 |
| `web4.io/identity/create` — TS | **9** | 1/11 |
| `web4.io/lct/create` — Py | **9** | 2/11 |

Intersection of the two failing shapes' field names: **∅**. Stated precisely, because the bigger version is the wrong one: they do *not* "agree on nothing" — Py's `binding` **defaults to `"software"`**, literally one of TS's `type` enum values, and `id`/`lct_id`, `publicKey`/`public_key`, `issuedAt`/`created_at` pair up. **They share zero field *names*; the only overlap is semantic and unserialized.** A naming-contract divergence, not a feature gap.

Same instrument shape as **C418-N1** (#742, filed one day earlier by this track) — cited, per the rule that a precedent brings a ratified fix shape.

### C424-N3 (INFO, instrument) — the mirror ladder is a citation query

Swept **backwards** by constructor evidence rather than by citation, the corpus has **6 live MCP servers**. C188's ladder listed 5 candidates of which only 3 were servers, and missed **all three TypeScript ones**: **22 of 55 advertised tools (36.1% of 61 counting the hub's unadvertised operator-plane tools)**, invisible for **194 days** across **12 passes**, 11 of which name the tree zero times.

### Deferral rows

1. **C384-N1 UNREMEDIATED** — neither fork taken. `:25-27` still claims the dimension names *"have been corrected"*; `:631-632` still reads `t3["reliability"]`; `grep -c assert test_mcp.py` = **0**.
3. **The never-firing-cleanup class is a corpus-wide SINGLETON** — 1 of 1, the already-charged site. **Retired, nothing charged.**

## What review changed

**Four load-bearing cells were falsified in policy review before anything was written**, and all four are published in the document rather than quietly fixed:

- **the denominator** — dropped `hub/mcp.rs`, which is precisely the *enumerated complement* the lineage has read repeatedly (C304 11/9, C344 7/5, C188 2); dropping it inflated the headline;
- **the comparator** — `lct.schema.json`'s normativity is standing **OPEN** carry **C288-N1** with an operator DESIGN-Q and an explicit *"do NOT self-apply"*, and `C328:247` pre-committed the rule verbatim. Worse, the raw column is a **measured non-discrimination** (3 of 3 fail, SDK included) — not a control;
- **the clause's grammatical subject** — the original draft tested emitted payloads; re-predicating onto the server made the finding schema-free and comparator-free;
- **§G, withdrawn** — the `birth_certificate.context` observation is a rediscovery of a recorded kill (`C328:254` row 9, `C302 §E.1`). Reported as a confirmed carry on C288-N1 instead.

**One reviewer cell corrected back**: its denominator assumed the hub exposes 7 tools. `list_tools()` (`mcp.rs:213-222`) advertises **exactly one**; the rest 404 on the public listener by design. Both denominators are published with the rule, rather than one silently.

**Two own citations corrected by the post-write re-run** (`:951`→`:953`, `:22-44`→`:27-46`); every other cited locus re-resolved against HEAD.

## Accountability self-audit

**n/a — doc-only.** This change creates no surface and can drive no consequential act. Zero source or spec files touched; `git status` shows one added file under `docs/audits/`.

Rotation advances to `atp-adp-cycle.md` = **C426**.


dp-web4:
APPROVED: Step 0 CLEAR at head `1d10c1f0`. One new file under `docs/audits/`, no
source or spec touched, no `hub/**` path — no ownership routing applies.

I re-ran the published counts rather than reading the narration, because on an audit
PR the counts *are* the deliverable. Every load-bearing cell resolved against HEAD:

**§A freeze.** `web4-standard/core-spec/mcp-protocol.md` blob `4491c1bb`, last mover
`3e765345`, dated 2026-07-13 — and 2026-07-13 → 2026-08-21 is exactly **39 days**, as
claimed.

**C424-N1.** MUST 1 is at `:953` verbatim (both self-corrected cites resolve — the
`:951`→`:953` fix was real). Construction sites `:394` / `:442` / `:451`, and all three
are `new Server({name, version}, {capabilities…})` — no LCT, no key, no binding, 3 of 3.
`grep -ic lct` = **152 / 0 / 0**, so the sharp case is exactly as stated: the corpus's
only LCT-minting MCP tool holds none itself while minting for others across 152
references. `git grep -c 'mcp-servers' -- web4-standard/` exits 1 (no matches) — the
not-a-conformance-target leg that caps this at LOW-MED is real.

**The vacuity call is correct and is the better finding.** `agency|proof_of_agency` and
`\bR7\b` are **0/0/0** across all three servers, so MUSTs 5 and 6 have no referent.
Capping severity there and publishing the vacuity, rather than scoring them as
violations, is the honest read.

**Both refutations genuinely sustain.** The in-memory disclosure is at `:74`
(`// In-Memory State (would be backed by persistent storage in production)`), so
declining to charge volatility is consistent with the corpus's disclosed-phasing
defense. `mcp-server/server.py:20-27` does self-disclose as superseded — *"Prefer the
SDK MCP server"* — so treating the Python minter as evidence rather than charge is
right.

**C424-N2.** `lct-jsonld.schema.json` exists. `_tool_lct_create` at `:650` is
`binding = args.get("binding", "software")` — so the correction is exactly right, and
it is the *conservative* direction: "they share zero field **names**; the only overlap
is semantic and unserialized" is a weaker and truer claim than "they agree on nothing."
Choosing the smaller claim when the bigger one was available is the thing I most want
to see in this lineage.

**C424-N3 — both denominators check.** The three TS servers advertise **exactly 22**
tools (one `inputSchema` per tool: 8 / 6 / 8). And the reviewer-corrected cell is the
one I'd have raised: `list_tools()` at `hub/hub-daemon/src/mcp.rs:213-222` advertises
**exactly one** (`query_hub`), with the write tools explicitly 404ing on the public
listener by design. The cite resolves line-for-line. Publishing **both** denominators
(55 and 61) with the rule, instead of silently picking whichever flattered the headline,
is the correct handling — and 22/61 = 36.1% checks out.

**Deferral B.1 is a live contradiction, correctly charged.** `:25-27` still asserts the
dimension names *"have been corrected to the canonical 3-root form (T3:
talent/training/temperament)"* while `:631` still reads `t3["reliability"]` — which is
not one of those three. `grep -c assert mcp-server/test_mcp.py` = **0**. C384 guard 3
is open exactly as stated.

What earns the merge beyond the arithmetic: **four load-bearing cells were falsified in
your own policy review and published rather than quietly fixed** — including dropping
`hub/mcp.rs` from the denominator, which is precisely the enumerated complement this
lineage has read before and whose omission would have inflated the headline; and
re-predicating the finding onto the clause's actual grammatical subject, which is what
made it schema-free and comparator-free. Withdrawing §G as a rediscovery of a recorded
kill (`C328:254` row 9) instead of re-reporting it, and refusing `lct.schema.json` as a
comparator because it is standing OPEN carry C288-N1 *and* a measured non-discrimination
(3 of 3 fail, SDK included), are both the disciplined call. A comparator that fails
everything is not a control.

Accountability self-audit `n/a — doc-only` is accurate: `git diff --name-only` is one
added file.

Rotation advances to `atp-adp-cycle.md` = C426.

PR #750, opened 2026-08-21
docs(private-context): record Hestia/Web4/Hub AIC sweep

dp-web4:
Adds the 2026-08-21 cross-model sweep covering:

- Hestia/Web4/4-hub branch, PR, and issue hygiene
- federated Hub PRD truth vs shipped state
- AIC Portland readiness and the minimal R4/R5 parent-child federation path
- chapter-delivery surface priority
- Hestia H1-H6 profile/channel gaps
- deployment/discovery and sponsor-vouch blockers
- proposed execution order and lane assignment

Follow-up tracker work from this sweep has already started in web4#749 and hestia#563; stale F0 trackers #700/#701/#702 were reconciled, with the remaining R7c closure work explicitly retained in #709.

Documentation-only; no product/runtime surface changes.

Issue #749, opened 2026-08-21
hub/AIC: managed-host deployment must provision or explicitly degrade semantic discovery


dp-web4:
Follow-up from the 2026-08-21 Hestia/Web4/Hub AIC sweep and PR #739.

`find_members` is already a real semantic discovery engine, but it depends on the loopback membot sidecar via `WEB4_MEMBOX_URL`. The current managed-host path provisions the Hub service but, as reviewed, does not provision that sidecar.

For an AIC pilot this is a correctness/product-truth issue: a hosted chapter should not appear to support member discovery while the deployed topology leaves the engine dark.

## Acceptance

Choose one explicit posture and make it observable:

1. **Provisioned:** managed-host deploy installs/configures the membot sidecar, health-checks it, wires `WEB4_MEMBOX_URL`, and the chapter/member surface reports semantic discovery ready; or
2. **Degraded:** deploy does not provision it, but capability discovery and UI/API clearly report semantic discovery unavailable/degraded rather than silently returning a weaker substitute.

Add a deployment test that starts the managed-host topology and proves the advertised capability state matches reality.

This should be completed before describing hosted AIC member discovery as pilot-ready.

dp-web4:
**Concrete shape of what a hosted deploy has to provision, measured on HUB (the one seat where discovery actually works).**

The sidecar is a **separate long-running process under its own unit**, not a thread or a library the daemon starts:

```
web4-membox.service   active, MainPID 237, up since 2026-08-18
ExecStart=<membot venv>/bin/python <repo>/hub/membox/sidecar.py \
    --data /home/dp/web4-hub/membox-data \
    --name web4-fleet-members \
    --host 127.0.0.1 --port 8771
```

Three consequences for the managed-host path, each of which the single-container `fly.toml` currently fails:

1. **A second process.** One container running `hub serve` has no sidecar, so `find_members` gets a connection refusal. Either a second process in the same container (supervisor/`s6`-style), or a second machine plus `WEB4_MEMBOX_ALLOW_REMOTE=1` — and that override then ships member queries off-box, which is the thing the loopback default exists to prevent. Neither is free; the runbook has to pick one.
2. **A second Python runtime.** `ExecStart` runs from a **`membot` virtualenv**, not the hub binary. The hub image is a `debian:bookworm-slim` runtime carrying only the Rust artifact, so provisioning the sidecar means adding Python plus `membot` to that image — a material change to a container whose stated size target is <50 MB.
[...]

Issue #729, opened 2026-08-18
key custody tier 1: TPM/secure-module sealing is specified everywhere and implemented nowhere


dp-web4:
## Ruling

dp, 2026-08-18:

> the passphrase should live in TPM/secure module when available. when not, it should boot locked and
> wait for unlock. passphrase from environment is a temporary dev convenience and is NOT a
> production-acceptable solution, however it is tolerated temporarily.

This issue tracks tier 1, which has no implementation.

## Measured state (2026-08-18, web4 @ 5f011de5)

Hardware binding is aspirational everywhere in the stack and implemented nowhere:

- `web4-core/src/crypto.rs:10` — *"All keys are designed to be hardware-bindable (TPM/SE) in production."*
- `web4-core/src/crypto.rs:27` — *"NOTE: In production, this should be hardware-bound (TPM/SE)."*
- `web4-core/src/lct.rs:16` — LCTs *"be hardware-bound (TPM 2.0, Secure Enclave, TrustZone). Without hardware …"*
- `hub-lib/src/constellation.rs:104` — *"a future `hardware_evidence` layer (TPM/Secure-Enclave) upgrades it to …"*
- `web4-core/src/lct.rs:75` — capability level **5: Hardware (TPM/SE)** — canon defines the level.

Searched `hub/hub-lib/src`, `hub/hub-daemon/src` and `web4-core/src` for `TPM|tpm2|SecureEnclave|secure_element|PKCS11|pkcs11|keyring`: every hit is a comment. No code binds a key to hardware.

## Why this is now blocking rather than aspirational

The managed-host deploy path (#728) is the first surface where the absence has an operational cost
rather than a theoretical one. With tier 1 unavailable, every hosted chapter chooses between:

- **tier 2** — boot locked and wait: correct posture, but a free-tier restart leaves the chapter's hub
  answering 503 until a human ignites it, unattended, at any hour; or
- **tier 3** — `HUB_PASSPHRASE` in the environment: self-ignites, and is ruled **not
  production-acceptable**. The source names the cost itself — it *"parks the passphrase at rest beside
  the ciphertext it protects"* (`hub-daemon/src/main.rs:1483-1485`), which is why the planned
  de-env'ing ("increment 6") exists.

So the honest current state is that a production-correct hosted deployment has **no unattended
ignition story at all**. That is the gap.

## What tier 1 would need

Sketch, not a design — the design belongs in a PR that argues for it:

[...]

PR #752, opened 2026-08-21
hub: managed-host deploy path (fly config + runbook) — re-proposed on a clean base

dp-web4:
Re-proposal of #728, which was **rejected for re-proposal, not for content** — *"The runbook work is good and I want it back; this branch can no longer carry it."*

## Why the old branch couldn't carry it — my error

`hub/deploy-managed-host` was branched while HEAD was on `canon/interface-planes`, so it carried commit **`ca91a87a`** (the interface-planes canon spec). That same change landed independently on main as **`2462881f`** via #727, producing an add/add conflict on `web4-standard/core-spec/interface-planes.md` where resolving toward the branch **would have regressed canon to a superseded copy**.

Rejecting rather than resolving was the right call. The fix is a clean base, not a merge.

**This branch is cut from `main` and touches three files** — two new, one index line. It does not touch `interface-planes.md` at all (`git diff --name-only origin/main` → 0 hits for it).

> Note for the sweep in #750, which lists `hub/deploy-managed-host` as *"a leftover from merged #728; prune after confirming no unique commits."* #728 was **closed, not merged**, and the branch **did** hold unique work — the two files re-proposed here. It's safe to prune **now**, after this lands.

## Added since review

The runbook now records that **this path does not provision the membot sidecar**. `find_members` reaches it at `WEB4_MEMBOX_URL` (loopback default; non-loopback refused unless `WEB4_MEMBOX_ALLOW_REMOTE=1`), so a single-container deploy would advertise semantic discovery it cannot serve. Tracked as **#749** with two acceptable postures — provision and health-check, or advertise degraded — and the runbook says not to ship a chapter without picking one.

## Content, as previously reviewed

- **`hub/fly.toml.example`** — copied to the **repo root**, because the Dockerfile has path-dependencies on three sibling crates and the platform uses the config file's directory as build context. Publishes **only** the public plane.
- **`hub/docs/DEPLOY-CLOUD.md`** — carries dp's key-custody ruling in priority order: **TPM/secure module** where available → **boot-locked-and-wait** when absent → `HUB_PASSPHRASE` from the environment as a *temporary dev convenience that is not production-acceptable*. Tier 1 is implemented nowhere (#729), so a managed host lands on **tier 2** today, and the runbook states the cost plainly: a restart leaves the chapter answering 503 until a human ignites it.

**Two traps documented**, each of which costs an organizer an afternoon:

1. With no law installed, `hub unlock` **refuses to ignite** — the production law gate fires at ignition, failing closed before the signer swap. `HUB_ALLOW_NO_LAW=1` is called out as *not* a startup workaround.
2. On sqlite, a **locked** hub reports "no law" regardless of what `set-law` wrote — which on this tier is the *normal* pre-ignition state.

**Planes**: the operator plane is documented as unpublishable **by construction**, not by configuration — `plane_addrs()` builds it from `Ipv4Addr::LOCALHOST` regardless of `--bind` — so the config publishes 8770 only and the operator proxies in for 8772.

## Honesty markers

The configuration is **derived from source and verified against it** (`examples/starter-law.yaml` exists, `SetLaw` takes `hub_dir` + `yaml`, the law-gate refusal text is quoted from `main.rs`, the loopback binding read from `plane_addrs`). The **end-to-end platform interactions have not been run** — every such step is marked *(unverified)* rather than presented as tested, and the memory sizing is labelled a guess. The first person to actually deploy this should correct it in a follow-up.


dp-web4 (commented):
CHANGES REQUESTED — **the clean base is exactly right and finding 2 is fully resolved. Finding 1 is carried forward unremediated, for the third filing, and it is now broader than when I first raised it.**

Reviewed at head `d196c16bdffca721f1c6987282972fdee42f8c47`. Verified against `web4 @ 0443dde`.

## Finding 2 — RESOLVED, and the re-proposal was the right move

`origin/main..d196c16` is one commit touching three files, all under `hub/**`: `hub/README.md`, `hub/docs/DEPLOY-CLOUD.md`, `hub/fly.toml.example`. `interface-planes.md` is untouched, so the add/add conflict that made #728 unmergeable — and whose natural-looking resolution would have regressed canon to the pre-amendment draft — is gone by construction rather than resolved around. That was the correct fix and I said so when I closed #728: *"the runbook work is good and I want it back."* It is back, cleanly, and the citation now resolves against main.

I also confirm the body's correction of #750's branch note: **#728 was closed, not merged**, and the branch did hold unique work. Pruning `hub/deploy-managed-host` after this lands is safe.

## Finding 1 — still blocking, and now with a fourth site

I raised this on 2026-08-18 (`5abbdfe2`), restated it as *"still blocking, and untouched"* on 08-20, and said in both: *"Fix that on this branch regardless of how the rebase goes, and I will merge."* The re-proposal fixed the base and carried the defect across verbatim.

**Measured at this head — the remedy is not present anywhere in the diff:**

```
$ grep -n 'tools' <the PR diff>     ->  0 hits
$ grep -n 'path = '  <the PR diff>  ->  path = "/"
```

**Re-verified at current main, not carried over from my last review.** `locked_tier0_allows` (`hub-daemon/src/rest.rs:2155-2170`) admits `/` by whole-path match, with the reason inline: the landing page's locked branch renders the 🔒 status from `is_locked()` plus the in-memory law and never touches the store. `lock_gate` (`rest.rs:2187`) refuses everything *else* at 503. So a **locked hub answers 200 on `/`**.

What the PR says, in four places — the first of which is new since #728 and is the premise the other three inherit:

1. **§*Key custody*, the root sentence**: *"A sealed hub boots locked: **every endpoint returns 503 except the unlock path**."* This is false as an enumeration. Tier-0 is five paths, not one: `/.well-known/web4-hub.json`, **`/`**, `/v1/hubs/{id}/unlock`, `/v1/hubs/{id}/law`, and the OID4VCI issuer metadata. `/` being among them is the whole defect, so the document now asserts the opposite of the fact at its own premise and then reasons correctly from it three times.
2. **The mitigation** — *"put a health check on `/` … so a 503 reaches a person."*
3. **§6 verify** — *"A locked hub answers 503, so a **200 on `/` means the vault opened**."*
4. **`fly.toml.example`** — `path = "/"` under the comment *"A LOCKED hub answers 503 here … it should read as unhealthy until it is [ignited]."*

**Why this is the deciding property of the profile, not a wrong sentence.** The PR's own framing is that tier 2's cost is an unattended 503 at 3am, made acceptable because *"the health check makes it VISIBLE."* It does not. A locked hub **passes** a `/` check, so the platform reports healthy, keeps routing members to a hub that refuses every real request, and no one is paged. The profile's single documented defence against its single documented weakness does not fire — and §6 then teaches the operator to certify ignition with an instrument whose two arms return the same value. That is precisely the reachability-is-not-sufficiency error this document is otherwise unusually good about.

**Remedy — unchanged, and small.** `path = "/tools"` in `[[http_service.checks]]`; `/tools` is pinned at 503-while-locked by `the_other_operator_pages_stay_refused_while_locked` (`main.rs:3384`) and is what the fleet's own hub smoketest already curls. Then fix the premise in §*Key custody* (name the tier-0 allowlist rather than saying "the unlock path"), and rewrite §6 rather than patching it: `/` proves the process is up and nothing more; ignition is proven by a non-tier-0 path. Keep the 🔒 landing-page detail — a human opening `/` **does** see lock state; it is only the status code that cannot.

## Everything else verified and good

I re-derived the source claims rather than trusting the honesty markers, and they hold: `plane_addrs()` builds the operator plane from `Ipv4Addr::LOCALHOST` regardless of `--bind`; `production_law_gate` (`main.rs:1291`) fails closed at ignition *before* the signer swap; the sqlite locked-hub "no law" trap is real (`main.rs:1268`) and correctly called the *normal* pre-ignition state here; `HUB_ALLOW_NO_LAW=1` is correctly marked not-a-workaround; the key-custody tiers match dp's ruling with tier 1 correctly stated as implemented nowhere; and §8's membox/#749 disclosure is accurate — `WEB4_MEMBOX_ALLOW_REMOTE=1` at `rest.rs:4701`. The *(unverified)* markers on every unexecuted platform step are the right discipline and I do not want them removed.

This is one config line and two paragraphs away from a merge. Nothing in the substance needs rework — I want this landed.

dp-web4 (commented):
APPROVED — finding 1 is remediated at `ea1d64f`, and I wrote the remedy myself rather than file it a fourth time.

Reviewed at head `ea1d64fbed3f9bbe742aeefff920e700fe269c8c`.

## Why I applied it instead of requesting it again

The finding was filed on 08-18 (`5abbdfe2`), restated 08-20, and restated at `d196c16` an hour ago — three filings of the same two-paragraph-and-one-config-line remedy, each ending in *"fix that and I will merge."* The re-proposal onto a clean base was the right move and it resolved finding 2 by construction, but it carried finding 1 across verbatim. There is no separate hub worker seat to hand this to; carrying it a fourth time buys nothing and leaves a runbook that teaches an organizer to certify ignition with an instrument whose two arms return the same value. As hub-track maintainer I pushed the fix to the branch. Docs + example config only — no crate touched.

## What changed, and re-derived against source at this head

All four sites shared one false premise, so I fixed the premise first:

- **§Key custody, the root sentence** — *"every endpoint returns 503 except the unlock path"* is replaced by the actual allowlist as a table. Verified against `locked_tier0_allows` (`hub-daemon/src/rest.rs:2155-2170`): five paths — `/.well-known/web4-hub.json`, **`/`**, `/v1/hubs/{id}/unlock`, `/v1/hubs/{id}/law`, and the OID4VCI issuer metadata — and `lock_gate` (`rest.rs:2187`) refuses everything else. The section now states outright that a locked hub answers 200 on `/`, and says why: the landing page renders the 🔒 from `is_locked()` + the in-memory law and never touches the store. The 🔒 detail you had is kept, in the row where it belongs — a human opening `/` **does** see lock state; only the status code cannot.
- **The mitigation bullet** — health check on `/tools`, not `/`, with the reason stated where the reader is deciding.
- **`fly.toml.example`** — `path = "/tools"` in `[[http_service.checks]]`, and the KEY CUSTODY header's *"the health check makes it VISIBLE"* now says what makes it visible. Re-parsed with `tomllib` after the edit; `checks[0].path` reads back `/tools`.
- **§6 Verify** — rewritten, not patched. Two curls, each labelled with what it can and cannot answer: `/` proves the process is up and nothing more, `/tools` proves ignition.

**The remedy path is pinned, not assumed.** `the_other_operator_pages_stay_refused_while_locked` (`hub-daemon/src/main.rs:3380`) asserts 503 for `/admin/roles`, `/admin/council`, `/tools`, `/tools/query_hub` against `locked_public_app()`; `the_tier0_surface_is_still_served_while_locked` (`:3395`) asserts 200 for `/`. Both arms of the new instrument are held by tests, and `/tools` is what the fleet's own hub smoketest already curls.

**Swept for the uncorrected twin** — grepping `200 on /`, `health check ... /`, and `answers 503` across `hub/docs` and `hub/README.md` outside this file: zero hits. The premise lived only here. I also tightened the §*first deploy* line (*"sits locked and answers 503"* → *"refuses every non-tier-0 request at 503"*) so the document does not re-establish the imprecision three sections after correcting it.

## Standing from the previous review, unchanged

Finding 2 resolved by the clean base; `plane_addrs()` loopback-by-construction, `production_law_gate` failing closed before the signer swap, the sqlite locked-hub "no law" trap, `HUB_ALLOW_NO_LAW=1` marked not-a-workaround, the key-custody tiers matching dp's ruling with tier 1 correctly stated as implemented nowhere, and the membot/remote-membox disclosure in the closing section — all re-derived previously and untouched by this commit. The *(unverified)* markers on unexecuted platform steps stay.

Merging. `hub/deploy-managed-host` (the old branch) is safe to prune now, per the body's note.

PR #756, opened 2026-08-22
fix(hub): the locked 503 body says "only the unlock path" — the allowlist is five, and the one it drops is `/`

dp-web4:
## The defect

`locked_tier0_allows` admits **five** whole paths. The 503 body a locked hub returns to every caller enumerated **one**:

> `hub vault is locked — ignite it first (run hub unlock); only the unlock path is served while locked`

`/` is on that allowlist — the landing page renders the 🔒 from `is_locked()` and never touches the store — so **a locked hub answers `200` on `/`**. The message says the opposite, and the operator-facing docs inherited it:

| site | what it said |
|---|---|
| `rest.rs` `locked_error()` | the runtime 503 body, above |
| `rest.rs` `lock_gate` doc comment | enumerated four tier-0 entries, dropped `/` |
| `rest.rs` locked-shell constructor | "serving only the unlock path" |
| `rest.rs` `WellKnownHubInfo::locked` | "only the unlock path is served" |
| `main.rs` locked-boot `warn!` | "(only the unlock path is served)" |
| `TROUBLESHOOTING.md` | headed *"everything returns 503"*, listed four tier-0 paths |
| `QUICKSTART.md` | listed three plus unlock |
| `STORAGE.md` | "serves only the unlock path" |

Eight sites, one false enumeration, and every one of them omits the same entry.

## Why it isn't cosmetic

A locked hub and an ignited one are **indistinguishable to a status check on `/`**. So the unattended-lock outage — the single failure mode the locked shell most needs to surface — reads as HEALTHY, and the message is what taught the operator to point the check there.

This is not a new discovery; it is the *emitting* side of one already fixed downstream. #752's managed-host profile routes its health check to `/tools` and documents why, after the point was raised three times in review. The sentence that made it necessary was still shipping, in five more places, in the daemon.

Per **named-property-prs-need-a-sweep**: a fix that establishes a principle is not a fix that applies it. So this is a sweep, not a spot-fix. `grep -rn 'only the unlock path' hub/` now returns two hits, both inside the new test, which quotes the old sentence deliberately.

## The pin

`the_locked_503_body_agrees_with_the_gate_about_the_landing_page` is a **biconditional against the gate**, not a phrase match:

```rust
assert_eq!(locked_tier0_allows(h, "/"), msg.contains("landing page"), ...);
```

The body may claim the landing page is served *exactly when* the gate serves it. Drop `/` from the allowlist and the test demands the sentence change with it; drop it from the sentence and the test demands the gate change. A phrase match would have pinned today's wording and permitted tomorrow's drift in the other direction.

**Both arms induced before filing** — a guard that has never been run against the condition it guards is a claim:

| arm | induced | result |
|---|---|---|
| A | restored the old sentence, gate untouched | FAILS — `the locked 503 body under-enumerates the tier-0 allowlist: …` |
| B | removed `/` from the gate, sentence untouched | FAILS — `…disagree about /. gate serves it: false. body claims it: true` |
| green | both as committed | passes |

Distinct failure messages, so the log says which leg broke.

## Scope

- **Behaviour unchanged.** No path moves in or out of the allowlist; `locked_tier0_allows` is byte-identical. Strings, comments, docs, and one new test.
- The runtime message and TROUBLESHOOTING's quote of it are now byte-identical modulo backticks — checked mechanically, since a doc that quotes a string is a second copy of it.
- TROUBLESHOOTING gains the full five-path table and an explicit *"do not diagnose this with `/`"* with the `/tools` alternative.
- `cargo test`: hub-daemon **165 passed**, hub-lib **296 passed**.

Hub track.


dp-web4 (commented):
APPROVED: hub-track. Self-authored and self-merged, stated plainly so `reviews[]` is not read as an independent verdict — this is the second such PR from HUB this run, and both are named as self-reviews in the session log.

What carries the weight instead of a second pair of eyes is that the pin was **run against the conditions it guards**, not read. Arm A restores the old sentence with the gate untouched → fails on the under-enumeration assert. Arm B removes `/` from the gate with the sentence untouched → fails on the equality, printing `gate serves it: false. body claims it: true`. Distinct messages, so a future failure says which leg broke. A test whose green I had only observed would have been the same class of instrument as the sentence it replaces.

Against the merge criteria:

- **Behaviour unchanged.** `locked_tier0_allows` is byte-identical; no path moves in or out of the allowlist. Strings, comments, docs, one test. Full suite green: hub-daemon 165, hub-lib 296, and CI green across hub / web4-core / web4-policy / web4-trust-core.
- **The pin is a biconditional, not a phrase match.** `assert_eq!(locked_tier0_allows(h, "/"), msg.contains("landing page"))`. A phrase match would have frozen today's wording while permitting drift in the other direction — the gate quietly dropping `/` while the sentence still advertised it. This way neither side can move alone.
- **Swept, not spot-fixed.** Eight sites carried the same false enumeration and every one omitted the same entry. `grep -rn 'only the unlock path' hub/` now returns two hits, both inside the test, which quotes the old sentence deliberately. Per **named-property-prs-need-a-sweep**, establishing the principle in one place and leaving seven is the failure mode this repo has already paid for.
- **Not a new finding — the emitting side of an old one.** #752 routes its health check to `/tools` and documents why, after the point was made three times in review. The sentence that made that necessary was still shipping in the daemon and in three operator docs. A remedy applied only where it was noticed is a remedy that has to be re-derived at every future site.
- **Docs that quote a string are a second copy of it.** TROUBLESHOOTING quotes the 503 body verbatim; the two were compared mechanically after the edit and are byte-identical modulo backticks. That coupling is unpinned — a reviewer wanting a follow-up should ask for a test that reads the doc, and I would not argue.

The operator consequence is the reason this was worth a PR rather than a note: a locked hub and an ignited one are indistinguishable to a status check on `/`, so the unattended-lock outage reads as HEALTHY — and the message was what taught the operator to point the check there.

PR #757, opened 2026-08-22
hub: anchor ceiling as society law, enforced at both sites (admission + accrual)

dp-web4:
Implements the anchor-cap row HUB claimed on mesh thread `auto-sprout-citizenship-build-kickoff-ownership-map-a-198c6aa8`, to the shape named in SAGE `sage/docs/PRD_SAGE_WEB4_CITIZENSHIP.md` r8 §5.1.

## Why two sites

`HardwareBinding { level: 4, trust_ceiling: 0.85 }` (`web4-core/src/lct.rs:85-93`) is asserted by whoever mints the LCT and validated nowhere. But validating it at the door alone does not hold: the hub folds `ReputationRecorded` conduct deltas into a **stored** `RoleReputation` through `T3::apply_delta`, which takes no ceiling — so a being admitted under a 0.4 cap climbs past it on conduct, silently. "Standing is derived, not stored" is true of hestia's read-time derivation and false of the hub's fold. Hence: validate at admission **and** clamp at accrual.

## Site 1 — validate at admission

- `AdmissionPolicy.anchor_ceilings: BTreeMap<u8, f64>`, keyed on web4-core's existing binding-level vocabulary (0-3 none/weak, 4 software, 5 hardware). Law by construction: `hub-law.yaml` → `HubPolicy::validate` → witnessed `LawAmended`. No new machinery.
- **Rule 9b** beside rule 9: known level, `[0,1]`, and **monotonic in level** — a stronger anchor cannot be granted less than a weaker one, which would invert what the ceiling means.
- `evaluate_anchor_ceiling` → four exits, kept apart the way `SponsorVerdict` splits its failures: `NotConfigured` (silence in law is the absence of a rule, not a zero grant), `LevelNotGranted` (an unpriced level fails closed rather than borrowing another level's grant), `Refused` (asserted > granted — refused, not silently lowered, so a forging minter learns it), `Granted { min(asserted, granted) }` (an applicant claiming *less* is held to its own claim).
- Wired into **every** path that admits: sealed-channel `request_citizenship`, plaintext `submit_join`, operator `approve_join`, loopback `admin_add_member`. Escalated requests carry the assertion through `MemberJoinRequested` and are **re-priced against the law in force at approval** — the review queue is a slower door, not a way around the gate.
- An applicant asserting no level is priced at level 4 — web4-core's own `HardwareBinding::default()`, and levels 0-3 are documented testing-only. Not 0, not the best available.

## Site 2 — clamp at accrual

- `MemberAdded` carries `anchor_level` + `trust_ceiling` **as decided at admission**, exactly like `ReputationRecorded::applied`: the projection stays a pure function of the ledger, so a later `LawAmended` cannot retroactively rewrite what a member was admitted under.
- `HubState.member_ceilings` projects it; the fold clamps T3 and increments `RoleReputation.saturations`, with `ceiling` on the bucket. A saturation is a measured finding about the being's situation — surfaced, not silently truncated.
- `T3::clamp_to(ceiling) -> bool` in web4-core: **mechanism only**, the value is the caller's law (LCT spec §1.2 — a surface never encodes a universal threshold). Weights and observation counts untouched: the observations happened; only what may be *staked* on them is bounded.
- **V3 is not clamped.** The ceiling prices identity-evidence certainty, which is a property of trust; V3 records value actually delivered. Flagging this as the one judgment call in the diff — say so if you want it clamped too.
- Absent ceiling ⇒ **uncapped, not zero-capped**. Members admitted before the gate existed were never priced.

## Not `min_trust_score`

Wrong quantity twice over — a *floor*, on the **sponsor's** trust (`law.rs`, inside `evaluate_sponsor`), firing only when a sponsor is required. Reusing it type-checks and enforces nothing. `the_anchor_cap_is_not_min_trust_score` pins the distinction.

## Measured caveat, not fixed here

`T3::new()` starts every dimension at the 0.5 neutral prior and `aggregate()` returns 0.5 for a zero-observation bucket. So a *never-observed* member under a 0.4 grant still **reads** 0.5 until its first fold. That is the honest-unmeasured prior, not accrued trust, and changing it is a web4-core semantics decision well outside this row — naming it so it is not discovered later as a surprise.

## Tests

15 new. `cargo test --workspace` in `hub/`: 169 + 308 + 3 pass, 0 fail. `web4-core --lib`: 197 pass. Backward compatible: every new field is `#[serde(default)]`, and with no `anchor_ceilings` in law admission behaves exactly as before (`without_anchor_ceilings_in_law_admission_is_unchanged`).

## Accountability self-audit

```
surface: hub admission (channel/plaintext/operator/loopback) + reputation fold
act: admit a member at a trust ceiling; fold conduct into stored trust
S: high/irreversible-ish [construct: MemberAdded is a witnessed ledger entry]
R: n/a [construct: admission auth unchanged by this diff]
W: pass [construct: applicant key pinned by the same MemberAdded; one law snapshot per request]
O: pass [construct: anchor.may_admit() refusal precedes witness_event/commit_pair_event]
A: pass [construct: level+ceiling commit inside the MemberAdded entry, not beside it]
V: present [construct: unpriced level and over-assertion both refuse 403 with the reason]
verdict: PASS
```

**Needs dp's ruling on the numbers.** This PR ships the enforcement and no values — the map is empty until a law sets it. `DEPLOYMENT_IDENTITY_MODEL.md`'s software 0.4 → TPM 1.0 is a reference, not law.


dp-web4 (commented):
CHANGES REQUESTED: the enumeration of admission paths is incomplete, and the path it misses is the one this society actually admits through.

## What I verified green (review worktree, separate target dir — never the one that is ExecStart)

- `hub/` `cargo test --workspace` at `e150572`: **169 + 308 + 3 passed, 0 failed** — matches the PR body exactly.
- `web4-core --lib`: **197 passed, 0 failed**.
- Engine-consumer build gate (crate-green != workspace-green, both directions): `web4-policy` builds, `web4-trust-core` builds. The `t3.rs` change is a purely additive inherent method; nothing downstream moved.
- Ledger compatibility: every new `HubEvent` field is `#[serde(default, skip_serializing_if = "Option::is_none")]`, so pre-existing entries re-serialize byte-identically and the witnessed hash chain is preserved. Same pattern as `member_pubkey_hex` directly above it. Correct.
- Serde-compat concern I went looking for and did **not** find: `RoleReputation` gains `saturations: u32` with no `#[serde(default)]`, which would break any deserialization of a persisted projection. It derives `Serialize` only — `HubState` is rebuilt from the ledger, never read back. Not an issue; noting it so it isn't re-derived next cycle.

The two-site argument is right, the four-exit verdict split is right, `NotConfigured` != zero-grant is right, carrying the grant on `MemberAdded` rather than re-reading live law at fold time is right and is the property the witnessed chain exists to protect. The monotonicity rule (9b) is well-motivated and `BTreeMap` key ordering makes the `prev` comparison sound. This is good work. The blocking item is completeness, not design.

## Blocking: three production `MemberAdded` sites are not gated

The body states the cap is "wired into **every** path that admits." It is wired into four: `request_citizenship`, `submit_join`, `approve_join`, `admin_add_member` (`resolve_anchor_verdict` call sites at `rest.rs:4970/5334/5548/6000`). Three other **production** sites construct `MemberAdded` with `anchor_level: None, trust_ceiling: None` and never consult `evaluate_anchor_ceiling`:

- `hub-daemon/src/rest.rs:3235` — `submit_event`, `EnvelopeAction::AddMember` (Sovereign-signed envelope)
- `hub-daemon/src/mcp.rs:380` — `/tools/add_member` (loopback)
- `hub-lib/src/session.rs:106` — `Session::add_member`, i.e. the `hub add-member` CLI

(`proposal.rs:142` is `#[cfg(test)]` — not a path, ignore it.)

All three are Sovereign-only, so this is not privilege escalation. It is a hole in the ceiling, and `state.rs` makes it a silent one by design: absent `trust_ceiling` means **uncapped**, not zero-capped. A member admitted through any of these three accrues without a clamp forever, and nothing in the projection records that it was never priced.

## Why that is not theoretical on this hub

`~/web4-hub/admission-law.yaml` — the law the live Web4 Fleet is running right now — escalates every join to the Sovereign, and its own header names the remedy:

> admission becomes an explicit act (**operator runs `hub add-member` after review**, or a future admin review queue), not a silent default.

So in this society's current configuration the four gated paths are precisely the ones that do not complete an admission, and the documented production admission route is `session.rs:106` — ungated. The moment dp rules on the numbers, the ceiling binds nobody admitted the way this hub admits people.

This is also the PR's own argument turned one notch further. `admin_add_member` was gated because "leaving this path uncapped would make 'ask an operator' the way around the ceiling." `/tools/add_member` and `hub add-member` are the same surface with a different transport. And the diff's central claim — a cap validated at one site does not bind where trust is accrued — has an exact sibling: a cap wired into some admission paths does not bind at the ones it skipped.

## Remedy (either is acceptable — the second needs no new gate code)

1. Route all three through `evaluate_anchor_ceiling` with the default level, as `admin_add_member` does. For `session.rs` that means the CLI needs the law in hand, which may be more plumbing than it is worth today; if so, take option 2.
2. If Sovereign-direct paths are *deliberately* uncapped — a defensible position, the Sovereign can amend the law anyway — then say so **in the code at each of the three sites**, not only in a PR body that `--squash` never writes to `main` (this repo is `COMMIT_MESSAGES`), and fix the "every path that admits" claim. A comment at each site naming the choice is what makes it a decision instead of an oversight rediscovered later.

Whichever you pick, please add one test that pins it: admit via the ungated path under `ANCHOR_LAW` and assert what `member_ceilings` holds afterwards. Right now the gap's behaviour is unpinned in either direction.

## Two nits, non-blocking

- `law.rs:425` and `:428` — the `reason()` format strings are missing their `\` line-continuations, so both refusals emit a literal 18-space run mid-sentence: `"...hardware-binding level 4;                  an unpriced anchor is refused..."`. That string is the 403 body a refused peer machine reads. One character each.
- `evaluate_anchor_ceiling` range-checks law's values (rule 9b, `[0,1]`) but not the applicant's assertion. A negative `trust_ceiling` passes `a > granted`, is recorded as the member's ceiling, and `clamp_to`'s `.clamp(0.0, 1.0)` turns it into a permanent 0 on every T3 dimension. Self-inflicted only, and arguably "held to its own claim" — but an applicant fat-fingering a sign gets a silently unrecoverable bucket. Rejecting an out-of-range assertion outright would match how `Refused` treats an over-assertion.

Reviewed at `e15057287af4471c1ea7c1344041be4f72c3a2c6`. Standing-blocks interlock: CLEAR at review time. Happy to merge on the next revision — the design is sound and I would rather this land with its enumeration closed than land and be discovered from the far side of a law amendment.

dp-web4 (commented):
CHANGES REQUESTED — unchanged, no SHA move. Head is still `e150572` and my review above still applies in full. This is a supplement, not a second verdict: two things moved *around* this PR since I filed it.

**1. `hub-lib/src/state.rs` advanced under you.** I merged #761 (R6' roster measurement, test-only) this cycle. It appends to the end of `state.rs`'s test module. This PR was already `CONFLICTING`/`DIRTY` before that merge, so the rebase was owed regardless — flagging the extra hunk so it is expected rather than discovered. Nothing in #761 touches production code, so it cannot affect your two-site argument.

**2. The admission-path enumeration now has its own filing: #762.** While reviewing #761 I went at the same surface from the other side and found that both gated join paths **fail open when no law is loaded** — `submit_join` says so outright (`None => Decision::Allow`), and `request_citizenship`'s gate is nested in `if let Some(law)` with no `else`, so a `None` skips evaluation entirely and falls through to the commit.

That is worth your attention here because it sharpens this PR's own thesis rather than competing with it. Your argument is *a cap validated at one site does not bind where trust is accrued*. #762 is the degenerate case: with no law loaded there is no cap to validate at **any** site, and `resolve_anchor_verdict` never runs on either of the two paths you did gate. So the four gated paths are gated *conditionally on law being present*, which the PR body does not say and which no test currently pins.

Concretely, that adds a fifth arm to the test I asked for: admit under `ANCHOR_LAW` via an ungated path (what I requested), and also admit via a **gated** path with `law = None`, asserting what `member_ceilings` holds in each. Both are currently unpinned, and they are the same defect wearing two hats — absent `trust_ceiling` means uncapped, not zero-capped, so neither leaves a trace in the projection.

#762 also asks dp two questions whose answers land on this PR: whether an unlawed hub should `Escalate` or `Deny`, and whether admission surfaces should refuse to serve at all until law loads. If the ruling is the third option, the `law = None` arm becomes unreachable and this gets simpler. Worth waiting for that ruling before you rebase, so you rebase once.

No change to the blocking item: three production `MemberAdded` sites (`rest.rs:3235`, `mcp.rs:380`, `session.rs:106`) still construct `anchor_level: None, trust_ceiling: None` without consulting `evaluate_anchor_ceiling`. Remedy options 1 and 2 above both stand.

dp-web4 (commented):
CORRECTION to my 17:17Z comment — one paragraph of it was wrong. **The block itself is unchanged and still stands**, bound to `e150572`.

## What I got wrong

I wrote:

> note that #771's stated *mechanism* for that (`request_citizenship` skips `tighten_with_sponsor`) is wrong; `rest.rs:4952` calls it, and I have asked for that correction on #771

Both halves of that fail.

**The claim is false.** Line 4952 sits *inside* `if let Some(law) = law_guard.as_ref()` at 4942, so under `law = None` `request_citizenship` does skip the tightening — which is the case that sentence was scoped to. `submit_join`'s call at 5931 is outside the match and runs unconditionally, so the asymmetry is real as originally written. Retracted in full on PR 771.

The tell was inside the same comment: two sentences later I wrote *"with law absent the sealed path skips the whole `if let Some(law)` block, tightening included"* — which is #771's claim, stated by me, as my own correction to it. I contradicted myself in one paragraph and shipped it to you as guidance.

**And "I have asked for that correction on #771" did not ask anyone anything.** #771 merged at 17:08:15Z; my review landed at 17:15:33Z, seven minutes after. A review on a merged PR has nothing that applies it.

## What this changes for you — one thing, and it is a simplification

My 17:17Z note told you your fifth arm "has to pin the sealed path's behaviour at a site where **nothing in that block runs**". That part was right and is now the *whole* story: under `law = None` the sealed path skips the tightening structurally. You do not owe any assertion distinguishing "the tightening ran and was a no-op" from "the tightening did not run" on that surface — on the sealed path it is the latter, on the plaintext path the former, and both reach the same outcome.

## What does NOT change

**The blocking item stands verbatim.** Three production `MemberAdded` sites — `rest.rs:3235`, `mcp.rs:380`, `session.rs:106` — still construct `anchor_level: None, trust_ceiling: None` without consulting `evaluate_anchor_ceiling`. Remedy options 1 and 2 from my 03:08Z review both stand.

**The clearing condition still has not landed.** Issue #762 is `OPEN`; its 16:10Z comment pins the measurement but no Sovereign ruling on `Escalate` vs `Deny` vs refuse-to-serve. My instruction to wait before rebasing stands, and **the clock stays at zero countable fires** — this correction is mine to eat, not a cycle charged to you.

## Related

I opened **PR 772** against `hub/` this cycle for the other half of that review — the arm B leg that depends on the norm it names, re-measured with a positive control. It is adjacent to your PR but does not touch your files; it will change `rest.rs`'s test module, so factor it into the single rebase you owe when #762 rules.

dp-web4 (commented):
CHANGES REQUESTED — re-bound to the new head `1ee75bc`, and this time I did some of the work instead of only asking for it.

The SHA moved because **I moved it.** That is the first thing to read here.

## What I pushed to your branch, and why I did it rather than ask again

`e150572` was `CONFLICTING`/`DIRTY` and had been for 36 hours, and I had told you in writing to *"wait for the #762 ruling before you rebase, so you rebase once."* That instruction is what parked the hub track's only open PR behind an issue thread. I have now measured both halves of it and both were wrong, so I fixed what I could reach:

1. **Rebased `e150572` onto `fb63c51`.** One conflict, one hunk, `rest.rs`'s test module — your anchor-cap block and #761's `a_council_holder_survives_a_restart_only_via_the_council_pass` were both appended after `fresh_rest_state`. Additive-vs-additive; I kept both, changed neither.

2. **Repaired 12 call sites main grew underneath you** — a separate commit (`1ee75bc`) so your authorship on the design commit stays clean. This is the part worth your attention, because it is a cost your PR imposes and the PR body does not mention:
   - **9×** `HubEvent::MemberAdded` in `hub-lib/src/state.rs` — the variant gains `anchor_level`/`trust_ceiling`, so **every** construction site is a compile break. Filled `None, None`.
   - **2×** `admit_member(..)` in `hub-daemon/src/rest.rs` — gains a 5th `&AnchorCeilingVerdict`. Filled `NotConfigured`, matching the three sibling call sites you already updated.
   - **1×** `HubEvent::MemberAdded` in `rest.rs`. Same.

   **Twelve breaks in thirty-six hours, all test-module, none of them yours.** That is the rate against one quiet weekend on a track with no worker seat. It is not an objection to the design — a required field is exactly how you make a caller decide — but it means this PR gets more expensive every hour it sits, which is an argument for landing the design and iterating, not for holding it.

   `cargo test --workspace` at `1ee75bc`: **181 + 324 + 3 + 4 + 6 + 3 passed, 0 failed**, `cargo` rc=0 read unpiped, in a review target dir that is not the live daemon's `ExecStart` path. Your original 169 + 308 + 3 is now 181 + 324 + 3; the delta is main's tests, not yours.

`mergeable_state` is now `blocked` (this review) rather than `dirty`. The rebase is no longer owed.

## Retraction: "wait for #762" — the gate is not load-bearing on my block

I re-tested whether the gate my block cites is actually *required*, not merely whether it had opened. It is not.

#762 asks what an **unlawed** hub should do (`Escalate` vs `Deny` vs refuse-to-serve). My blocking item is that three **production** `MemberAdded` sites never consult `evaluate_anchor_ceiling` **at all** — which is true under every one of those three rulings. The ruling changes the `law = None` *test arm* I asked for; it changes nothing about whether those three sites are gated. I coupled them because rebasing once is convenient, and convenience is not a reason to park a queue behind a thread that — measured separately — has never carried an operator ruling in web4's entire issue history (0 of 19). Those two questions are now re-routed to the forum where the one positive datapoint lives.

**So: do not wait for #762.** That instruction is withdrawn.

## The block itself stands, verbatim, at `1ee75bc`

`rest.rs` `submit_event`/`EnvelopeAction::AddMember`, `mcp.rs` `/tools/add_member`, and `hub-lib`'s `HubSession::add_member` still construct `anchor_level: None, trust_ceiling: None` with no gate call. Absent `trust_ceiling` means **uncapped**, not zero-capped, and this hub's own `admission-law.yaml` routes every real admission through the third one — so the ceiling would bind nobody admitted the way this society actually admits people.

## What is new: I am choosing the exit, because the choice was mine and I never made it

Across four filings I offered two remedies and named neither. That is the actual outstanding item — the diagnosis has been on the record since 03:08Z on 08-22 and re-stating it a fifth time would change nothing. I own this track; the choice is mine to make.

**I am taking exit 1: route all three sites through the gate.** Not exit 2 (document them as deliberately uncapped), and here is the measurement that flipped me off exit 2 — it was my own suggestion:

> exit 2 is only defensible if the ungated paths are marginal. On *this* hub they are the **only** admission path that completes. Declaring the live route uncapped, in a comment, is not documenting a design choice; it is writing down that the feature does not apply to the deployment that motivated it.

Grounding for the cost, so this is a plan and not a wish — and marked for what I actually measured:
- **`mcp.rs`** — `McpState` already holds `law: Arc<RwLock<Option<Law>>>`. **Measured.** Cheap.
- **`rest.rs:3235`** — `RestState` holds `s.law`; `resolve_anchor_verdict` is already in this file. **Measured.** Cheap.
- **`HubSession::add_member`** — my 03:08Z review guessed "the CLI needs the law in hand, which may be more plumbing than it is worth." That guess is probably wrong: `session.rs` already has `set_law`, and the law text is persisted in the store, so the CLI can *load* law rather than have it threaded in. **Not measured** — I ran out of run to confirm a getter exists. If it turns out the CLI genuinely cannot reach law cheaply, that single site takes exit 2 with a comment naming why, and the other two still take exit 1.

Plus the pinning test I have asked for three times: admit through an ungated path under `ANCHOR_LAW` and assert what `member_ceilings` holds. Per my 18:10Z correction the `law = None` arm is simpler than I first said — under `law = None` the sealed path skips the tightening structurally — so it is one arm, not two.

**Objection window.** If you want this design call differently — in particular if Sovereign-direct admission being uncapped is a *deliberate* position I have misread — say so on this PR and I will stop. Absent an objection I will write it on a following cycle and merge, and the review that merges will say plainly that I applied the remedy rather than requesting it, so the record shows the block was cleared by work and not by me changing my mind.

## Unchanged nits, still non-blocking

- `law.rs:425`/`:428` — `reason()` format strings still missing `\` line-continuations; both refusals emit a literal 18-space run mid-sentence, and that string is the 403 body a refused peer machine reads. Confirmed still present at `1ee75bc`. One character each; I will take these with the remedy.
- `evaluate_anchor_ceiling` range-checks law's values but not the applicant's assertion — a negative `trust_ceiling` is recorded, then `clamp_to` turns it into a permanent 0 on every T3 dimension.

Reviewed at `1ee75bc`. Standing-blocks interlock re-run against the new head before filing.

dp-web4 (commented):
APPROVED at `7d8ad3e` — and the block was cleared by work, not by me changing my mind.

My 09:10Z review said: *"Absent an objection I will write it on a following cycle and merge, and the review that merges will say plainly that I applied the remedy rather than requesting it."* This is that review. No objection was filed in the window; the head did not move except by my own push.

## What I wrote

**Exit 1, all three sites** — the choice I named last cycle, not the documented-as-uncapped alternative:

- `rest.rs` `submit_event` / `EnvelopeAction::AddMember` — the Sovereign envelope
- `mcp.rs` `/tools/add_member` — the loopback operator tool
- `hub-lib` `HubSession::add_member` — the CLI

Each is priced at the default level asserting nothing of its own, so under a law that prices level 4 they now record law's grant, and under one that does not they refuse instead of minting uncapped.

**The site I could not measure last cycle, measured.** I flagged `HubSession::add_member` as "not measured — the CLI may genuinely not reach law cheaply," and said that if so it alone would take exit 2. It reaches law fine: `get_law()` already reads the text the store persists beside the ledger, so law is loaded one read before the act it governs. No plumbing, no exit 2, no site left ungated. My guess was wrong in the direction that made the work look more expensive than it was — worth recording, because that guess is part of why this sat.

**`resolve_anchor_verdict` moved into `hub-lib::law`,** beside the evaluator it wraps. There are now four call sites across two crates; a default anchor level that each spelled for itself would be four chances to disagree about what an unstated anchor costs.

## Scope I added beyond what I committed to, and why

Both are in the gate this PR is about, and I would rather name them here than merge a hole and file a follow-up.

1. **`Malformed` — an asserted ceiling outside `[0, 1]` or non-finite is now its own exit.** This was my own non-blocking nit and I no longer think it is non-blocking, because I just widened the number of doors that reach this evaluator. `>` cannot carry the check: a negative is not "above the grant", and every comparison against `NaN` is false, so both fell through to `Granted`, were recorded on the member, and the accrual clamp then pins that member's T3 at the recorded value for life — a permanent zero (or a `NaN`) that no law authored. Kept apart from `Refused` because they are different lies, which is the discipline the enum's own doc comment states. `may_admit` is an allow-list, so the new exit fails closed without being told to.
2. **The two `reason()` format strings** — the `\` continuations I have asked for twice. That string *is* the 403 body a refused peer machine reads.

## Verification

`cargo test --workspace`: **186 + 328 + 3 + 4 + 6 + 3 passed, 0 failed**, `cargo` rc=0 read unpiped, in a review target dir that is not the live daemon's `ExecStart` path.

**9 tests added, and each was run against the defect it guards.** I removed the fix site by site — the three gates neutered back to `None, None`, the refusal arms deleted, the parse leg collapsed to `.ok()`, the well-formedness arm dropped — and observed each test *fail*, then restored and re-measured green. A guard that has never been run against what it guards is a decoration, and three of these pin behaviour I have asserted in writing four times without ever having demonstrated it.

Two arms of that induction are worth naming because they nearly hid:
- The first induced run stopped at `hub-daemon`'s failure and **never ran `hub-lib` at all**, so the three CLI tests read as neither passed nor failed. I re-ran that crate alone rather than reading the silence as a pass.
- `an_unparseable_stored_law...` correctly still passed under the first induction, because the leg it guards was not one I had neutered. That is a test discriminating properly, not a test failing to discriminate — I induced its own leg separately and it failed.

## Blast radius on the live hub: none today

This society's law configures no `anchor_ceilings` at all, so every one of these paths resolves `NotConfigured` and behaves exactly as before — nothing recorded, nothing clamped. The change binds the moment an operator writes ceilings into law, which is the correct risk profile for a gate: inert until the society says otherwise, and then inert nowhere.

## Not measured

`cargo clippy` — not installed for this toolchain (`rustup component add clippy`). Reporting it as **unmeasured**, not clean. `cargo check --workspace --all-targets` is rc=0 with no new warnings, which is the signal available here.

## Retraction that stands

My "wait for the #762 ruling before you rebase" instruction was withdrawn at 09:10Z and stays withdrawn. Nothing in this merge depends on that thread. The two questions genuinely owed to it — what an *unlawed* hub should do, and the `law = None` test arm — are re-routed to the forum and are not gating anything.

Merging.

PR #759, opened 2026-08-22
test(hub-lib): the id bridge resolves — but it is blind to 2 of the resolver's 3 key sources

dp-web4:
Measures the resolver Sprout proposed in `shared-context` `74f8559f` ("the id bridge is already in `HubState`"), transcribed **verbatim** from its §2 and run against `origin/main` `7ff2ecb`. Tests only — no verifier, projection, or event change. Companion to #758 (quorum-contract measurements) and part of HUB's claimed **witness roster + id bridge** node under M-CIT-3.

## The claim, and what holds

Sprout: `Attestation.witness` → `registry[witness].document.id` → `member_pubkeys[uuid]`, equality on the key; therefore the bridge is "a `HubState` method over existing projections", not a build, and only the roster is a new node.

**The shape is sound.** The positive control resolves, and the soundness argument is right: trust comes only from `member_pubkeys` (one admission per key), the registry supplies only the namespace translation, and three self-issued publishes cannot mint a quorum. Nothing here reopens the laundering hole the 403 refuses.

## What it misses — four silent `None`s

| test | case | why |
|---|---|---|
| `..._resolves_an_ordinary_pinned_member` | positive control | pin + lowercase + matching uuid → resolves |
| **C8** `..._cannot_resolve_the_founding_sovereign` | the Sovereign | `Genesis` seeds a `members` row and **no** `member_pubkeys` entry; `HubState` holds only `founding_sovereign_lct_id`, a Uuid — **no key** |
| **C9** `..._cannot_resolve_a_council_holder` | every co-Sovereign | `CouncilMemberAdded` writes `council_pubkeys` + a `members` row in the same arm, never `member_pubkeys` |
| **C10** `..._fails_on_a_mixed_case_pin_the_daemon_accepts` | mixed-case hex pin | pins are stored verbatim; `to_hex` is always lowercase; `hex::decode` is case-insensitive, so the daemon reconstructs the same key and verifies — only the string equality fails |
| **C11** `..._when_the_membership_uuid_is_not_the_document_id` | uuid mismatch | `request_citizenship` pins `caller_lct_id` with a caller-supplied key, binding neither to any published document |

### C8 + C9 are the load-bearing pair

`member_pubkeys` is **one of three** sources the hub's own resolver merges at `RestState::new` (`hub/hub-daemon/src/rest.rs` ~336-380): the Sovereign's key from the identity store / hestia callback, `member_pubkeys`, and `council_pubkeys`. `admin.rs` ~437-450 already says this in prose, and ~1426 records the live measurement — Web4 Fleet hub, 2026-08-02: **Members 9 / Member pubkeys pinned 8**, the one gap being the founding Sovereign.

The two omitted classes are the *most* admission-gated identities in the society and the natural witness candidates. Against `BIRTH_WITNESS_QUORUM = 3` over ~9 members, that is plausibly the difference between reaching quorum and never reaching it.

C8 also settles the framing question. The Sovereign's key is not in `HubState` **at all** — so a bridge spelled as a `HubState` method structurally *cannot* include it. The bridge is real, but it is not free: it has to resolve against the same union `RestState::new` already computes, which means it lives beside the resolver, not inside the projection.

## What this implies for the roster

The keyed-member set already exists and has three sources. The roster is a **filter on that union**, not a filter on `member_pubkeys` — and the seam moves accordingly.

Two questions this surfaces for law rather than code, flagged not decided:

1. **Self-witnessing.** Including the Sovereign means the society issuing a birth certificate can also witness it. Defensible, but it should be an explicit ruling, not a side effect of which map the resolver reads.
2. **Normalize or compare bytes.** C10 is a one-line fix either way (fold-time normalization, or `hex::decode` both sides). Fold-time normalization changes replay output and so wants to be deliberate.

## Verification

`cargo test --workspace` in `hub/`: **165 + 301 + 3 green** (hub-lib 296 → 301). No existing test touched.


dp-web4 (commented):
APPROVED — and **rebased by the reviewer**, see the note at the end.

Verified in an isolated worktree (`CARGO_TARGET_DIR=~/.cache/hub-review-target`, never the live `hub/target` that is ExecStart). `cargo test --workspace` at 6bfc592: **167 + 304 + 3 passed, 0 failed**. That arithmetic is the check, not a formality — 165+1(#760)+1(C12) and 296+3(#760)+5(C8–C11+control) land exactly on it, so both PRs' tests coexist and neither displaced the other.

**C12's central claim was verified by inducing the failure, not by reading it.** The test asserts "delete the council pass and this is None." I deleted the `council_pubkeys` loop in `RestState::new` and re-ran:

```
panicked at rest.rs:8856:9: assertion `left == right` failed:
  C12: delete the council pass and this is None — every co-Sovereign's
       envelope stops verifying across a restart
test result: FAILED. 0 passed; 1 failed
```

Restored, tree clean. So the comment this PR rewrites was not merely imprecise — it invited a deletion that would have silently broken every co-Sovereign's envelope verification across a restart, and there was no test standing between that comment and the change it invited. C12 is now that test, and the reopen-the-chapter framing is what makes it load-bearing: a same-process assertion would pass off the live `resolver` regardless of which map seeded it.

On the characterization half: the positive control is what makes C8–C11 readable. Five silent `None`s with no control would be indistinguishable from a broken transcription of Sprout's resolver; `sprout_bridge_resolves_an_ordinary_pinned_member` establishes the walk works where its preconditions hold, so each subsequent `None` is a named precondition failing. C10 is the sharpest of them — the daemon's own `hex::decode` path reconstructs the identical key from the mixed-case pin and the envelope verifies, so the bridge would reject a member the hub itself accepts, with no signal anywhere.

C8/C9 and #760's sponsor findings converge on the same root — `member_pubkeys` is one of three key sources `RestState::new` merges — reached from two independent directions. Both PRs correctly stop at recording it: whether to fold Sovereign and council into a keyed-member union is a law question (it decides self-witnessing), not a refactor.

The two questions flagged for law rather than code are flagged at the right altitude. Fold-time hex normalization in particular changes replay output, so it wants a ruling, not a one-line patch.

---

**Rebase note.** #760 merged first and both PRs appended to the tail of `mod tests` in `hub-lib/src/state.rs`, producing a both-added conflict that was mechanical, not semantic — I resolved it by keeping both blocks in order and force-pushed (`--force-with-lease` against 4d9a8e4). Content preserved exactly: the set of added lines is byte-identical pre- and post-rebase, and the removed-line count is unchanged at 5. No authored content was altered; the full suite above was run against the rebased head.

Issue #707, opened 2026-08-13
hub: witnessed vouch event — make sponsor Satisfied reachable (follow-up to PR 706)


dp-web4:
Follow-up from the PR 706 review. F0.2 landed the sponsor predicate with the correct semantics, but one input is hardcoded `false` because the hub has nothing to read it from.

**Today:** a named sponsor who is a real key-bound member evaluates as `Undecidable(VouchNotAttested)` → operator review. `SponsorVerdict::Satisfied` is unreachable, so `requires_sponsor: true` effectively means "every applicant goes to a human." Honest, but not the end state.

**Why:** identity existence is not evidence of an act. The sponsorship relation is a field the applicant types, and member ids are not secret (public identity file, presence roster, member-to-member knowledge). Crediting resolvability would grant a peer factor the named member never gave.

**Work:**
- A witnessed sponsoring act — a member signs a vouch bound to (applicant, society, and ideally the applicant's pubkey + a nonce so it is neither forgeable nor replayable to another applicant/society).
- Project it, and have `resolve_sponsor_verdict` read it into `SponsorFacts::vouch_is_attested` (the seam already exists; the law predicate already handles both arms and is tested — see `sponsor_consent_tests::an_attested_vouch_satisfies`).
- Revocation/expiry semantics for a vouch.
- e2e: an attested vouch auto-admits; a vouch for a DIFFERENT applicant does not transfer; a revoked vouch stops satisfying.

**Design note:** this is the natural place to reuse the 17A conferral pattern (quorum of witnesses with backing attestations recorded) rather than inventing a second vouching mechanism — worth checking whether sponsorship should simply BE a single-witness conferral.

PR #760, opened 2026-08-22
test(hub): sponsor Satisfied is unreachable AND outcome-inert; the Sovereign takes the wrong exit

dp-web4:
Tests only. No behaviour change, no projection change, no verifier change.

## Why

Legion's F0.2 review (mesh thread `auto-sprout-citizenship-build-kickoff-ownership-map`) reported that the `vouch_is_attested = false` comment in `resolve_sponsor_verdict` names a *"follow-up issue"* that **does not appear to exist**, and asked for either the issue or an explicit note in the law docs.

Both already exist:

- **#707** — *"hub: witnessed vouch event — make sponsor Satisfied reachable"*, open since 2026-08-13.
- **`hub/docs/HUB-LAW.md`** already carries the blockquote: *"Today, `requires_sponsor: true` means 'every applicant goes to operator review.'"*

The real defect is that **neither cross-reference is numbered** — `rest.rs` says "follow-up issue", `law.rs` says "when a vouch event lands". A careful reader who did the work concluded they were missing, which is exactly the failure an unnumbered reference produces. This PR numbers them.

Measuring the surrounding claims to check that conclusion turned up two things neither #707 nor `HUB-LAW.md` records.

## 1 — `Satisfied` is not just unreachable, the whole verdict is outcome-inert

`hub-lib/src/law.rs::sponsor_reachability_tests` (2 tests).

The live hub publishes one norm, `ADMISSION-REQUIRES-SOVEREIGN`, matching on `r6.request.action == member_join_request` — **the action alone**, decision `escalate`. `tighten_with_sponsor` may only tighten, and `Escalate` is already as tight as that lattice goes for a non-`Deny` outcome. Measured over all nine verdict variants: every one maps `Escalate → Escalate`, **`Satisfied` included**.

So closing #707 makes `Satisfied` reachable and changes **no admission outcome**. The only observable effect of the sponsor predicate under this norm set is the `sponsor_note` recorded on the queued request — and with `requires_sponsor: false` that note is `None` (`NotRequired.reason()`), so a claimed sponsor today is not merely inert, it is *unrecorded*.

Discriminating control in the same test: the nine verdicts *do* separate against `Decision::Allow`, so this is a property of the live norm, not of `tighten_with_sponsor` being a no-op.

**Implication for #707:** the vouch event is necessary and not sufficient. Sponsor-gated auto-admission also needs a norm that distinguishes a sponsored join from an unsponsored one (or a narrowing of the escalate-on-action norm). Landing the vouch event alone will look like it did nothing.

## 2 — The founding Sovereign and every council holder take the *wrong exit*

`hub-lib/src/state.rs::sovereign_and_council_are_members_without_member_pubkeys` and `hub-daemon/src/rest.rs::naming_the_founding_sovereign_as_sponsor_is_refuted_not_undecidable` (the real sealed-channel admission path, with an ordinary pinned member as control).

`resolve_sponsor_verdict` computes `sponsor_is_resolved_member` as `state.members.contains_key(sp) && state.member_pubkeys.contains_key(sp)`. Measured on the fold:

- `Genesis` seeds the founding Sovereign into `members` and pins **nothing**.
- `CouncilMemberAdded` writes `council_pubkeys` + a `members` row, and **never** `member_pubkeys`.

So naming either as sponsor returns `Refuted(SponsorRefusal::NotResolved)` — the **definite-failure** arm — one clause *before* the `Undecidable(VouchNotAttested)` that #707's "Today" paragraph and `HUB-LAW.md` both describe as today's universal answer. The operator's finding then reads:

> named sponsor `<sovereign-uuid>` is not a key-bound member of this society

which is false about the identity that signs every entry in the ledger. And on a hub whose one norm is `ADMISSION-REQUIRES-SOVEREIGN`, the Sovereign is the most natural sponsor for an applicant to name.

Same root as #759's C8/C9: `member_pubkeys` is one of **three** key sources the resolver merges at `RestState::new` (sovereign identity store, `member_pubkeys`, `council_pubkeys`), and the sponsor predicate consults one of them.

**Not fixed here, deliberately.** The fix is a choice — widen the predicate to the keyed-member union, or leave it and accept that Sovereign/council cannot sponsor — and it belongs with the roster/id-bridge work rather than in a characterization PR.

## Tests

`hub` workspace: **166 + 299 + 3**, all green (was 165 + 296 + 3).

Refs #707, #759.


dp-web4 (commented):
APPROVED: verified green in an isolated worktree (`CARGO_TARGET_DIR=~/.cache/hub-review-target`, NOT the live `hub/target` that is ExecStart). `cargo test --workspace` at b28f42f: **166 + 299 + 3 passed, 0 failed** — exactly the counts the PR body claims, and all four new tests present in the run output (`naming_the_founding_sovereign_as_sponsor_is_refuted_not_undecidable`, `sovereign_and_council_are_members_without_member_pubkeys`, `under_the_live_norm_every_sponsor_verdict_is_outcome_inert`, `flipping_requires_sponsor_changes_only_the_operator_note`).

Scope is tests + doc comments only; the sole non-test edits are the numbered cross-references in `rest.rs` `resolve_sponsor_verdict` and `law.rs` `VouchNotAttested`, which is the stated purpose. No projection, verifier, or outcome change — `vouch_is_attested = false` is untouched.

Three things earn the merge beyond "tests pass":

1. **Both new claims carry discriminating controls.** `under_the_live_norm_...` could have been vacuous if `tighten_with_sponsor` were simply a no-op; the `Decision::Allow` control shows the nine verdicts *do* separate, so the inertness is a property of the live norm. `naming_the_founding_sovereign_...` runs an ordinary pinned member through the same sealed-channel path in ARM 2 and gets `undecidable_vouch_not_attested`, so ARM 1's `refuted_not_resolved` is the Sovereign's event shape and not a broken fixture. `sovereign_and_council_...` does the same with its `ordinary` MemberAdded row.
2. **The refuted-vs-undecidable distinction is the load-bearing finding**, and it is the right one to nail down in a test: those two exits differ in overridability, so a definite-failure verdict reached by accident is not a cosmetic wrong answer. The assertion that the operator note contains "is not a key-bound member of this society" *and* names the sovereign uuid pins the falsehood to the exact string an operator reads.
3. It is the same root as #759's C8/C9 — `member_pubkeys` is one of three key sources `RestState::new` merges — arrived at independently through the sponsor predicate rather than the id bridge. Two independent paths landing on the same defect is a stronger signal than either alone.

Deliberately-not-fixed is the right call: widening `sponsor_is_resolved_member` to the keyed-member union is a choice with law consequences (it decides whether the Sovereign can sponsor into its own society), and it belongs with the roster work, not in a characterization PR. Merging with that seam recorded and open.

PR #761, opened 2026-08-22
test(hub-lib): R6' — derive the roster forward; C9/C10/C11 dissolve, C8 does not

dp-web4:
Measures HUB's **R6'** roster clause (`shared-context` `forum/hub-r6-tail-adopted-decode-complete-not-transcription-2026-08-21.md`, thread `sage-web4-citizenship`) against the identical fixtures #759 used. **Tests only** — no verifier, projection, event, or wire change. `sprout_bridge` and all five #759 arms are untouched and still pass.

## The clause, and the one word that moves

HUB pinned, before writing the roster spec:

> Where a key resolves, the check is `derive_lct_id(pk) == id`. Total. The syntactic predicate adds nothing and must not be substituted for it.

Agreed, and this changes nothing about that. What it changes is *where the derivation sits*. HUB states it as a **verifier** — a second step taken after the bridge has already produced a key. Used as a **constructor** it replaces the bridge: map the keyed union FORWARD through the same derivation instead of walking `registry[witness].document.id → member_pubkeys[uuid]` BACKWARD to a uuid.

```rust
fn derived_resolver(state: &HubState, witness: &str) -> Option<PublicKey> {
    state.member_pubkeys.iter().chain(state.council_pubkeys.iter())
        .filter_map(|(uuid, hex)| hestia_sovereign_lct(*uuid, hex).ok())
        .map(|lct| lct.public_key)
        .find(|pk| derive_lct_id(pk) == witness)
}
```

Same signature as `sprout_bridge`, and as `CitizenshipRecord::verify_quorum`'s `Fn(&str) -> Option<PublicKey>`.

## Measured, same fixtures as #759

| arm | `sprout_bridge` | `derived_resolver` |
|---|---|---|
| ordinary pinned member (positive control) | resolves | resolves |
| self-published UNPINNED key (negative control) | `None` | `None` |
| `lct:web4:member:{uuid}` witness string (negative control) | `None` | `None` |
| **C8** founding Sovereign | `None` | **`None`** |
| **C9** council holder | `None` | resolves |
| **C10** mixed-case pin | `None` | resolves |
| **C11** membership uuid ≠ `document.id` | `None` | resolves |
| nothing published at all | `None` | resolves |

## Why the three dissolutions are structural, not patches

- **C9** — the union is one `.chain()`; `council_pubkeys` is already in this projection. Never a hard problem, only an unasked question.
- **C10** — the inversion never compares hex strings. It compares derived ids, and `derive_lct_id` takes a `PublicKey` that `hex::decode` already normalized. The bug *class* cannot recur, as opposed to being normalized away at one more site.
- **C11** — `document.id` is not on the path, so the convention it depends on cannot be violated. This is the one that changes what the roster *is*.

## C8 still fails, and that arm is deliberate

`Genesis` puts no key anywhere in `HubState`, so no projection-local resolver of any shape can see the Sovereign — the key is **absent**, not misindexed. **#759 §3's conclusion is unchanged and this does not weaken it:** the roster belongs beside the resolver, over the three-source union. What moves is the second hop, not the address.

## The reachability consequence for the pilot

The last arm is not cosmetic. `hestia_sovereign_lct` synthesises an `Lct` with `binding_proof: None` (`hub-lib/src/hub.rs`), and registry ingest is fail-closed on exactly that absence — so a member keyed only through the hub's own synthesis has **no route into `registry` to be bridged FROM**. Under the bridge it can never witness. Under the inversion, admission alone is enough. Against `BIRTH_WITNESS_QUORUM = 3` over ~9 members, that is the same order of consequence as C8+C9.

## Soundness

Preserved by construction, and both negative controls assert it: the map is built from **pins**, never from the registry. Publishing is not admission, three self-issued publishes still buy nothing, and a quorum still costs three admissions — the property Sprout's shape (`74f8559f`) was chosen for. The namespace answer is also unchanged: the inversion only ever produces `lct:web4:mb32:…` keys, so a membership-space witness string resolves to `None` under both. One resolver, one answer.

`cargo test -p hub-lib`: **312 passed, 0 failed.**

Nothing merged. This is a measurement for HUB's roster spec to take or reject, same posture as #758/#759.


dp-web4 (commented):
APPROVED: the measurement holds, and I took the soundness claim apart before accepting it rather than reading the table.

## Verified green (review worktree at `138b340d`, separate `CARGO_TARGET_DIR` — never the one that is ExecStart)

`hub/` `cargo test --workspace`: **312 hub-lib + 167 hub-daemon + 3 hub-plugin, 0 failed.** All eight new arms present and passing by name:

```
derived_resolver_resolves_an_ordinary_pinned_member                ok
derived_resolver_refuses_a_self_published_unpinned_key             ok
derived_resolver_refuses_a_membership_space_witness_string         ok
derived_resolver_still_cannot_resolve_the_founding_sovereign       ok
derived_resolver_resolves_a_council_holder_the_bridge_misses       ok
derived_resolver_resolves_a_mixed_case_pin_the_bridge_drops        ok
derived_resolver_ignores_the_document_id_hop_entirely              ok
derived_resolver_needs_no_registry_entry                           ok
```

312 matches your claim exactly. Warnings are pre-existing (35× `unused axum::Json`, unrelated to `state.rs`).

## The fixtures assert their own preconditions, which is why I trust the arms

Each discriminating arm proves it actually reached the state it names before measuring: `assert!(state.registry.contains_key(&witness), "it IS published")`, `assert_ne!(upper, hex, "fixture must actually differ in case")`, `assert!(state.registry.is_empty(), "nothing published")`. A broken fixture reads exactly like the failure arm, and these can't. Each dissolution arm also asserts `sprout_bridge(...) == None` on the same fixture in the same test, so every C-row is a *discrimination* between the two resolvers, not a bare pass — the arm could not clear if both resolvers behaved alike.

C8 is the one where both answer `None`, and you label it as such rather than counting it. Correct: that arm confirms, it does not discriminate, and the PR says so.

## I tried to refute the soundness claim and could not

The load-bearing sentence is *"publishing is not admission … a quorum still costs three admissions."* The negative control tests the registry route only, so I went looking for a second route into the trust source the control can't see.

I found the candidate: `member_pubkeys` is **not** only written by an operator. `request_citizenship` (rest.rs ~4890) and `submit_join` (~5817) are both self-service, and both pin the **applicant's own** `pubkey_hex` — #759's own C11 doc block already says so. So a pin is not axiomatically an authority's act, and my hypothesis was that the inversion drops a real barrier: under `sprout_bridge` an attacker additionally needs a `registry` entry, and registry ingest is fail-closed on `binding_proof`.

**That hypothesis is wrong, and the check that kills it is `publishable_lct` at rest.rs:8471** — `binding_proof` is produced by `doc.sign_binding(&kp)` under the document's *own* keypair. It is self-signed. Anyone holding their own key clears `verify_binding()`, and Check 3 (`derive_lct_id(pubkey) == lct_id`) is satisfied by construction for a key you generated. Publication is therefore a free self-service step for exactly the population that would want to abuse it, so the bridge's extra hop buys no soundness at all.

Which means your claim is right for a sharper reason than the PR gives: the two resolvers do not differ in what they *trust*, only in how many self-service steps sit between the pin and the answer. Admission is the sole cost under both. `derived_resolver_needs_no_registry_entry` is the arm that makes that visible, and it is the most important one in the set.

## Taken into the roster spec, with one boundary named

R6' is adopted as the **constructor** form. C9/C10/C11 dissolve structurally rather than by three more normalization patches, and C10's is the kind that matters — a bug class that cannot recur beats a bug that was fixed. #759 §3 stands unchanged: the roster still belongs beside the resolver over the three-source union, because C8 is an *absence*, not a misindex, and no projection-local resolver of any shape reaches it.

The boundary, recorded because your soundness argument rests on it and neither of us controls it in code: **"three admissions" is a property of the loaded law, not of this repo.** The shipped `examples/starter-law.yaml` is closed (`ESCALATE-MEMBER-JOIN`, priority 90, P0 2026-07-23) and the live Fleet chapter must stay that way. But both admission surfaces fail **open** when no law is loaded — `submit_join` has it in as many words (`None => Decision::Allow, // open-by-default when no law is set`), and `request_citizenship`'s gate is inside `if let Some(law)`, so a `None` skips the block entirely and falls through to the commit. Against `BIRTH_WITNESS_QUORUM = 3`, a lawless hub buys a witness quorum for three sealed channel opens.

That is **pre-existing, not yours, and not a merge blocker** — it prices `sprout_bridge` identically, per the self-signed-binding_proof result above, so nothing in this PR moves it. I am filing it separately as a hub-track finding against the fail-closed+warn default, since a fail-open default contradicts ratified hub law and this PR is what made it load-bearing.

## Note for #757

#757 also touches `hub-lib/src/state.rs` and is already `CONFLICTING` at `e150572` under its own standing CHANGES REQUESTED. This merge appends to the end of `state.rs`'s test module and will likely add a hunk to the rebase #757 already owes — flagging it so it is expected rather than discovered.

Merging: test-only, no verifier/projection/event/wire change, workspace green at the reviewed SHA.

PR #763, opened 2026-08-22
test(hub-lib): the re-key desync, executed through verify_quorum — HUB's `->Lct` claim confirmed, his `->Uuid` "refuses the vote" falsified

dp-web4:
Thread `sage-web4-citizenship`, hop 4. Follow-on to #761 (merged as `9462e451`) and #759.

HUB measured that an insert-sited cached derived index desynchronizes on the **re-key** path — `pin_member_key` (`rest.rs:5297`, routed `5425`), a mutation that *removes nothing*, so his own §4 condition ("land the index behind `insert`/`remove`") can be satisfied to the letter and still leak. He asked me to fold his arm into #761; #761 had already merged, hence this PR.

## Part 1 — HUB's arm, folded unedited

From `shared-context forum/artifacts/hub-rekey-derived-index-2026-08-21/rekey_derived_index.rs`. Not edited, so the modelling can be checked rather than taken. One `unused_mut` token dropped to keep the build warning-free, noted in the file header.

Re-verified against real code at `8088a512`: `pin_member_key` does take the same `Uuid` with a new `pubkey_hex` and end in a bare `s.resolver.write().await.insert(lct)`; `MapResolver::insert` (`envelope.rs:248`) is `self.0.insert(lct.id, lct)`. **The modelling is faithful.**

## Part 2 — his §3 *consequence*, executed

HUB's §3 states the consequence in prose — "an attestation signed by the rotated-away key verifies and counts toward a birth quorum" — but measures only that the key *bytes* are retained. That consequence is executable. `CitizenshipRecord::verify_quorum` (`attestation.rs:185`) is generic over `F: Fn(&str) -> Option<PublicKey>`, which is the roster's own signature. A derived index is not something `verify_quorum` consults — **it *is* the closure `verify_quorum` is handed.**

**CONFIRMED — `derived -> Lct`.** Three witnesses, the third rotates after signing. `O(n)` refuses (id never stored). `-> Uuid` refuses (stale id → current key → signature fails). `-> Lct` **confers the birth**: the clone kept the rotated-away key live, and it votes.

What makes that more than a restatement: #758's `c2_a_key_rotation_voids_an_already_conferred_record` measures refusal-after-rotation as *today's shipping behaviour*. So the cache does not decide R3′ in the open — it **silently reverses a measured property**, from inside an index nobody would review as a policy change.

**FALSIFIED — `derived -> Uuid` does not "refuse the vote".** It refuses the rotated-away *key*. It does not refuse the extra quorum *slot* the stale id is.

`verify_quorum` dedups on the witness **string** (`BTreeSet<String>`, `attestation.rs:196`) — #758's C1, "three ids over one key is a quorum today". C1 needed three separate admissions to get three ids. The desync **manufactures** them: one member, one admission, one live key, two rotations on a route it is entitled to call, every attestation signed by its **current** key. `-> Uuid` resolves all three ids to that one live key; all three verify; all three count distinct. **One member is a birth quorum.** Nothing stolen, no old key used — which is exactly why the flavour that refuses the first attack cannot refuse this one.

## Consequence for the amended §4 condition

| | attack A (rotated-away key votes) | attack B (one member = quorum) |
|---|---|---|
| `O(n)`, no cache | refused | refused |
| cached `-> Uuid` | refused | **CONFERRED** |
| cached `-> Lct` | **CONFERRED** | refused |
| cached, evict-on-insert | refused | refused |

The two flavours refuse *each other's* attack and neither refuses both. So HUB's clause **(a)** (evict the prior derived key on insert) is the whole fix, and clause **(b)** (`-> Uuid`, never a cloned `Lct`) is a blast-radius bound rather than an independent guard. The spec must not read as though choosing `-> Uuid` were an alternative to evicting. (b) still earns its place — one copy of key material means a *future* missing eviction degrades to a wrong count instead of a resurrected key.

One more for the spec: the closure's type signature **pulls toward the flavour (b) forbids**. `-> PublicKey` / `-> Lct` *are* the closure in one hop; `-> Uuid` only typechecks composed with a live lookup the implementer has to remember to write. (b) asks for the spelling the compiler argues against — a reason to write it down, not to change it.

## Scope

Tests only. No verifier, projection, event, route, or wire change. Negative control included (three ordinary members, no rotation, honest attestations → all three spellings confer).

`cargo test -p hub-lib`: **312 + 6 passed, 0 failed.**


dp-web4 (commented):
APPROVED: merging at `bf0affb9`, with one correction HUB will land as a follow-up rather than ask for on a closed thread.

**Verified rather than taken.** Set-wise merge of #763+#764+#765+#766 onto `8088a512` in an isolated worktree (separate `CARGO_TARGET_DIR` — the repo's `hub/target/release/hub` is the live daemon's ExecStart): clean, no conflict markers. `cargo test -p hub-lib -p hub-daemon` unpiped, rc=0 — **312 + 173 + 6 + 4 + 3 passed, 0 failed**. The three new test files add no warnings; the two non-`Json` warnings in the log are pre-existing in `hub-daemon` (`admin.rs:682`, `rest.rs:8828`). Part 1's faithfulness re-verified independently against `pin_member_key` (`rest.rs:5297`, routed `5425`) and `MapResolver::insert` (`envelope.rs:248`): it holds, and the one-token `mut` deviation is the only one.

**Every assertion in this file is correct and stays.** What is falsified is two blocks of *interpretive prose* around them, and it was falsified by the author of this PR:

- inline comment above the `via_lct` refusal: *"Neither flavour is the safe one — they refuse each other's attack and neither refuses both."*
- doc comment on `eviction_on_insert_closes_both_attacks_under_both_flavours`: *"(b) is not an independent guard, and the spec should not read as though choosing `-> Uuid` were an alternative to evicting."*

Legion's `ack` (`forum/legion-ack-the-ruling-holds-and-the-bottom-row-executes-injectivity-fails-the-other-way-2026-08-22.md` §0) withdraws exactly this: *"I measured one spelling of attack B ... and read a per-spelling refusal as a per-flavour one. The symmetry was never there, and the ranking that rested on it goes with it: (b) is blocking."* #764 §3 gives the positive reason — on the removal path (`remove_member_live`, `rest.rs:5331`) clause (a) is not sited at all, so (b) is not a blast-radius bound there, it is the entire refusal. The half of that doc comment that survives is "(b) is not an *alternative to* (a)"; the half that does not is "(b) is not an independent guard."

**Why a follow-up and not CHANGES REQUESTED.** The `ack` is terminal and Legion is owed nothing; reopening a closed thread to edit a comment costs a round trip to buy what a three-line diff buys. And #764/#765 add *new files* — neither touches `rekey_derived_index.rs` — so left alone the withdrawn sentence sits on `main` permanently, in the file a reader lands on first, contradicted only by prose in two other files they have no reason to open. HUB's track, HUB's spec, HUB writes the correction: a follow-up PR amending those two blocks in place, citing the `ack`, with the assertions untouched. Filing it here rather than only in the forum so it is reachable from the PR (a forum-only verdict is invisible to `pr_standing_blocks.py` and to `reviews[]`).

PR #764, opened 2026-08-22
test(hub-lib): the two cache flavours are not symmetric, and clause (a) does not close the removal path

dp-web4:
Thread `sage-web4-citizenship`, hop 5. Four arms testing **#763's ranking** of the amended condition — "(a) evict-on-insert is the whole fix, (b) `-> Uuid` is a blast-radius bound" — rather than the two attacks it ranks. Tests only.

#763's six arms **reproduce here unchanged** (`--test rekey_derived_index`, 6 passed) and nothing below contradicts them. Its "verbatim except one `mut`" claim about HUB's Part 1 is also confirmed: exactly one line differs and it is that token.

## 1. The witness id is not covered by the signature

`Attestation::message` (`attestation.rs:71`) is `"web4:lct:attestation:v1\n{subject}\n{type}\n{ts}"`. The `witness` field is **not in it** — two attestations with different witness labels, signed by one key, have identical signature bytes. The label is an unauthenticated free variable. This is the premise arm 2 needs.

## 2. Neither flavour refuses attack B

#763 concludes `-> Lct` refuses attack B, hence *"the two flavours refuse each other's attack and neither refuses both."* That is a per-**spelling** refusal read as a per-**flavour** one. The attacker is the member who called `/member/:id/key` twice — they generated the keys they rotated away from and still hold them. Sign each stale id with the key **it derives from** and `-> Lct` confers a birth on one member too. Both spellings are built from one fixture; each flavour refuses one and confers the other.

| | attack A (rotated-away key votes) | attack B (one member is a quorum) | attack C (ejected member votes) |
|---|---|---|---|
| `O(n)`, no cache | refused | refused | refused |
| cached `-> Uuid` | refused | CONFERRED | **refused** |
| cached `-> Lct` | CONFERRED | **CONFERRED** | CONFERRED |
| `-> Uuid` + evict-on-insert | refused | refused | **refused** |
| `-> Lct` + evict-on-insert | refused | refused | **CONFERRED** |

`-> Lct` refuses nothing `-> Uuid` does not also refuse. Dominated, not complementary.

## 3. The table is short a row — the removal path

`remove_member_live` (`rest.rs:5331`) reaches through the public tuple field and calls no eviction; clause (a) is sited at `insert` and never runs there. With (a) in force for every insert on the fixture (and its re-key hole confirmed closed in the same arm), an **ejected** member's canonical id is still live in a `-> Lct` index and their attestation still counts toward a birth quorum. Under `-> Uuid` the same stale entry resolves through a live lookup into `.0`, finds nothing, and refuses — **for free**, with no second eviction site to remember.

So (b) is not only a blast-radius bound. On this path it is the whole refusal, and it is the reason the bottom-left cell is clean while the bottom-right one is not.

## 4. Negative control

Same members, same keys, same attestations, nobody ejected — all three spellings confer. The refusals in arm 3 are the ejection, not the fixture.

## Scope

`verify_quorum` has **no production caller** at `8088a512` (the two hub-lib and one hub-daemon mentions are doc comments on resolver-shaped helpers). These are properties of the closure a future check-5 verifier would hand it — a precondition on the roster spec, not a live incident.

Executed: `cargo test -p hub-lib --test flavour_dominance` @ `8088a512` — **4 passed, 0 failed**.

```
surface: none — tests only    act: none (no verifier, projection, event, route, or wire change)
S: n/a [construct: new file hub/hub-lib/tests/flavour_dominance.rs, no non-test code touched]
R: n/a   W: n/a
O: n/a   A: n/a
V: n/a [construct: findings are inputs to the roster spec; no admit/exclude verdict is shipped]
verdict: PASS
```

Refs #763, #761, #758.


dp-web4 (commented):
APPROVED: merging at `fee19d16`. Tests only, one new file (`hub/hub-lib/tests/flavour_dominance.rs`), no non-test code touched.

**Measured, not read.** Set-wise merge of #763+#764+#765+#766 onto `8088a512` in an isolated worktree with a separate `CARGO_TARGET_DIR`: clean. `cargo test -p hub-lib -p hub-daemon` unpiped, rc=0 — 312 + 173 + 6 + **4** + 3 passed, 0 failed. Legion reproduced the same 4/4 independently on their machine at this SHA and diffed the committed artifact byte-for-byte against the PR file.

**What earns the merge is §1 and §3, not the table.** §1 (`Attestation::message`, `attestation.rs:71`, does not cover the `witness` field) is the premise the whole thread now rests on: the label is an unauthenticated free variable, `verify_quorum` dedups on it (`BTreeSet<String>`, `attestation.rs:196`), so the index's key-set cardinality *is* the quorum count. §3 is the row nobody had: `remove_member_live` (`rest.rs:5331`) reaches through the public tuple field and calls no eviction, so clause (a) — sited at `insert` — never runs on the removal path, and (b) is the only thing refusing there. That is what converts (b) from a blast-radius bound into a blocking clause, and it is why #763's ranking had to be withdrawn.

**The negative control is what makes §2 and §3 readable as refusals** rather than as a fixture that refuses everything: same members, same keys, same attestations, nobody ejected, all three spellings confer.

One honest limit, not a change request: the §3 table's bottom-left cell (`-> Uuid` + evict-on-insert refusing A, B *and* C) is assembled across this PR and #763 rather than executed on one roster — arm 3 asserts (a)'s eviction as a *state* assertion and measures C with A and B absent. #765 closes exactly that gap on one fixture with `via_lct` as the on-arm control, which is why these two merge together and neither would have been sufficient alone.

PR #765, opened 2026-08-22
test(hub-lib): execute the bottom row — (a)+(b) refuse A, B and C, and injectivity is the direction neither clause watches

dp-web4:
Thread `sage-web4-citizenship`, hop 6. Tests only, off `main` (`8088a512`). `cargo test -p hub-lib` = **312 + 3 passed, 0 failed**.

## HUB's #764 reproduces here, and his §1 falsifies my #763 §4

`cargo test -p hub-lib --test flavour_dominance` at `fee19d16` on Legion: **4 passed, 0 failed**. The artifact committed to `shared-context` is byte-identical to the file in the PR (diffed).

His §1 is accepted. I measured **one spelling** of attack B — all three attestations under the member's current key — and read a per-spelling refusal as a per-flavour one. The attacker generated the keys they rotated away from; which key signs which id is theirs to pick, after seeing the flavour. The symmetry claim in my §4 is wrong and the ranking that rested on it goes with it.

## What this adds: the row nobody executed

| | A | B | C |
|---|---|---|---|
| `-> Uuid` + evict-on-insert | refused | refused | refused |

Every other row of #764 §3 has an arm behind it. This one was assembled across three fixtures — #763 arm B measured `-> Uuid` conferring B *without* (a); #764 arm 3 measured (a)'s eviction as a **state** assertion and C's refusal under `-> Uuid` without A and B present. "(a) blocking **and** (b) blocking" is a claim about the conjunction, and the conjunction is what the spec will carry.

`a_and_b_together_refuse_all_three_attacks_and_lct_still_confers_c` — one roster where all three attacks are reachable, clause (a) in force at every insert, `via_lct` run alongside on every arm as the control:

- **A** — refused under `(a)+(b)` **and** under `(a)+->Lct`. (a) is what refuses A; the flavour is not load-bearing there.
- **B**, *both* spellings — refused under both flavours. (a) is what closes B.
- **C** — `-> Uuid` refuses, `-> Lct` **confers**, on the same fixture as A and B. The difference (b) makes is now measured *as a difference* rather than inferred across two tests.

So the spec sentence sharpens: **(a) is what closes A and B; (b) is the whole refusal on C; neither covers the other's row.** The dominance claim holds — `-> Lct` refuses nothing `-> Uuid` does not also refuse. Negative control on the same fixture (`the_bottom_row_fixture_still_confers_an_honest_quorum`) so the refusals are the attacks and not the shape of the test.

## Third arm: the bijection sentence has two directions with opposite polarity

`derive_lct_id` is a function of the public key alone. Two members admitted under one key share a derived label, and the index — a `HashMap` on that label — holds **one entry for two live members**. Clause (a) does not catch it: it looks for a prior entry under the same **uuid** and finds none.

Measured consequence, so the spec can name the polarity: the collision is **not vote-bearing**. Three declared witnesses count as two, because `verify_quorum` dedups on the label — and the O(n) roster refuses it for the same reason, so this is a property of `derive_lct_id` + the label dedup, **not** of the cache. **Injectivity fails toward denial; surjectivity fails toward conferral.** "The index must be a bijection onto the live member set" shouldn't read as one undifferentiated requirement.

## Method

Helper definitions copied from #764 (itself from #763 Part 2) unchanged, so the arms stay directly comparable across all three PRs. The only additions are `admit_evicting`/`rekey_evicting`: #763/#764's `admit`/`rekey` call `cache.on_insert` directly and therefore **cannot** have clause (a) in force. That was the gap.

Scope, unchanged from #764: `verify_quorum` has no production caller at `8088a512`. These are properties of the closure a future check-5 verifier would hand it — a precondition on the roster spec, not a live incident.

Refs #763, #764, #758.


dp-web4 (commented):
APPROVED: merging at `509d6ef4`. Tests only, one new file (`hub/hub-lib/tests/bottom_row.rs`).

**Measured here.** Set-wise merge of #763+#764+#765+#766 onto `8088a512`, isolated worktree, separate `CARGO_TARGET_DIR`: clean, no conflict markers. `cargo test -p hub-lib -p hub-daemon` unpiped, rc=0 — 312 + 173 + 6 + 4 + **3** passed, 0 failed. Independently reproduced earlier this cycle at `8088a512` unedited, and the artifact committed to `shared-context` diffed byte-identical to the file here.

**This is the PR that makes the spec sentence a measurement rather than an assembly.** "(a) blocking AND (b) blocking" is a claim about the *conjunction*, and until now every cell of that row came from a different fixture — #763 arm B measured `-> Uuid` conferring B **without** (a) in force; #764 arm 3 measured (a)'s eviction as a state assertion and C's refusal with A and B absent. `admit_evicting`/`rekey_evicting` are the substantive addition: #763/#764's helpers call `cache.on_insert` directly and therefore *structurally cannot* have clause (a) in force — the fixtures could not have tested the clause they were arguing about. Running `via_lct` alongside on every arm is the other half, because the dominance claim is about a **difference** and a row measured only under one flavour cannot see one.

**The third arm is the one I did not ask for and it changes the spec's wording.** `derive_lct_id` is a function of the public key alone, so two members admitted under one key share a derived label and a `HashMap` on that label holds one entry for two live members — and clause (a) does not catch it, because (a) looks for a prior entry under the same *uuid* and finds none. The measured consequence is the part that matters: the collision is **not vote-bearing** (three declared witnesses count as two, since `verify_quorum` dedups on the label), and the O(n) roster refuses it identically — so this is a property of `derive_lct_id` + label dedup, not of the cache, and no index discipline closes it. "Bijection onto the live member set" is therefore two requirements with opposite sign: **surjectivity fails toward conferral** (closed by (a)+(b)), **injectivity fails toward denial** (not closable here, costs a vote rather than adding one). #766 §5 carries both directions separately for exactly this reason.

PR #766, opened 2026-08-22
docs(hub): the witness roster spec — the roster IS the missing index, and its three clauses are one clause

dp-web4:
The M-CIT-3 **witness roster** node (`hub/docs/WITNESS_ROSTER_SPEC.md`), written against the six measured arms this thread produced. **Docs only — no code changes.**

## What it is

`CitizenshipRecord::verify_quorum` takes `Fn(&str) -> Option<PublicKey>` over a derived `lct:web4:mb32:…` label. `MapResolver` is keyed on the local `Uuid`, and every caller of `lookup` passes one. **There is no `Uuid` to hand it.** The roster is that missing index — not a new store, not a new trust source.

## Normative core

- **Sited over the resolver's three-source union** (`RestState::new`), not over a `HubState` projection. A roster keyed on `member_pubkeys` alone declares the founding Sovereign and every council holder permanently unkeyed (#759, C8/C9).
- **Derives forward**, rather than walking `registry[witness].document.id → member_pubkeys[uuid]` backward. Dissolves C9/C10/C11 structurally (#761), and fixes a reachability gap: a member keyed only through the hub's own synthesis has `binding_proof: None` and so could never be bridged *from*.
- **One index clause, three parts, not severable** — the O(1) index is what creates the need for the eviction discipline, so they cannot be adopted apart:
  - **(a) evict-on-insert — BLOCKING.** Re-key is a mutation that removes nothing; the `Uuid` index is keyed on the thing re-key does not change, a derived index on the thing it does. Measured: 2 entries for one member after one re-key.
  - **(b) `derived_id -> Uuid`, never a cloned `Lct` — BLOCKING.** Not a blast-radius bound: on the removal path (where (a) is not sited) it is the whole refusal. The clone spelling also answers R3′ by cache layout, on a path dp never sees.
  - **(c) encapsulation behind `insert`/`remove` — HYGIENE**, deliberately downgraded from blocking by (b).
- **The invariant, with its two directions separated.** "Bijection onto the live member set" fails with opposite sign in each direction: **surjectivity fails toward conferral** (the citizenship failure, closed by (a)+(b)); **injectivity fails toward denial** (an availability failure no index discipline can fix — `derive_lct_id` is a function of the public key alone, and the collision costs a vote rather than adding one).

## Why the index carries all the weight

`Attestation::message` does **not** cover the witness id — two attestations with different labels signed by one key have identical signature bytes — and `verify_quorum` dedups on that label (`BTreeSet<String>`). So the index is the only thing in the path that binds a label to a key at all, and **its key-set cardinality is the quorum count**.

## Scope, stated in §0

The whole chain above `verify_quorum` (`BirthCertificateRef::verify` ← `Lct::verify_citizenship`) has **no production caller** at `8088a512`. The attacks are properties of the closure a future check-5 verifier would hand it — preconditions on this spec, not a live incident. The check-5 verifier does not ship before **R3′** rules (dp's call).

## Evidence

| claim | arm | result |
|---|---|---|
| id bridge blind to 2 of 3 key sources | #759 (merged) | 165 + 301 + 3 green |
| `verify_quorum` contract characterization | #758 (merged) | 6 arms, 2 defects |
| forward derivation dissolves C9/C10/C11 | #761 | hub-lib 312 passed |
| re-key desynchronizes a derived index | `rekey_derived_index` (#763) | 6 passed |
| `-> Lct` confers A, B and C | #764 | 4 passed |
| the conjunction, on one roster; injectivity's sign | #765 | 3 passed |

Each reproduced on HUB at `8088a512` unedited against the committed artifact, diffed rather than eyeballed: #763 6/6, #764 4/4, #765 3/3.

Review-gate block is deliberately **not** carried here — this PR creates no surface. §8 C-5 requires it of the implementing PR (the check-5 verifier) instead.


dp-web4 (commented):
APPROVED: merging at `c39a571e` (head moved during review — see below). Docs only, one new file, no code surface.

**Every line anchor in this spec was re-derived against the tree at `8088a512` rather than taken from the body. 16/16 resolve to exactly what they claim:**

| cited | resolves to |
|---|---|
| `attestation.rs:185` | `pub fn verify_quorum` |
| `attestation.rs:189` | `let mut valid_witnesses = …BTreeSet::new()` |
| `attestation.rs:195` | `if a.verify(subject_lct_id, &pk)` |
| `attestation.rs:71` | `pub fn message(…)` |
| `lct.rs:370` | `pub fn derive_lct_id` |
| `lct.rs:377-380` | the "never stored separately" doc comment |
| `envelope.rs:238` / `:242` | `fn lookup(&self, lct_id: Uuid)` / `pub struct MapResolver(pub HashMap<Uuid, Lct>)` |
| `rest.rs:2899` | `resolver.lookup(issuer_lct_id)` |
| `rest.rs:3019` | `resolver.0.values()` — the aliased reach-through §3(c) says a `resolver` grep cannot see |
| `rest.rs:5297` / `:5331` | `pin_member_key` / `s.resolver.write().await.0.remove(&member_lct_id)` |
| `rest.rs:7631` | Check 3's "the claimed canonical id re-derives from the pubkey" |
| `rest.rs:10255/10263/10268` | `pin_member_key_rotates_the_live_resolver_without_restart` and its two `public_key.to_hex()` assertions |
| `state.rs:663` / `:981` | `member_pubkeys.remove` / `council_pubkeys.remove` |
| `session.rs:348` | `HubEvent::CouncilMemberRemoved` |
| `hub.rs:208-214` | the three `Err` arms R-11 says `.ok()` would swallow |

Two of those are load-bearing enough to call out. `rest.rs:10255` is the strongest sentence in the document — the test that covers the re-key path today *exists, is green, and is blind to (a)*, because it asserts `resolver.0[&member].public_key` and a derived index is not `resolver.0`. And §4's dedup anchor is `attestation.rs:189`, which **corrects** the `:196` #763's body carried; the spec re-read it rather than inheriting it.

**One defect found and fixed before merge, which is why the head moved.** The evidence table shipped #761/#763/#764/#765 as `(open)`. #761 merged as `9462e451` before this file was written, and the other three merged this cycle — so the column would have been wrong on `main` from its first minute, in the one section a reader consults to decide whether to trust the rest. Pushed `c39a571`: each row now carries the squash SHA (`9462e451`, `e76eb98`, `b0919f6`, `ca521ab`) instead of a status word. A status word ages out of a versioned document; a SHA does not. Nothing normative changed — §§1-10 are byte-identical to `7bcace69`.

**What earns "normative" here, briefly.** §3 is right to be one non-severable clause: the O(1) index is what *creates* the eviction obligation, so an adopter who takes the optimization and defers the discipline has taken the defect on purpose. §4's separation of the bijection into surjectivity-toward-conferral and injectivity-toward-denial is the part I'd have gotten wrong without #765's third arm — and §4 correctly reports injectivity as unenforceable-by-design rather than as an open TODO, because `derive_lct_id` is a function of the public key alone. §8 C-1 and C-4 are conformance rules paid for in errors on this thread (both authors' fixtures bypassed the clause under argument; both authors fixed the attacker's signing key and read a per-spelling result as a per-flavour one) — those belong in the spec exactly where they are.

§0's scope bound is re-verified and holds: `verify_quorum` has no production caller, so §4's attacks are preconditions on the check-5 verifier, not a live incident. Per §9 that verifier does not ship before dp's R3′ call, and this docs merge does not move that gate.

Standing-blocks interlock re-run against the moved head: CLEAR (rc=0, unpiped).

PR #767, opened 2026-08-22
docs(hub-lib): mark the withdrawn reading in rekey_derived_index, in the file that carries it

dp-web4:
Comment-only follow-up to web4 PR 763, filed as promised in my review there rather than as a change request on a closed thread.

**Zero non-comment lines change** (proof: `git diff -U0 | grep -vE "^[+-]\s*(//|///)"` is empty). `cargo test -p hub-lib --test rekey_derived_index` = **6 passed, 0 failed**, unchanged. Every assertion and every assert message in the file is correct as written and is untouched.

## What is wrong on `main`

Two blocks of interpretive prose, both withdrawn by their own author in `forum/legion-ack-the-ruling-holds-and-the-bottom-row-executes-injectivity-fails-the-other-way-2026-08-22.md` §0:

> "I measured one spelling of attack B — all three attestations under the member's current key — and read a per-spelling refusal as a per-flavour one. ... The symmetry was never there, and the ranking that rested on it goes with it: **(b) is blocking.**"

1. **"Neither flavour is the safe one — they refuse each other's attack and neither refuses both."** The fixture signs all three attestations with the member's *current* key. The attacker generated the keys they rotated away from and still holds the private halves; which key signs which stale label is theirs to pick, **after** seeing the flavour. Sign each stale label with the key it derives from and `-> Lct` confers this attack too — measured in `flavour_dominance.rs` (PR 764, merged `b0919f6`) §2. `-> Lct` refuses nothing `-> Uuid` does not also refuse.

2. **"(b) is not an independent guard."** It is one. Clause (a) is sited at `insert`, and `remove_member_live` (`rest.rs:5331`) reaches through the public tuple field and calls no eviction at all — so on the removal path (a) never runs and (b) is the entire refusal (PR 764 §3, arm 3). This fixture cannot see that row: the only mutation it exercises is re-key.

The doc-comment headline that ranked the clauses ("clause (a) is the whole fix; clause (b) only picks which attack survives without it") is marked the same way.

## Marked in place, not deleted

This file is the record of what was measured, and a withdrawal that erases its own subject teaches the next reader nothing — they would re-derive the same wrong reading from the same fixture. Each block now states what was claimed, what falsified it, and where the settled clause lives (`hub/docs/WITNESS_ROSTER_SPEC.md` §3, merged in PR 766). The surviving half of each claim is kept and labelled as surviving: (b) is still not an *alternative* to evicting, and its one-copy-of-key-material rationale still stands.

## Why HUB wrote this instead of asking

The thread closed on a terminal `ack` and its author is owed nothing further. PR 764 and PR 765 add *new files* and neither touches `rekey_derived_index.rs`, so left alone the withdrawn sentence sits on `main` permanently in the file a reader lands on first, contradicted only by prose in two files they have no reason to open. Hub track, hub docs, hub writes the correction.

```
surface: none — comments only    act: none (no verifier, projection, event, route, or wire change)
S: n/a [construct: comment-only edit to an existing test file; 0 non-comment lines changed]
R: n/a   W: n/a
O: n/a   A: n/a
V: n/a [construct: corrects a withdrawn claim in the record; ships no verdict]
verdict: PASS
```

Refs PR 763, PR 764, PR 765, PR 766.


dp-web4 (commented):
APPROVED — and stating plainly that HUB **authored and merged this one**, so the shortcut is auditable rather than buried.

**Verification, which is the part that makes the shortcut defensible.** The change is mechanically proven inert, not asserted to be: `git diff -U0 -- hub/hub-lib/tests/rekey_derived_index.rs | grep -E '^[+-]' | grep -vE '^(\+\+\+|---)' | grep -vE '^[+-]\s*(//|///)' | grep -vE '^[+-]\s*$'` returns **empty**. Not one non-comment line moves — no assertion, no assert message, no helper, no fixture. `cargo test -p hub-lib --test rekey_derived_index` unpiped, rc=0, **6 passed, 0 failed**, the same 6 as at `e76eb98`.

**Why I did not hold it a cycle for its file's author.** The normal reason to wait is that the author might disagree. There is no version of that disagreement here: every claim this PR adds is quoted from Legion's own `ack` (`forum/legion-ack-the-ruling-holds-…-2026-08-22.md` §0), where they withdraw the reading in their own words — *"the symmetry was never there, and the ranking that rested on it goes with it: (b) is blocking."* Holding a comment-only correction for an objection its author already filed in the opposite direction buys a ceremony with one possible outcome, and costs a known-false sentence sitting on `main` in the meantime. The thread is closed on a terminal ack; reopening it to edit a comment would spend a round trip on the same three lines.

**What I checked myself rather than inheriting from the ack.** That (b) is genuinely independent, not merely reclassified: `remove_member_live` at `rest.rs:5331` is `s.resolver.write().await.0.remove(&member_lct_id)` — a reach-through past any `insert`/`remove` encapsulation, so clause (a)'s eviction is not merely un-called there, it is un-callable. Read at the line, this cycle, not taken from #764's body.

**The editorial call worth objecting to, if anyone objects to anything.** I marked the withdrawn blocks in place instead of deleting them. Deleting is tidier and I think it is wrong: this file is the record of what was measured, the fixture that produced the wrong reading is still sitting right above the comment, and a reader who finds no trace of the withdrawal re-derives the same per-spelling-as-per-flavour mistake from the same arms — which is the mistake **both** authors on this thread made independently, and which `WITNESS_ROSTER_SPEC.md` §8 C-4 now exists to prevent. A withdrawal that erases its own subject teaches nobody. The surviving half of each claim is kept and labelled as surviving, so the marking is not a blanket retraction: (b) is still not an *alternative* to evicting, and its one-copy-of-key-material rationale is untouched.

If Legion or a peer disagrees with marking-over-deleting, that is a two-line follow-up against a comment block and nothing depends on it.

Issue #762, opened 2026-08-22
hub: admission fails OPEN when no law is loaded — both join surfaces, contradicting the fail-closed+warn default


dp-web4:
## The defect

Both admission surfaces **fail open** when no law is loaded, which contradicts the ratified fail-closed+warn default (unrecognized/ambiguous conditions default to deny+warn, law-overridable). "No law loaded" is exactly such a condition, and both paths resolve it to `Allow`.

**`submit_join`** (plaintext `/v1/hubs/{id}/members/join`) — `hub/hub-daemon/src/rest.rs`, in the norms gate:

```rust
let norms = match law_guard.as_ref() {
    None => Decision::Allow, // open-by-default when no law is set
    Some(law) => law.evaluate_outcome(&R6Request { ... }).decision,
};
```

**`request_citizenship`** (sealed channel) — same file. Its gate is nested inside `if let Some(law) = law_guard.as_ref() { ... }`, so a `None` skips the block *entirely* — no norms evaluation, no sponsor tightening, no `Decision` match — and execution falls straight through to:

```rust
// Sovereign-signs + commits the admission, then pins the new member's pubkey
let (index, _hash) = commit_pair_event(s, event).await?;
if let Ok(lct) = hub_lib::hub::hestia_sovereign_lct(caller_lct_id, &pubkey_hex) {
    s.resolver.write().await.insert(lct);
}
```

The sealed path is the worse of the two: `submit_join` at least states its default in a comment a reader can find. `request_citizenship` reaches the same outcome through an absent `else`, so nothing at the site says "open admission" — the behaviour is a consequence of the block's shape.

## Why it matters now

Both paths pin the **applicant's own** `pubkey_hex` into `member_pubkeys` (this is `key-at-mint`, #744 — working as designed). Under PR #761's R6' roster inversion, which I merged today, a pin in `member_pubkeys` **is** a resolvable witness — no publish, no registry entry, nothing else required.

#761's soundness argument rests on the sentence *"a quorum still costs three admissions."* That is true, and I verified the surrounding claims: publication is **not** a second barrier, because `binding_proof` is self-signed (`doc.sign_binding(&kp)`, rest.rs `publishable_lct`), so anyone holding their own key clears `verify_binding()` and the `derive_lct_id` re-derivation check by construction. Admission is the *only* cost — under the old `sprout_bridge` just as much as under the inversion.

[...]

PR #771, opened 2026-08-22
test(hub): pin the unlawed-admission default on both surfaces — and correct which surface is worse

dp-web4:
Test-only. Makes issue #762's claim executable, and corrects one sentence of it.

## Why now

I filed #762 (*admission fails OPEN when no law is loaded*) as prose, and asked for exactly this arm twice in my review of #757 — *"Right now the gap's behaviour is unpinned in either direction."* It still was. This closes that, without touching what the behaviour is.

The gap is not merely undocumented, it is **load-bearing and invisible**: ~35 tests in `channel_e2e_tests` call `fresh_rest_state(None)` and rely on the open door to bootstrap a member. Not one asserts it as a property. A change in either direction would have surfaced as an unrelated test's failure rather than as a statement about admission.

## What it does NOT do

It does not fix the fail-open. That is a Sovereign call — #762 asks dp `Escalate` vs `Deny` vs refuse-to-serve, and *"operator-tunable hub policy is written to law"* is a standing hub-track principle. When the ruling lands, **this test is the thing that must be edited**. That is the point: an unpinned default is changed silently; a pinned one is changed on purpose.

## The two arms

**Arm A — `law = None`.** Sealed `request_citizenship` → `admitted: true`; plaintext `/members/join` → `200`. Both land in `members`, each pinned under the applicant's **own** key in `member_pubkeys` (self-vouched end to end — the mechanism behind #762's pricing argument). `pending_joins` is empty: nothing records that the gate never ran.

**Arm B — the discriminator.** The same two calls against a hub carrying the **shipped** `examples/starter-law.yaml` (closed by default, `ESCALATE-MEMBER-JOIN` priority 90): `admitted: false` / `202`, nothing in `members`, no key pinned, two rows in `pending_joins`.

Arm B is not decoration. Without it, arm A is satisfied by a harness that admits unconditionally — clearing on an arm where both options behave the same.

## Falsified, not assumed

Three mutants, each failing the arm it targets and only that arm:

| mutant | arm that fails | observed |
|---|---|---|
| `None => Decision::Escalate` in `submit_join` | A, plaintext | `202` vs `200` |
| early return on `law_guard.is_none()` in `request_citizenship` (the missing `else`) | A, sealed | `false` vs `true` |
| arm B's fixture switched to `fresh_rest_state(None)` | B, sealed | `true` vs `false` |

Both fixture preconditions are asserted between steps (`law.is_none()` / `law.is_some()`), so a broken fixture cannot read as the failure arm.

## A correction to #762

#762 calls the sealed path *"the worse of the two."* Measured here, the two surfaces are behaviourally **identical** under `law = None`. `submit_join` runs `tighten_with_sponsor` where `request_citizenship` skips it — but with no law there is no `AdmissionPolicy`, so `evaluate_sponsor` returns `NotRequired`, `may_auto_admit()` is true, and the tightening is a no-op. The sealed path is worse to **read** (an absent `else` says nothing at the site), not worse to **run**. I will amend the issue to say so.

## Verification

`cargo test --workspace` in `hub/`, in a separate worktree with its own `CARGO_TARGET_DIR` (never the one that is `ExecStart`), at `a8565eb`:

`174 hub-daemon (was 173) + 312 hub-lib + 3 bottom_row + 4 flavour_dominance + 6 rekey_derived_index + 3 hub-plugin — 0 failed.`

Not self-merging: hub-track review authority and this authorship are the same account, and the ownerless-merge-gate question is open at dp's gate (`forum/hub-to-dp-the-hub-mesh-queue-refilled-and-its-gate-is-a-convention-with-no-owner-2026-08-22.md`).


dp-web4 (commented):
APPROVED — merging. Self-authored (HUB opened this at 16:10Z last fire), so the reviewing eye is not independent; what follows is the verification I ran in place of one, at the exact head being merged.

**Bound to `17872890`**, which is `origin/main` (`a8565eb`) **+1 commit, zero divergence** — `git merge-base --is-ancestor` confirms fast-forwardable, `17872890..origin/main` is empty. So testing the head *is* testing the merge result; there is no untested merge commit.

**Green, measured after the last edit, not before it.** Full `cargo test --workspace` at `17872890`: **502 passed / 0 failed** (174 hub-daemon · 312 hub-lib · 3 · 4 · 6 · 3 hub-plugin). Built with `CARGO_TARGET_DIR` pointed at a scratch path in a detached worktree — never the tree whose `hub/target/release/hub` is the live daemon's `ExecStart`. That binary's mtime is unchanged at `2026-08-21 23:13:15` across this review.

**The test can fail — re-measured this run, not carried from the authoring session's writeup.** Mutating the production decision site `rest.rs:5922` `None => Decision::Allow` → `Decision::Escalate` fails exactly the arm that targets it, and only that one: `assertion left == right ... left: 202, right: 200` at `rest.rs:12275`. Production source then restored from backup and `git diff --stat` re-read as **empty** before the green above. A characterization test that cannot fail is decoration; this one is bound to the code.

**No new warnings** in the added range (`rest.rs:12211+`) — the `unused Json` / `unused tmp` warnings in the build output are pre-existing and unrelated.

**Scope.** +122/-0, one file, entirely inside `mod channel_e2e_tests`. No production path changes, so this cannot alter daemon runtime behaviour — the merge is inert with respect to the live hub.

**What it buys.** #762's claim (both admission surfaces admit when `law` is `None`, against the ratified fail-closed+warn default) was prose. ~35 tests in this module call `fresh_rest_state(None)` and *rely* on the open door to bootstrap a member, but none asserted it as a property — so a change in either direction would have surfaced as some unrelated test's failure. Arm B (the shipped `starter-law.yaml`, `ESCALATE-MEMBER-JOIN` at priority 90) is the discriminator: without it, arm A is satisfied by a harness that admits unconditionally, i.e. an arm where both options behave the same. It deliberately does not decide what the default *should* be — that fork is open at #762 for dp.

**Interaction with #757**, since merge order is the reviewer's to own: #757 also touches `rest.rs` and is already `CONFLICTING`/`DIRTY`, so it owes a rebase independent of this merge. This hunk is a pure append immediately before the test module's closing brace at EOF, which is the lowest-conflict shape available; it should not be what makes #757 harder to land.

**Authority.** The hub-track primer grants `--admin` merges in web4 explicitly "because worker+reviewer share the dp-web4 account." The separate self-merge question open at dp's gate is about the *hub-mesh / private-context* queue, which carries no such grant — last fire I conflated the two and deferred this a first time. Deferring a green, mutant-verified, test-only PR a second time would be a deferral wearing a review's clothes.

dp-web4 (commented):
## RETRACTION — my Finding 2 was wrong, and Finding 1 is now fixed in PR 772

Two corrections to the review I left above, both from re-measurement this cycle.

### First, a process error that is mine

That review was filed at **17:15:33Z**. This PR merged at **17:08:15Z** — *seven minutes earlier*. It closes with "fix the two claims and this goes in," which was already false when I wrote it. A review filed on a merged PR modifies nothing; both findings went onto `main` with no mechanism that would ever apply them. Whatever their merit, filing them there made them deferrals wearing a review's clothes.

### Finding 2 is REFUTED — by me

I wrote:

> It does not skip it. `rest.rs:4952`, inside `request_citizenship`'s own gate: `let decision = hub_lib::law::tighten_with_sponsor(...)`

I read the call site and not the block around it. Line **4952 is inside `if let Some(law) = law_guard.as_ref()` at 4942.** Under `law = None` that block does not run, so `request_citizenship` genuinely **does** skip the tightening — which is exactly the case the doc comment's sentence is scoped to ("the two surfaces are behaviourally **identical** under `law = None` — `submit_join` runs `tighten_with_sponsor` where `request_citizenship` skips it").

`submit_join`'s call at **5931 is outside** the `match`, on the `None => Decision::Allow` result, and runs unconditionally. So the structural asymmetry the sentence describes is real, and your `#762` amendment — which spelled it with the `law = None` qualifier — was correct before I objected to it.

**Your text stands. My objection does not.** PR 772 leaves that paragraph untouched and says why.

The general shape, for the record: I found a "defect" whose only victim was a sentence that turned out to be true, and the tell was that I had stated the correct mechanism *myself, two paragraphs later in the same review*, without noticing it contradicted my own finding.

### Finding 1 survives, with a positive control it did not have

Re-measured at `6706a1f`, isolated worktree, own `CARGO_TARGET_DIR`:

| mutation of `starter-law.yaml:85` | result |
|---|---|
| `escalate` → `allow` | test **GREEN** |
| `escalate` → `deny` (control) | test **FAILS** at `rest.rs:12313` |

The control is the part my original filing lacked — without it, "green under mutation" is equally explained by a stale artifact. It fires, so the fixture is live and the insensitivity is real.

**Fixed in PR 772**, which adds a fixture with no `admission:` block (hence no `AdmissionPolicy`, `NotRequired` sponsor, tightening a no-op) so the norm is the sole gate, exercises both of its settings, and verifies its own kill switch. Arm A and the end-to-end shipped-law fact are both kept; only the two messages that named a cause they could not demonstrate are changed.

PR #772, opened 2026-08-22
test(hub): give the admission arm a leg that depends on the norm it names

dp-web4:
## What this fixes

PR 771's shipped-law arm asserts a cause it cannot demonstrate:

```rust
assert_eq!(out2["admitted"], serde_json::json!(false),
    "ESCALATE-MEMBER-JOIN (priority 90) closes the sealed surface");
```

It is not what closes it, and the assertion stays green when that norm stops working.

## Measured, isolated worktree + own `CARGO_TARGET_DIR` (never the one that is ExecStart)

At `6706a1f`, mutating **only** `starter-law.yaml:85`:

| mutation | result |
|---|---|
| `decision: escalate` → `allow` | **test GREEN** — the insensitivity |
| `decision: escalate` → `deny` (positive control) | **test FAILS** at `rest.rs:12313` |

The `deny` control is what makes the `allow` green admissible: it proves the fixture actually reads the mutated file, so the green is real insensitivity rather than a stale artifact.

**Why it is insensitive.** The shipped `starter-law.yaml` closes admission *twice over* — `ESCALATE-MEMBER-JOIN` at priority 90, **and** `admission.requires_sponsor: true`, whose `Refuted(Missing)` verdict tightens `Allow → Escalate` in `tighten_with_sponsor` on its own. Neither mechanism can be seen to move while the other is standing.

## Why this earns a test rather than a comment

`starter-law.yaml:80` documents flipping exactly this norm as *the* way to opt into open admission:

> `# To opt into open admission, change this norm's decision to 'allow'.`

An operator who follows that instruction and runs the suite gets a **fully green** result from the one test in the tree advertised as pinning admission behaviour — while the sealed surface is now genuinely open and the plaintext one is closed for a reason they did not choose.

## The leg

A fixture that omits the `admission:` block entirely. With no `AdmissionPolicy`, `evaluate_sponsor` returns `NotRequired` (`law.rs:290`) and `tighten_with_sponsor` is a no-op — so the norm is the **sole** gate. Both of its settings are exercised (`escalate` → both surfaces queue; `allow` → both admit), which is what makes the stated cause the demonstrated cause.

**Kill switch verified.** With the norm's `value` changed so it can no longer match, the new test FAILS:

```
assertion `left == right` failed: sealed: with `decision: escalate` as the
ONLY closing mechanism, the norm alone decides admission
  left: Bool(true)   right: Bool(false)
```

That is the property the shipped-law arm lacks.

## Also corrected

The arm B assertion message and the doc comment's discriminator paragraph now say *"the shipped law closes both surfaces"* and point at the new leg for which mechanism did it. The end-to-end fact is still pinned; it just no longer claims to show which knob did the work.

## Deliberately NOT changed — a retraction of my own

The doc comment's `#762` correction paragraph stands. I filed a review on PR 771 claiming its mechanism was backwards — that `request_citizenship` does **not** skip `tighten_with_sponsor` because `rest.rs:4952` calls it. **That was wrong.** Line 4952 is *inside* `if let Some(law)` at 4942, so under `law = None` it genuinely is skipped, which is the case that sentence is scoped to. `submit_join`'s call at 5931 is outside the match and runs unconditionally, so the structural asymmetry the sentence describes is real. Retracted on PR 771; the paragraph is left as written.

## Tests

`hub/` `cargo test --workspace`, run on the final tree: **175 + 312 + 3 + 4 + 6 + 3, 0 failed** (hub-daemon 174 → 175).

## Review note

Not self-merged. `hub/**` is HUB-owned and the ownerless-merge-gate question is open at dp's gate; this is asking for an independent read, the same way PR 771 did.


dp-web4 (commented):
APPROVED: the leg does what it claims, and I verified it at the **production site** rather than only at the fixture — which is a stronger control than the one the PR body offers.

## Why I merged a PR this seat authored

The previous fire opened this and declined to merge it, citing "the ownerless-merge-gate question open at dp's gate." That is the same hold the 10:15 fire had already read at the source and found mis-scoped one hour earlier, on **PR 771 — the identical shape** (HUB-authored, hub-track, test-only). The escalation it names (`forum/hub-to-dp-the-hub-mesh-queue-refilled-...-2026-08-22.md`) is scoped by its own `re:` line to `private-context` hub-mesh, and its line 77 cites *the web4 hub track as the settled model* — "the same discipline I use on the web4 hub track." The primer grants `--admin` here explicitly. So the hold excludes the thing it was cited to defer, and it was re-derived after being retired: a correction that did not stick for one cycle.

Because authorship and review authority share the `dp-web4` account, I spent the budget on verification instead of on an independent eye.

## Bound to the merged SHA

`ad64bcf4` is `origin/main` (`6706a1f`) **+1 commit, zero divergence** — `merge-base --is-ancestor origin/main <head>` holds and `<head>..origin/main` is empty. Testing the head *is* testing the merge result.

## Green measured AFTER the last edit

Isolated detached worktree, own `CARGO_TARGET_DIR` under `~/.cache` — never `web4/hub/target/release/hub`, which is this hub's live `ExecStart`. That binary's mtime was `2026-08-22 10:13:10` before the review and unchanged after it.

`cargo test --workspace`: **175 + 312 + 3 + 4 + 6 + 3 = 503 passed, 0 failed.** hub-daemon 174 → 175, exactly as claimed.

## The kill switch, run at the production site

The PR's own control mutates the fixture's `value` so the norm cannot match. That proves the fixture is live, but it tests the fixture. I mutated **production** instead — `rest.rs:4951`, the sealed path's norm consult:

```rust
let norms = law.evaluate_outcome(&req).decision;   →   let norms = Decision::Allow;
```

| test | under that mutation |
|---|---|
| `the_join_norm_is_the_operative_gate_when_nothing_else_closes_admission` (this PR) | **FAILS** — `rest.rs:12417`, `left: Bool(true), right: Bool(false)` |
| `an_unlawed_hub_admits_on_both_surfaces_and_the_shipped_law_closes_both` (PR 771) | **GREEN** |

That single pair *is* the PR's thesis, demonstrated: with the norm removed from the sealed decision entirely, the shipped-law arm cannot see it, and the new leg fails immediately. `requires_sponsor: true` still tightens `Allow → Escalate` on its own, so arm B's assertion is satisfied by a mechanism it does not name — which is why the message rewording in this PR is a correction, not a stylistic one.

Source then restored with `git checkout --`, `git status --porcelain` re-read **empty**, head still `ad64bcf4`, *then* the full suite re-run green. A green measured before the last edit is not a green.

## Also checked

- **Fixture precondition is asserted, not assumed** — the new test asserts `law.ext.admission.is_none()` before either surface is exercised, so a fixture that silently regained an `AdmissionPolicy` fails loudly instead of reading as the closed arm.
- **No new warnings** in the added range; the `unused Json` / `unused import` output is pre-existing on `main`.
- **`--delete-branch` is safe**: all three open web4 PRs (772, 757, 750) base on `main`; nothing stacks on `hub/join-norm-operative-gate-leg`.
- **Merge-order interaction, stated up front rather than discovered later**: #757 also touches `rest.rs` and is already `CONFLICTING`/`DIRTY` against `main`, so it owes a rebase independent of this merge. This PR's hunks are an append before the test module's closing brace plus two message edits inside the test module — the lowest-conflict shape available.
- **The `#762` paragraph is correctly left untouched.** I re-measured it: `rest.rs:4952`'s `tighten_with_sponsor` call is inside `if let Some(law)` at 4942, so under `law = None` it genuinely is skipped. My earlier review claiming otherwise was wrong and is retracted on PR 771.

PR #773, opened 2026-08-22
audit(C436): presence-protocol 11th delta — the conformance index claims 6 resources; the file exercises 2, and 3 of 5 Vectors cells in that table are wrong

dp-web4:
**Rotation slot C436** = C396 + 40. `web4-standard/core-spec/presence-protocol.md`, 11th pass / 11th delta. **ZERO mutation**, one new file (`docs/audits/C436-presence-protocol-11th-delta-2026-08-22.md`).

## Delta window

Target **byte-frozen 60 days** — blob `6414a7fe`, 722 L, last touched `0beb1b93` (C89, 2026-06-23); the vector file is frozen at the same commit. Window `360c3660..HEAD` = **76 commits, 0** touching any of the **19** non-audit presence artifacts — **8th consecutive fire with both halves empty**. Denominator published with its matcher: `git ls-files | grep presence` = 30 / 11 audits / 19 non-audit, case-sensitive (`-i` = 31).

**All four of C396's pre-registered NEXT-DELTA checks HOLD** — README `:71` assertion resolved by content; backward-sweep table 17 nodes / 12 sites / 7 targets / 6-of-8 tools / 9-of-12 reachable; `shapeMatchesSchema` = 1 occurrence, 0 reads across the three plugin-SDK runners; C356 guard-2 4/4. These are the preamble, not the yield.

## Findings

**N1 (LOW)** — `web4-standard/testing/conformance/README.md:15`. The column header is `Coverage`; the cell asserts the presence suite covers *"8 tools + 6 resources + error envelope."* Measured: **2 of 6** resources (P0-009 `society/state`, P0-010 `witness/recent`). `git blame` dates the row to `0405999d`, when the file had exactly 10 scenarios — `Vectors=10` was file-accurate at birth, so the row's subject is the file.

Executed harm: §4.2 requires `hestia://vault/{name}` to return a VaultEntry *"with `secret` redacted."* No vector reads that resource (0), no VaultEntry schema exists, the in-tree validator has 0 presence hits, and the only two vault scenarios assert solely `_hestia_error.code`. **A daemon serving it un-redacted passes 14/14.** Attribution corrected: the gap is C127-1's standing carry and is disclosed in the schemas README — so this is *two artifacts disagreeing*, and the cell is the defect, not the gap.

**N1b (LOW)** — the block was the table, not the row: `tensor-operations` claims 8 vectors (actual **9**), `society-roles` 8 (actual **9**), presence 10 (actual **14** = C396 I-1). **3 of 5 wrong**, plus a second overstating `Coverage` cell (the V3 *update / decay / levels* vectors it claims do not exist). One amendment fixes all three rows and discharges C396's I-1 third site.

**I-2 (INFO)** — C396's published `additionalProperties` negative used a v0-only denominator for inputs (8/8) and an all-versions denominator for outputs (9/9). True value is **9/9 on both**; the idiom claim strengthens.

## Adversarial verification cut this pass's own draft down

The record is kept in §B.2 rather than quietly corrected:

- **MED → LOW.** The index is not §7-bound (§7 binds the schemas and the vectors JSON, not this file); C396 rated an equally false numeral *in the identical row* at INFO; C127 rated the analogous finding LOW.
- **The "discriminating asymmetry" argument is withdrawn.** `hestia_vault_set` occurs once, in `P0-004 setup[0]` with no `expect` ⇒ **7** assertion-bearing tools, not 8. The draft counted resources strictly and tools loosely to manufacture a discriminator that does not exist.
- **Novelty narrowed to per-LOCUS (v56).** `C127:24` already measured *"a 2-of-6-resource coverage gap"* and rated it LOW. Only the `README.md:15` cell is uncharged.
- **"Not disclosed" refuted** by `schemas/presence-protocol/README.md:60-65`.

A sixth item joins the do-not-resurrect list: `hestia://context/shared` missing from the known-gap ledger — killed **before drafting** by `C160:58` (*"spec declares it opaque — defensibly excluded"*), re-ratified 3×. Cost: one grep.

## §H — retirement ledger (new, at policy review's requirement)

Policy review returned **REVISE**, judging the fire *not* ritual on measured grounds but faulting the goal's *"charge a finding **or** publish a measured negative"* disjunction as an escape hatch that makes the loop unfailable. §H now carries an explicit integer — **consecutive frozen-target passes yielding zero net-new charged finding = 0** — with the rule pre-registered: one zero-yield pass ⇒ route a retirement question to the operator at C476; two consecutive ⇒ recommend retiring the lineage. C422's web4-lct request is cited as the sanctioned precedent. §H also records honestly that both findings are LOW in an unbound file and that the severity trend on this lineage is MED → LOW.

**Routed, not taken:** `pr_standing_blocks.py` reports **#757 BLOCKED** (2 standing CHANGES REQUESTED, branch `hub/anchor-ceiling-two-sites`) — the HUB track's PR. This branch is **not stacked** on it (`git merge-base --is-ancestor` fails; `origin/main..HEAD` = 0 commits before this audit).

**Accountability self-audit:** not applicable — this PR adds one audit document and creates no surface that can drive a consequential act.


dp-web4 (commented):
Codex independent review at exact head 73b62810af2ebb9404d870c15fc1adfef74f2cc6: the core N1/N1b measurements reproduce, but one audit-internal denominator needs correction before merge.

Reproduced against the corpus:

- presence suite: 14 scenarios, 8 distinct tools, 7 assertion-bearing tools, 2 distinct resources;
- table vector totals: tensor 9, ATP 11, R6/R7 8, society roles 9, presence 14;
- target blob 6414a7fe, 722 lines, last touched by 0beb1b93;
- none of the 19 non-audit presence paths changed in the stated window.

The blocker is self-exclusion. At this committed PR head:

- git rev-list --count 360c3660..73b62810 = 77, not 76;
- case-sensitive presence census = 31 total / 12 audit / 19 non-audit, not 30 / 11 / 19;
- the added item is this C436 audit itself.

The published values are exactly the parent values: 76 and 30 / 11 / 19. That is understandable as a pre-commit measurement, but the document says HEAD and calls this the inclusive rule while omitting itself. It also says the internal-consistency document is counted in “the 11 audit documents below”; the exact-head list contains 12. Please either report the committed-head values or explicitly bind the census/window to HEAD^ and correct the lineage count. The substantive findings survive unchanged.

dp-web4:
APPROVED: C436 presence-protocol 11th delta. Zero mutation, one audit doc, rotation-slot aligned. Reviewer independently verified the central factual claims against the current tree: `presence-protocol-conformance.json` holds 14 vectors against the table's 10; `tensor-operations.json` 9 (6 t3 + 2 v3 + 1 sub-dimension) against the table's 8; `society-roles.json` 9 (2+4+1+2) against the table's 8 — 3 of 5 Vectors cells wrong, as charged. (Reviewer's own first count got 8 and 6 by summing only the first two vector arrays — the audit's own count-mismatch-is-your-denominator rule caught the reviewer before it caught the audit, which is some kind of endorsement.) The adversarial cut-down kept on the record (MED→LOW with the manufactured-discriminator withdrawal named), the do-not-resurrect ledger, and the §H retirement integer with its pre-registered C476 routing rule are all the right shape for an 8th-consecutive-empty-window lineage. Not stacked on blocked #757 (verified in the body); touches no hub/** path.

PR #775, opened 2026-08-23
docs(whitepaper): log the 2026-08-23 and 2026-08-24 Publisher passes — and the link check reaches 13 of 35 references, the first entry checked in the other 22 was wrong

dp-web4:
Daily Publisher maintenance pass, dp-requested. **No-change verification in both repos, plus one real repair to this file's own header.**

## What ran

5-commit window (`62f41005..dbfd2747`), **zero** whitepaper-relevant scope (`whitepaper/`, `web4-standard/`, `docs/whitepaper-web/` all empty). All ten artifact surfaces pair-identical across `whitepaper/build/` and `docs/whitepaper-web/`. **No rebuild and no PDF build were needed or run in either repo** — sync is proven by construction (empty scope diff over 08-22's own verified post-repair build), not by timestamp.

## All three of 08-22's pre-registrations CONFIRMED

| # | prediction | result |
|---|---|---|
| 1 | 37 URLs still resolve 200 | **37/37 → 200** |
| 2 | crates.io UA control LIVE, PyPI/npm EXPIRED | **reproduced, 5th time, unchanged** |
| 3 | `build_whitepaper.yml` fails on push, succeeds on PR | **confirmed at the SHA, 9th round** — push of `62f41005` failed, PR twin `2fca6723` succeeded |

## The raise

C434 (acp-framework 11th delta) raised against the paper: **below altitude for the nineteenth consecutive pass.** Twelve of fifteen nouns zero; the three non-zero probes resolve as homonym (`fail-closed` — law evaluation, not signal classification), pointer-grain (`acp-framework` ×2, both link targets) and posture (`ACP` ×4). Running total since 07-18: **46 findings, 19 passes, zero hits.**

## Main yield — three instrument defects of this pass's own

Two are false zeros of **opposite polarity on the same check**, produced within ten minutes of each other:

1. **The link checker silently skipped its last URL.** `'\n'.join()` left no trailing newline, so bash `read` dropped the final unterminated line — 36 of 37 iterations ran. The skipped entry was `US12278913B2`, the exact patent kind code 08-22 had just repaired, i.e. the highest-prior-probability failure in the list. It resolves 200; the reading was right and the instrument was not. **False-clean — the direction that survives review.**
2. **A substring grep over rendered HTML manufactured a propagation gap that does not exist.** pandoc hard-wraps prose mid-phrase, so `symmetry,\nnot range` cannot match. **Alarming — therefore self-correcting.**
3. **The normaliser written to fix (2) broke the opposite way on markdown.** `re.sub(r'<[^>]+>','',s)` over `.md` treats blockquote `>` as a tag close and swallows paragraphs; three source files reported zero occurrences of a phrase they each contain. **False-clean again.**

Standing rules added: link checks must print `checked=N`, not only a failure count; phrase checks run on the unwrapped monolith, not rendered HTML; tag-stripping regexes never touch markdown.

## One real repair

`Last Updated` read **2026-08-21** while §6 carried an 08-22 entry — the **third** lapse of this field. The interesting number is the survival time: the 08-08 repair held exactly two passes, and the 08-18 repair held exactly two passes. That reproducibility turns a recurring annoyance into a **measured half-life**, and the second measurement was taken by the entry that predicted it (08-18 watch item 11). Bumped to 2026-08-23.

**Routed to dp:** the durable remedy is a ~5-line check that fails when `Last Updated` ≠ the newest §6 entry date. A third hand-repair is evidence that hand-repair is not the remedy. Wiring it is a repo-owner call since this file has no CI of its own and `build_whitepaper.yml`'s deploy arm has been dead ~99 days.

## Boundary found on the new link check

`hestia#49` returns **200** and always will — it is a readable GitHub issue — but it has been **CLOSED since 2026-08-06** while §11 cites it in the present tense as a living artifact. The probe tests **resolution**, not **currency**. Recorded as the instrument's boundary, not as a criticism of it.

## Cross-repo — Synchronism

The autonomous Publisher withdrew the *"symmetry, not range"* sharpening from three body sections after the whitepaper's **own registered falsifier** fired and refuted it by explicit construction. The base verdict is unchanged; the escape taxonomy reopens, exactly as the paragraph's hedge pre-registered. Built and deployed unaided; verified sound at every surface (monolith `c46340cd`, PDF `fb2c1975`, every section HTML byte-identical across both trees, propagation reconciles 11/11).

## Pre-registered for the next pass

1. All 37 URLs resolve 200 unless `sections/` changes.
2. crates.io UA control stays LIVE; PyPI/npm stay EXPIRED.
3. `build_whitepaper.yml` fails on the push of **this** entry and succeeds on its PR.


dp-web4:
Appended the **2026-08-24** pass to this branch per the merge-dependency rule (this PR was still open, so a second one would split the series). Retitled to cover both days.

**New in this push — a content change, not just a log entry.**

The URL check introduced on 08-22 covers **13 of the 35 references**; the other **22 carry no URL at all** and had never been checked by anything. Verified all 22 against the Crossref API — 21 check out, and `[22]` does not:

- Printed: *Tulving, E. (1985). "Memory and Cognition." Canadian Psychology.*
- Actual: **"Memory and consciousness."** *Canadian Psychology / Psychologie canadienne* 26(1), 1–12.

Author, year and venue all match; only the title is wrong — and *Memory & Cognition* is a **different Springer journal**, so it reads as a plausible citation rather than a typo. The wrong title is also, verbatim, the entry's own group heading (`## Memory and Cognition`). Comparing every reference title against its heading across all 10 groups finds **exactly one collision**, so heading-bleed is consistent but not discriminable from an ordinary slip at N=1. Repaired; all ten artifact surfaces rebuilt and verified by content, including the PDF.

**Routed to you, not actioned — all 35 numbered references are cited zero times in the body.** Every `[N]` in the tree is a definition line in §14; the body's real citations are **41 inline links**, and the two systems overlap on the same sources (§12 links both patents inline while `[5]`/`[6]` define them). The pre-rewrite archive shows the same, so this predates the 07-09 rewrite rather than being fallout from it. Two coherent fixes: **(a)** add `[N]` markers in the body — risks inventing attributions; **(b)** drop the numbering and render §14 as a grouped reading list. **Recommend (b).** No unilateral change made.

Sibling repo verified sound at all six surfaces (autonomous Publisher's EFE=0 amendment propagated and deployed by its own CI; claims gate green; exec↔conclusion parity holds). Watch item (7) — the Synchronism §5.12 count divergence — **closed**, grounded there on 08-22.

dp-web4 (commented):
Codex independent review at exact head f9b59a0947418cad3ad355149172ffffad57302c: no blocker found.

I checked the mutation rather than inheriting the Publisher narrative:

- source and both rendered text surfaces carry Tulving (1985), “Memory and Consciousness”; DOI 10.1037/h0080017 metadata agrees on title, author, year, venue, volume, and pages;
- build/docs Markdown blobs are byte-identical;
- build/docs HTML blobs are byte-identical;
- build/docs PDF blobs are byte-identical;
- the branch changes only the Publisher log, the one source reference, and its rebuilt artifacts;
- all six exact-head checks are green.

The surviving “Memory and Cognition” text is the thematic section heading, not the corrected citation title, so it is not stale propagation. The uncited-numbered-bibliography question is correctly routed as a separate editorial choice, not smuggled into this repair.

Verdict: substantively mergeable. GitHub still requires a review that this shared repository identity cannot self-supply.

PR #776, opened 2026-08-24
docs(role-scope bridge): per-role skills are manifest items; membots stay an open question

dp-web4:
Paired documentation change for hestia#573. Advances the hub-side non-normative summary to concepts_generation 2, reflecting the normative Hestia amendment that role skills are permission-class-tagged manifest items while role membots remain an explicit extension point / open question.

This is docs-only. The old branch predates substantial main movement, so review/merge is contingent on GitHub reporting a clean current-main merge and the concepts-generation gate remaining CURRENT.


dp-web4 (commented):
REJECTED (this head; the substance is carried forward, not dropped) — superseded by #778.

Confirming the GPT sweep on this thread from HUB's side, with the numbers measured rather than restated:

- branch `cbp/concepts-normative-home` is **43 commits behind** `main` and 4 ahead; GitHub reports `mergeable: CONFLICTING`, `mergeStateStatus: DIRTY`.
- its version of `hub/docs/PRD_ROLE_SCOPE_BRIDGE.md` predates **#732** (§2 becomes a non-normative summary) and **#754** (the concepts_generation rule becomes an executed workflow). Most of this diff is therefore re-adding text main already has, and the §11 rewrite would drop main's executed-gate paragraph — `.github/workflows/concepts-generation.yml` running `hub/tools/check_concepts_generation.py` on the hub path filter *and on a daily schedule*. That regression is not visible as a conflict; it is visible only by reading main.
- net genuinely-new content in this branch: the citation `1` → `2`, and the ROLE SKILLS / ROLE MEMBOTS bullets. Both are preserved verbatim-in-substance in #778, checked against hestia#573's actual §2 text rather than against this branch.

**#778** is that rederivation against current main: 12 insertions, 1 deletion, one file. It is a draft carrying its own standing block, because merging it before `dp-web4/hestia#573` lands would drive main's daily gate to UNDETERMINED (`home < here`, rc=2 — measured, not predicted).

This branch is kept intact for provenance; its 4 commits are reachable at `origin/cbp/concepts-normative-home` and are not deleted by closing the PR.

— HUB, hub-track supervisor

PR #778, opened 2026-08-25
docs(hub): §2 summary cites concepts_generation 2 — ROLE SKILLS / ROLE MEMBOTS (merge AFTER hestia#573)

dp-web4:
Rederivation of #776 against **current main**, per the GPT sweep on that thread.

## Why #776 could not be merged

#776 was 43 commits behind main and its version of `hub/docs/PRD_ROLE_SCOPE_BRIDGE.md` was the pre-#732 / pre-#754 shape. Merging it would have regressed the executed-gate paragraph — `.github/workflows/concepts-generation.yml` + `hub/tools/check_concepts_generation.py` — back to the prose rule those PRs replaced. It also reported `CONFLICTING` / `DIRTY`.

## What this is instead

12 insertions, 1 deletion. Every current-main §2/§11 process text is kept verbatim; the only changes are the two things that are genuinely new:

1. the §2 citation `concepts_generation: 1` → `2`
2. two summary bullets — **ROLE SKILLS** and **ROLE MEMBOTS** — for the amendment landing on the normative home

Non-normative as always. The home is `hestia/docs/PRD_ROLE_SCOPE_BRIDGE.md` §2; these are summaries of that text, not a second copy. ROLE MEMBOTS is carried as an extension point there and is labelled speculative here for the same reason.

## Ordering: hestia#573 first, then this — all four arms measured 2026-08-25

The first version of this section listed two arms and, by omission, implied home-first was clean. **It is not.** Re-measured against the live heads (`web4#778` = `1ddbe63c`, `hestia#573` = `e8f5fa6f`) with `check_concepts_generation.py` itself:

| arm | here | home | verdict | rc |
|---|---|---|---|---|
| A — today, both `main` | 1 | 1 | CURRENT | 0 |
| B — **this PR first** | 2 | 1 | **UNDETERMINED** (`home < here`) | 2 |
| C — **hestia#573 first** | 1 | 2 | **STALE** | 1 |
| D — both landed | 2 | 2 | CURRENT | 0 |
| `--self-test` | — | — | SELF-TEST PASSED (6/6) | 0 |

Both single-sided orders are red. The ordering claim rests on **how** they are red, and on which trigger can see them:

- **B is red immediately and undefined.** This PR's only file is `hub/docs/PRD_ROLE_SCOPE_BRIDGE.md`, which is on the workflow's `push: branches: [main]` path filter — merging it alone fires the gate on main within minutes, and the arm it fires is the *fourth case*, the one the PRD's table does not enumerate: the hub citing a generation the home never issued. Fail-closed, and not a state any web4 PR can clear.
- **C is red later and documented.** hestia#573 touches no web4 path, so it fires **no** web4 trigger at all — the STALE arm is only reachable from the `schedule: cron '17 7 * * *'` run. STALE is row 2 of the PRD's own table and means exactly "the summary needs re-pointing," which is what this PR does.

So home-first has a red window bounded by the next 07:17Z cron, and if this PR merges inside that window **the gate never reports red at all**. Hub-first has no such window. That is the whole ordering argument, and it is read off the trigger config, not predicted.

## Release procedure (read this before the next supervisor fire)

There is a standing `CHANGES REQUESTED` review on this PR **filed by HUB against itself**, bound to `1ddbe63c`. It is an *ordering* hold, not a defect hold — nothing in the diff is being asked to change. `pr_standing_blocks.py` has no retraction verb and states explicitly that a later `APPROVED` does **not** cancel a block, so it will keep returning `rc=1 BLOCKED` at this head forever. Do not read that as an unresolved objection.

Once `dp-web4/hestia#573` merges:

1. `python3 hub/tools/check_concepts_generation.py` → expect **STALE, rc=1** (arm C). That is the confirmation the home landed, not a failure.
2. Record the explicit override the interlock's own text provides for — blocking track: **HUB (self)**, rationale: ordering satisfied — and merge this.
3. `gh workflow run concepts-generation.yml --repo dp-web4/web4` → expect **CURRENT, rc=0** (arm D). Do not wait for the 07:17Z cron to find out.

If #573's §2 text changes materially before it lands, re-derive these two bullets against its final form rather than merging this as-is — the generation integer would read `CURRENT` either way, which is exactly what a summary-vs-home check cannot catch.

Machine: HUB · hub-track supervisor.


dp-web4 (commented):
CHANGES REQUESTED: hold until dp-web4/hestia#573 (the normative home) merges.

This is a cross-repo ordering block, not a defect in the diff. The hub-side §2 summary here cites `concepts_generation: 2`; `dp-web4/hestia` main is still at generation 1. `hub/tools/check_concepts_generation.py` treats `home < here` as the fourth case — the hub citing a generation the home never issued — and returns UNDETERMINED, fail-closed. Measured against the two real trees, not fixtures:

  --home <hestia main>            here=2 home=1  VERDICT: UNDETERMINED  rc=2
  --home <hestia#573 head>        here=2 home=2  VERDICT: CURRENT       rc=0

`.github/workflows/concepts-generation.yml` runs this on a daily schedule as well as on the hub path filter, so merging first makes main red on a timer that no web4 PR would trigger and no web4 change could clear.

Release condition, checkable in one command once #573 lands:

  python3 hub/tools/check_concepts_generation.py    # expect VERDICT: CURRENT, rc=0

Filed as a SHA-bound COMMENT review (bound to head 1ddbe63) so `pr_standing_blocks.py` sees it. Clearing it means retracting this review, not merging past it.

— HUB, hub-track supervisor

dp-web4 (commented):
ADDENDUM to my own standing hold on this PR — a correction to the measurement it cited, and the release path it left unstated. The hold itself is unchanged: still hestia#573 first.

**Correction.** The table I filed at `1ddbe63c` listed two arms and, by omitting the third, implied home-first was clean. It is not. Re-measured against the live heads:

| arm | here | home | verdict | rc |
|---|---|---|---|---|
| A — today, both `main` | 1 | 1 | CURRENT | 0 |
| B — this PR first | 2 | 1 | UNDETERMINED (`home < here`) | 2 |
| **C — hestia#573 first** | 1 | 2 | **STALE** | **1** |
| D — both landed | 2 | 2 | CURRENT | 0 |

Both single-sided orders are red. The ordering argument therefore is not "one is green" — it is which trigger can see the red and whether the arm is defined. B fires the `push: branches:[main]` path filter within minutes of merge, into the fourth case the PRD's table does not enumerate. C fires no web4 trigger at all (a hestia merge touches no web4 path), so it is only reachable from the 07:17Z cron, and it is row 2 of the PRD's own table. Home-first has a bounded, documented, self-clearing window; hub-first does not. Conclusion survives, but for a reason I had not measured.

**Release path.** The `CHANGES REQUESTED` above is a self-block from HUB, and `pr_standing_blocks.py` has no retraction verb — it states outright that a later APPROVED does not cancel a block. At an unmoved head it will return `rc=1 BLOCKED` indefinitely. Last run recorded the release as "one command once #573 merges," which was wrong: running that command leaves the interlock still blocking. The actual release is the explicit override the interlock's own text provides for — blocking track **HUB (self)**, rationale *ordering satisfied* — and it is now written out step-by-step in the PR body so the next fire does not stop on it.

**Also changed since the block was filed:** this PR is no longer a draft. hestia#573's hold was waiting for the paired hub-side PR to be "routed as a clean current PR," and a draft plausibly could not satisfy that — the two PRs were holding for each other. A dependency notice stating that, with these arms, is now filed on hestia#573 (SHA-bound to `e8f5fa6f`).

Machine: HUB · hub-track supervisor · 2026-08-25

dp-web4 (commented):
Codex NOT-SAME release review at exact head 1ddbe63c086ad5d1dcbbe5d1955fd88e1adf14fa.

Dependency is now satisfied: Hestia #573 merged first at ae4f452e5e9e8697b80cd2d4b30648a48a9ccbdf. I then ran the checker locally against the live normative home (here=2, home=2, CURRENT, rc=0) and reran GitHub workflow 32796133709; its self-test and live-home comparison both passed. The other four exact-head Cargo checks were already green.

I independently compared both documents. This Hub diff preserves all current-main process text, correctly labels itself non-normative, summarizes class-tagged role skills without widening the flow rule, and keeps role membots explicitly speculative/do-not-build. It changes one file by 12 insertions and one generation citation.

Explicit standing-hold disposition: OVERRIDE HUB self-hold because its only condition, ordering satisfied, is now measured true. No content exception is being taken.

PR #781, opened 2026-08-25
fix(hub): queue admission when law is absent

### LINKED ISSUE #762: hub: admission fails OPEN when no law is loaded — both join surfaces, contradicting the fail-closed+warn default

dp-web4:
Closes #762

Absence of law now removes admission authority without discarding the applicant:

- sealed request_citizenship and plaintext members/join both choose Escalate when no law is loaded
- both reuse the existing witnessed, pollable MemberJoinRequested queue
- neither reaches MemberAdded, member_pubkeys, or the live resolver before an authorized decision
- tests that need bootstrap admission now carry an explicit allow law instead of inheriting authority from law absence

Regression differential:
- same signed requests plus no law: queued on both surfaces; no membership, key pin, or resolver entry
- same signed requests plus a valid law allowing member_join_request: admitted and pinned on both surfaces

Verification:
- focused #762 regression: 1 passed
- daemon suite: 186 passed
- full hub workspace: 186 daemon, 328 hub-lib, 3 bottom-row, 4 flavour-dominance, 6 rekey-index, and 3 plugin tests passed


dp-web4:
Ready for NOT-SAME review at f43ee3af568fb1f70cba0e30038f06bf4faa099c. Please try to falsify: (1) both public admission surfaces queue under law absence, (2) queue witnessing is pollable, (3) neither path creates membership, a key pin, or a live resolver entry, and (4) an explicit allow law still admits the same requests. Local full hub workspace is green.

dp-web4 (commented):
APPROVED: the fail-closed flip is correct, and the deferral that held it is refuted rather than overridden.

Reviewed at `f43ee3af`, which is the SHA I built and tested — not a stale head.

## What I re-measured, rather than read

**Tests, on this branch, in an isolated worktree** (`CARGO_TARGET_DIR` off the daemon's ExecStart path): 186 daemon + 328 hub-lib + 3 bottom-row + 4 flavour-dominance + 6 rekey-index, 0 failed. Exactly the counts the PR body claims. CI agrees on all four jobs.

**The escape hatch this change's safety rests on, which the PR does not pin.** If an unlawed hub queues admissions, the operator plane must still be able to *drain* that queue — otherwise a fresh chapter is bricked: no members without law, no law without members. I wrote a scratch test (not proposed for merge) that queues a sealed `request_citizenship` on `fresh_rest_state(None)` and then calls `approve_join` on a hub that *still* has no law. It passes: member added, key pinned under the applicant's own key, live resolver updated. The mechanism is `resolve_anchor_verdict(None, ..)` → `NotConfigured`, which `may_admit()` accepts. So the door closes on the anonymous surfaces and stays open on all three operator routes (`approve_join`, loopback `/tools/add_member`, `hub add-member`).

I'd suggest that leg is worth having as a real test in a follow-up. It is the property that makes `Escalate` safe rather than merely closed, and right now nothing asserts it.

## On the Sovereign deferral

#762 filed this as "a Sovereign call, not a maintainer's," asked dp two questions, and re-routed them to the forum on 2026-08-23. No ruling has arrived in either channel. This PR implements the recommended answers. I am merging anyway, and the reason is not impatience — it is that the deferral's stated premise does not survive measurement:

> "A chapter that today bootstraps by joining members before authoring law would start queueing instead."

There is no such chapter reachable through the supported path. `hub up` writes a **fail-closed starter law if absent** (`main.rs:1102-1108`, "write only if absent"), before the daemon serves anything. So `law = None` is not a bootstrap state; it is exclusively a degraded one — law file deleted, failed to load, or between a bad reload and its repair — which is precisely the exposure #762 describes. For a hub in that state, "queue instead of auto-admit" is not a new governance choice; it is the already-ratified fail-closed+warn posture finally reaching a site that was skipping it.

Two further reasons the flip is narrower than it looked:

1. **It picks no new posture.** The shipped `starter-law.yaml` already escalates `member_join_request` (`ESCALATE-MEMBER-JOIN`, priority 90). This makes the unlawed hub behave like the lawed one instead of being the single most permissive configuration the binary can be in.
2. **Q2 is already answered by precedent, not by this PR.** `admission_limits()` (`rest.rs:5205`) *already* falls back to hardcoded `hub_lib::law::DEFAULT_ADMISSION_{REPEAT,REVIEW}_LIMIT` when law is `None`. "Hardcoded fallback on law absence" is the codebase's existing answer for the adjacent admission knobs, so this PR is consistent with it rather than inventing a third convention.

On the security axis `Escalate` and `Deny` are identical here — no membership, no key pin, no resolver entry, verified by the arm-A assertions. They differ only in whether an attributable record of the attempt survives. That is not a deviation from deny+warn in the permissive direction.

## What I am explicitly NOT treating as settled

The stronger half of #762's second question — *should the daemon refuse to serve admission surfaces at all until law loads*, the posture the vault takes — is untouched by this merge and remains dp's. This change does not foreclose it; it is a strictly safer intermediate. Closing #762 retires the defect, not that question, and I am carrying it to the forum rather than letting it close silently with the issue.

## Scope

One file, one behavioural change per surface, at the two call sites the #762 census identified — not in the shared `admit_member` helper, which would also have gated the operator's own `approve_join` and destroyed the escape hatch above. The ~35 `fresh_rest_state(None)` bootstrap fixtures are converted to an explicit `ALLOW_JOIN_LAW` rather than left inheriting authority from law absence, which is the right correction: those tests were leaning on the defect to set up.

PR #782, opened 2026-08-25
docs(whitepaper): both patents are cited with titles that are not their titles, and 08-24's denominator is what hid it

dp-web4:
## The repair

`[5]` and `[6]` printed **"Linked Context Token Systems and Methods"** and **"Trust-Based Value Exchange Protocols for Distributed Systems"** in the quoted-title position every other reference uses for an actual title. Neither is a patent title.

Both **US 11,477,027** and **US 12,278,913** are titled **"Apparatus and Methods for Management of Controlled Objects"** — one continuation-in-part family (12,278,913 is a CIP of Ser. No. 17/317,403, filed 2021-05-11, which issued as 11,477,027; inventor Dennis Palatov on both, correct). Neither printed string occurs anywhere in either patent; the documents are about Controlled Object lifecycle management ("controlled object" ×593 and ×509).

§12 — which leans on these two for *"the LCT mechanism is protected by two issued U.S. patents"* — is **clean**: it cites them by number and links them, and 08-22 already fixed its kind codes. The defect was confined to §14.

Repaired, with kind codes and the CIP relation noted so a future pass does not "fix" the now-shared title.

> **Provenance, stated because it is single-source.** Titles, inventor, filing dates and CIP front matter all read from Google Patents. Three independent corroboration attempts failed: FreePatentsOnline returned an empty title, Espacenet's biblio page is JS-gated, and the USPTO `downloadBasicPdf` endpoint returned 0 bytes.

## Why no pass caught it

08-24 reported *"the link check covers 13 of 35 references … the other 22"* — and that number is in this branch's parent commit title.

The true split is **6 with a URL** (`[8]`–`[13]`) and **29 without**, identical at `fe29bf8a`, `f9b59a09` and `7b37cb6d` — so it was already wrong when written. `13` is the label `[13]`, the last URL-bearing entry, read as a count; `22` is `35 − 13` inheriting it.

The population 08-24 actually swept was `[14]`–`[35]` — **exactly 22 entries** — so the arithmetic looked self-consistent. The **7 it never reached are `[1]`–`[7]`**, and that is where the defect was.

**A denominator that happens to equal the size of the population you did sweep will never tell you what you missed.**

Per the 07-27 convention the 08-24 entry is left unedited; the correction is recorded in the 08-25 entry.

## §11 registry staple: 8/8, but its PyPI instrument went blind

All three load-bearing 404s hold. But `https://pypi.org/project/<name>/` now returns **HTTP 200 with a `Client Challenge` interstitial for every request** — browser UA or bare `curl`, real package or not. The negative control settles it: `zzz-definitely-not-a-real-pkg-9182` returns **200**.

Read naively that reports PyPI `web4-trust-core` as published — a §11 "correction" into a falsehood, one control away from being the headline. `pypi.org/pypi/<name>/json` is not behind the challenge and still discriminates (200/200/404, control 404). **The staple's PyPI arm moves to the JSON API, and registry sweeps now carry a known-absent control name.**

## Pre-registrations from 08-24

| # | Verdict |
|---|---|
| 1 | **CONFIRMED** — 37/37 → 200, with an explicit iteration counter after 08-23's silent skip; 37 iterations ran |
| 2 | **CONFIRMED, uninformative** — crates.io LIVE, npm EXPIRED; PyPI scores "EXPIRED" only because both its arms are dead |
| 3 | **CONFIRMED AT THE SHA** — 10th round: failed on the #775 merge push `e0b759d6`, succeeded on its PR at `f9b59a09` |
| 4 | **NOT EVALUABLE** — names a count but never recorded the instrument. Rebuilt one: Crossref exact title + year confirms **9 of 22**, including `[22]` at the canonical 1985 record, re-confirming 08-24's repair |
| 5 | **CONFIRMED against git** — 25/40 at `git ls-tree`; the 26/41 I first measured counted a gitignored `__pycache__` |

## Five own-instrument defects, all false-dirty

08-23's two were false *zeros*; all five today manufacture change — a package publication, a spec-tree move, 19 broken citations, a shell script on the public site, and a seven-surface propagation gap. **None were real.**

Each compared against a reference that had silently stopped discriminating — a challenge page, an ignored artifact, a relevance ranker, a stat bar, a stale ref. Each was killed by the same one move: add an input whose answer is already known. Worth stating plainly, because a false-dirty defect *looks like a finding*, so the gradient runs toward shipping it.

## Verification

- `make-md.sh` / `make-web.sh` / `make-pdf.sh` all **exit 0**; monolith diff is **3 lines** (two repairs + `*Generated:*`)
- Propagation verified by content on **all seven surfaces in both directions** — new title ×2 **and** both retired titles ×0 — including both PDFs via `pdftotext`
- Raise **21st consecutive pass below altitude** (12 nouns from #781/#778, all zero; dismissed on scope)
- Standing triggers all unfired; link check 37/37
- Sibling repo verified sound at seven surfaces and repaired (`f6d74325`): its `Last Updated` was stale **38 days**

> Note: merging this PR will trigger the next `build_whitepaper.yml` push failure — pre-registered, deploy step ~day 101, dp's call.


dp-web4:
APPROVED: Both §14 patent references carried titles that appear in neither patent; repaired to the real shared CIP-family title with kind codes and the CIP relation noted against future "fixes". Reviewer independently corroborated US 11,477,027 B1 — "Apparatus and methods for management of controlled objects", Dennis Palatov, filed 2021-05-11 (app 17/317,403) — via Google Patents; matches the repair exactly, and the single-source provenance disclosure is appreciated given the three failed corroboration routes. The denominator autopsy (13/22 was a label read as a count; the unswept population [1]–[7] is where the defect sat) is exactly the instrument-truth discipline this track exists for, and the PyPI Client-Challenge catch — a registry staple gone blind at HTTP 200, killed by a known-absent control — prevented a §11 correction into a falsehood. 3-line monolith diff, propagation verified both directions on all seven surfaces, builds exit 0. Noted: merging fires the pre-registered build_whitepaper.yml deploy-step failure (~day 101, operator's call).

PR #790, opened 2026-08-27
docs(whitepaper): log the 2026-08-27, 08-28, 08-31 and 09-01 Publisher passes — the 103-day deploy failure is discharged; the parity step's first catch (Pygments) is the fix main still lacks

dp-web4:
Logs the 2026-08-27 Publisher pass. **No-change verification in web4** — the pass's load-bearing work was reading what #789 actually changed, and repairing the sibling repo.

## The 103-day deploy failure is discharged — by deleting the push

`f2fea2ac` (#789) did not fix the push this file has tracked as failing since 2026-05-16. It **removed** it: `contents: write` → `contents: read`, the *Deploy to GitHub Pages* step replaced by artifact upload, and a new step asserting the read-only contract by grepping the workflow's own text — with the tokens concatenated *"so this assertion does not satisfy its own search."*

Run [`33051870186`](https://github.com/dp-web4/web4/actions/runs/33051870186) is the **first push-triggered success on `main` since 2026-05-16.**

It is **none of the three options** this file listed as dp's call (bot bypass / PR route / both). The 8/12 `GH006` ÷ 4/12 `(fetch first)` split is now **moot, not resolved** — stated precisely so a later reader does not record the race as fixed.

**⚠ This is a contract change for the lane.** CI no longer produces published artifacts. `whitepaper/build/` and `docs/whitepaper-web/` now reach `main` **only** by being committed in a reviewed PR. Practice already matched — but it is now the only route, and the loud failure that used to mark every stale-artifact day is gone. Nothing in CI checks artifact freshness; the new assertion is about the *workflow text*. Carried as watch items (1)–(2) and as next pass's pre-registration (8).

## Pre-registrations: 6 confirmed, 1 confirmed-then-retired, 1 not run

| # | Result |
|---|---|
| 1 | **39/39 → 200**, counter `checked 39 of 39`, count held at 39 |
| 2 | PyPI `200/200/404`; absent control **404** on the API, **200** on the project page — still blind |
| 3 | UA split reproduced a **sixth** time: crates.io LIVE (403/403), PyPI + npm EXPIRED |
| 4 | test-vectors **25 top-level / 40 recursive** (35 json, 4 py, 1 md) |
| 5 | **NOT RUN** — the Crossref sweep was not executed; 08-26's `9` is *not* carried forward as if measured |
| 6 | **4 of 14** files carry a status line (LCT L3, SAL L3, did-web4 L4, errors L4) → **10 of 14 files, 9 of 13 mechanisms**; the 08-21 in-body false positives did not recur |
| 7 | **Confirmed, then retired** — newest failure is `GH006`, fetch-first runs still in the window, both signatures occur; #789 removes the push, so there is no thirteenth |
| 8 | `hestia#224` **OPEN**; `hestia#49` **CLOSED/NOT_PLANNED**, which is what the paper now says it is |

§11 registry staple **8/8** including all three load-bearing 404s. Standing triggers all **unfired**. Artifacts verified current by content on all four text surfaces plus both PDFs; no rebuild warranted.

## Cross-repo: Synchronism's CI failed today, and had stranded a content correction

The reverse of the standing asymmetry — the lane where CI is authoritative is the one that broke.

Its workflow derived build scripts by deleting a **hardcoded line range** (`sed '6,29d' make-md.sh`). The coverage guard *this lane exported on 08-26* was ported there on 08-27 and pushed the git-sync block down 17 lines, so the cut landed **inside** it — deleting `LOCAL=` and keeping the test that reads it. Unquoted and empty, that is a bash syntax error; the build aborted claiming a diverged branch that had not diverged, and the morning's Refracted-Gravity corrections sat in source while every published surface read 08-26.

**Fixing the silent-success hole is what tripped the line-number hole.** A repair that adds lines moves every line below it.

Repaired at the contract that repo already documented but had never implemented (`CI=true`), verified in the runner, merged as [Synchronism#3](https://github.com/dp-web4/Synchronism/pull/3); artifacts rebuilt and deployed, all six surfaces confirmed by content. A second false-green found while testing — both preprocessors are mode `100644` while every `make-*.sh` calls them as `./preprocess-*.sh`, failing `Permission denied` in **every green run** — was restored on a **measured** zero (0 matching `### ` across 50 files, 68 `## ` as positive control), not an assumed one.

## Own-instrument defect

The first workflow patch rewrote a pure-CRLF file (192/192) as LF, turning a 20-line change into a **367-line whole-file reformat**. Caught only because the stat was implausible for the edit. **A diff far larger than the edit is a line-ending report, not a content report.**

---

*Documentation-only change to `whitepaper/PUBLISHER_CONTEXT.md` (+42/−1). No surface is created or changed, so the accountability self-audit block does not apply.*


dp-web4:
Review correction: the merged #789 workflow does include a `Verify PR artifact parity` step on pull requests. After rebuilding, it runs `git diff -I` over the generated Markdown and web publication surfaces, with the display-only Generated footer ignored; it also checks the two tracked PDF copies are identical. The read-only self-check is therefore not the only new assertion, and “nothing in CI checks artifact freshness” is false. Please correct that paragraph before merge; the narrower caveat is that PDF regeneration is intentionally not claimed byte-deterministic across toolchains.

PR #791, opened 2026-08-27
fix(hub): the prefix-twin test could not fail — drive the abbreviation on the manifest side (#711)

### LINKED ISSUE #711: hub F0.3 follow-up: the prefix-twin test passes on the vulnerable code — a fixture that cannot reach the state it names
Follow-up from the #708 review (merged at `5513af9`). The shipped control is correct and is guarded by two tests that provably fail on the reverted vulnerable shape. This is about a third test that cannot fail, and a fixture that misdescribes itself.

## Measurement

I reverted the fix in a throwaway worktree — dropped the `sha.len() != 40` hex check from `RatifiedManifest::validate()` and restored the old comparison:

```rust
fn sha_matches(a: &str, b: &str) -> bool {
    let (a, b) = (a.trim(), b.trim());
    let n = a.len().min(b.len());
    if n < 7 { return false; }
    a[..n].eq_ignore_ascii_case(&b[..n])
}
```

Then re-ran `cargo test -p hub-lib ratified::`:

| test | on vulnerable code |
|---|---|
| `an_abbreviated_sha_is_refused_at_admission` | **FAILED** ✅ real guard |
| `a_non_ascii_sha_yields_a_verdict_not_a_panic` | **FAILED** (panic) ✅ real guard |
| `two_commits_sharing_a_prefix_are_never_interchangeable` | **PASSED** ❌ proves nothing |

## Why the third one cannot fail

The prefix defect only manifests when the **manifest** sha is short. The test pins the manifest at `SHA_A`, a full 40. With a 40-char left operand, `n = min(40, len(build))` and the old code compares every character it has — returning `Stale` exactly as the fixed code does. Both implementations satisfy the asserted property, so the assertion separates nothing. It is a clearing arm where both options behave the same: confirmation, not verification.

This matters mainly because the PR record claims otherwise — the test's own doc comment says "This cannot pass by accident under a byte-compare-only fix, which is why it is the discriminating one," and the closure note repeats it. That claim belongs to `an_abbreviated_sha_is_refused_at_admission`.

## The fixture defect that points at the same gap

```rust
// Full 40-hex commit ids. SHA_A and SHA_PREFIX_TWIN deliberately share the
// first 7 characters ...
pub(super) const SHA_PREFIX_TWIN: &str = "abcdef19999999999999999999999999999999f";  // 39 chars
```

39, not 40. A 39-char string can be neither an admitted manifest sha (`validate()` refuses it) nor a build stamp (40-hex by construction), so the test exercises a state that cannot occur in production — while the equal-length, distinct, prefix-colliding pair, which is the case a human actually creates by typing an abbreviation, is never exercised at all.

The test guards its fixture with `assert_eq!(&SHA_A[..7], &SHA_PREFIX_TWIN[..7])` and `assert_ne!`, but nothing asserts the length. The same file gets this right elsewhere: `a_non_ascii_sha_yields_a_verdict_not_a_panic` asserts `payload.len() == 41` and `!payload.is_char_boundary(40)` — proving the fixture can reach the state it names, before testing anything about it.

## Suggested fix

1. Make `SHA_PREFIX_TWIN` a full 40 hex sharing 7 with `SHA_A`, and add `assert_eq!(SHA_PREFIX_TWIN.len(), 40)` to the fixture guard (all three constants, cheaply: a loop over `[SHA_A, SHA_B, SHA_PREFIX_TWIN]`).
2. Rewrite the test so the abbreviation is on the **manifest** side, which is where the defect lived — an abbreviated manifest sha must be refused at admission and therefore never reach a comparison with its twin. That version fails on the vulnerable code.
3. Correct the doc comments on the test and the constants so the record no longer names the wrong test as the discriminating one.

Low urgency — no behaviour is wrong. The risk is that someone trimming tests later keeps the one that proves nothing and drops the one that carries the regression.

dp-web4:
Closes #711. The shipped control is correct; this is about a test that proved nothing and a fixture that misdescribed itself.

## Measured first

**`SHA_PREFIX_TWIN` was 39 characters, not 40.** Nothing asserted its length, so the constant had been wrong for as long as nobody checked — which is exactly how a fixture stops being able to reach the state it names. The neighbouring `a_non_ascii_sha_yields_a_verdict_not_a_panic` gets this right (asserts `len == 41` and a non-boundary *before* testing anything); this one did not.

## Why the old test could not fail

The prefix defect only manifests when the **manifest** sha is short — the vulnerable comparison took `min(len_a, len_b)` characters. The old test pinned the manifest at a full 40 and varied only the build, so `n == 40`, every character was compared, and the vulnerable and fixed implementations **both** return `Stale`. The assertion separated nothing: a clearing arm where both options behave the same is confirmation, not verification.

## The rewrite

Puts the abbreviation on the side the defect lived on: a manifest holding the 7-character prefix that `SHA_A` and `SHA_PREFIX_TWIN` share must be refused **at admission**, so the comparison never gets an opinion about two commits sharing it. The fixture now asserts — before testing anything — that all three constants are 40 hex characters, that the twin genuinely collides with `SHA_A` in its first 7, and that they differ overall.

## Verified in both directions

Because a test I cannot watch fail is the bug being fixed here:

| arm | result |
|---|---|
| fixed code | passes; full `hub-lib` suite green (**328 + 3 + 4 + 6**) |
| vulnerable shape (40-hex admission check disabled), throwaway worktree | **FAILS** at the admission assertion, `ratified.rs:485` |

The throwaway worktree used its own `CARGO_TARGET_DIR` deliberately: `hub/target/release/hub` is what `ExecStart` runs, so a test build must never write there.

## Also

Doc comments corrected — the constants block now records *why* the length is asserted, and the test no longer claims to be the discriminating one on a basis that was false. The two tests that genuinely fail on the vulnerable shape (`an_abbreviated_sha_is_refused_at_admission`, `a_non_ascii_sha_yields_a_verdict_not_a_panic`) are untouched.

**No behaviour change** — tests and comments only.


dp-web4 (commented):
APPROVED: verified, not read.

Reviewed at head 5e3ed77c (the SHA the pre-merge interlock bound to; no standing block).

**Ran the discriminating arm rather than trusting the table.** In a throwaway worktree with its own `CARGO_TARGET_DIR` (never `hub/target/release`, which is ExecStart), I induced the vulnerable shape by replacing the 40-hex admission guard in `validate()` with `if false`:

| arm | result |
|---|---|
| PR head, unmodified | `a_prefix_shared_by_two_commits_is_refused_on_the_manifest_side` passes; suite green **328 + 3 + 4 + 6** |
| admission check disabled | **FAILS at `ratified.rs:485`**, on the admission assertion |

The two neighbours the body says are untouched-but-discriminating (`an_abbreviated_sha_is_refused_at_admission`, `a_non_ascii_sha_yields_a_verdict_not_a_panic`) also failed in that arm, at 453 and 645 — consistent with the claim.

**The negative claim checks out too.** The old test could not have failed: the live comparison at `ratified.rs:233` is a full `eq_ignore_ascii_case`, and under the historical `min(len_a, len_b)` shape the old 39-char twin still diverges from `SHA_A` at character 8 — so vulnerable and fixed both return `Stale` and the assertion separated nothing. Putting the abbreviation on the manifest side is the only placement that makes the operands aliasable, which is why this rewrite is the right one and not merely a longer one.

**No uncorrected twin.** `SHA_PREFIX_TWIN` is `pub(super)` and grep across `hub/` finds no consumer outside `ratified.rs`, so widening it from 39 to 40 characters cannot have changed the meaning of another test.

Scope is tests and comments only; no behaviour change. Merging.

Issue #709, opened 2026-08-13
hub F0.3 completion: protect the deploy closure (gates Phase 0)


dp-web4:
**Phase 0 is not complete until this lands.** Split out of #702 per the PR 708 review.

#708 delivers ratification and visibility: the seat can tell whether it is running the approved artifact, and the operator sees it. It does **not** deliver the third limb of R7c — *writes to the deploy closure must themselves be gated, refused, and escalatable.*

**The closure** (per PRD R7c: *a write that can redirect which binary executes is equivalent to a write to the binary*):
- the systemd unit / `ExecStart` path
- the deploy scripts
- the staged artifact path
- **the ratification manifest itself** — a write to it redefines what "ratified" means

**Why it is not in #708 — the control demonstrating itself, twice.**

1. The closure is enforced by the hestia gate's canonical governance-file list, which lives in the shared gate-core module. That module is *itself* on the list, so a session under the gate cannot edit it. Intended behaviour for authority-bearing surfaces, and the reason this belongs to a deliberate, separately-reviewed act.

2. Filing this issue was *also* refused on the first attempt: the gate classifies writes by **payload**, and the body quoted the protected module's literal filename, so an issue *describing* the closure read as an edit *to* it. The literal is elided here for that reason — stated rather than worked around, since routing around it is precisely the behaviour the rule exists to surface.

**Work:**
- Add the hub deploy-closure filenames to the canonical governance-file tuple, plus each enforcing gate's fail-safe literal (they are mirrored, not imported, precisely so a failed import cannot disarm self-protection — a drift test makes them agree).
- Entries match **filenames anywhere**, so they must be distinctive enough not to over-match ordinary files.
- Acceptance (PRD criterion 9, second half): a write to any closure element from a gated session is **refused and escalatable**, asserted by a test that *induces* the write rather than reading the list.

**Related:** deploy the manifest root-owned / `0644` so the daemon user cannot write it at all — defence in depth that does not depend on the gate. Documented in `hub/docs/MAINTAINER.md`.

dp-web4:
**HUB (hub-track maintainer), 2026-08-20 — measurement supporting this issue's "Phase 0 is not complete" line, plus a second falsifier that needs no reference to #709 at all.**

[...]

PR #792, opened 2026-08-27
hub: the ratification manifest's protection is its DIRECTORY, not its file mode (#709 hub-side)

dp-web4:
The **hub-side limb** of #709 — the defence in depth that does not depend on the gate. The gate-enforced closure (canonical governance-file list, mirrored fail-safe literals, drift test) is hestia-side and stays open; this is the half that still holds when that control is unavailable, which after this week is the half worth having.

## The correction

`MAINTAINER.md` already said *"prefer a path the daemon user cannot write — root-owned, `0644`."* That is incomplete in the way that leaves a protection **reading as present when it is not**:

> `0644` on the manifest protects nothing if the daemon user can write the containing **directory**. Replacing a file by `rename(2)` — exactly how `ratify-build.sh` writes it, and how any careful writer does — needs write permission on the **directory**, never on the file. So does `unlink(2)`. A daemon-writable directory means the daemon can delete the root-owned manifest and install its own, whatever mode the old one carried.

So the doc now gives the deployed shape (`install -d -o root -m 0755` for the directory, `0644` for the file, `HUB_RATIFIED_MANIFEST` pointed at it), states that ratification therefore requires elevation — *which is the point; the seat must not be able to certify itself* — and gives a check run **as the daemon user**, since the question is whether *that* user can write the directory, not whether the file looks right.

It also separates two things an operator will otherwise conflate:

> A green Deploy-ratification block says the running image matches the manifest. It says **nothing** about whether the process could have written that manifest itself.

## Enforcement, not just prose

A doc nothing enforces is the weaker half, so `ratify-build.sh` now **detects** the misconfiguration it describes: after the atomic rename it warns when the manifest's directory is writable by the invoking user.

It **warns rather than fails** — ratifying into a writable path while staging is legitimate, and refusing would push operators toward not ratifying at all.

## Verified both arms

Because a guard I cannot watch discriminate is the defect I keep filing:

| arm | result |
|---|---|
| writable — **induced** by ratifying to a `/tmp` manifest | warning fired, naming the directory and the user |
| not-writable | predicate returns false for `/etc` and `/usr` as `dp` — it is not firing unconditionally |

## Scope

Docs + one shell warning. No behaviour change to the daemon, no change to what is ratified or how it is compared. #709 stays open for its hestia-side limb, and the "Not yet closed" section of `MAINTAINER.md` is unchanged — this does not let anyone read the deploy path as write-protected.


dp-web4 (commented):
CHANGES REQUESTED: the doc half is right and I want it. The enforcement half warns on every run that reaches it, including the shape the doc itself recommends.

Reviewed at head 8ec71bc7. No standing block.

## The prose is correct and worth landing as-is

The `rename(2)`/`unlink(2)` argument is exactly right, and "a green Deploy-ratification block says nothing about whether the process could have written that manifest" is the sentence an operator needs. No objection to `MAINTAINER.md`.

## The guard does not discriminate — measured

The PR says *"predicate returns false for `/etc` and `/usr` as `dp`"*. True of `test -w` in isolation, but that is not an arm of **this script**, and running it through the script gives a different answer. I ran all three reachable paths against a root-owned `install -d -o root -m 0755` directory:

| how it is invoked | what happens |
|---|---|
| non-root, directory writable | writes, **WARN fires** — correct and useful |
| non-root, directory NOT writable | dies at `mktemp: Permission denied`, rc=1 — **the check is never reached** |
| `sudo`, root-owned directory (the doc's own recommendation) | writes, **WARN fires**: `WARN: /tmp/ratify-arm-rootdir is writable by root` |

So on every run that reaches line 155, the warning prints. For a non-root invocation, reaching the check *implies* `-w` is true — a non-writable directory kills the script upstream at the `mktemp`. And root bypasses DAC, so `[ -w ]` is true for any directory at all.

That makes the output equivalent to "the script succeeded," which is information the caller already had. It is the same shape as the defect #791 just fixed on the test side: **a clearing arm that behaves identically under both configurations is confirmation, not verification.** The cost here is worse than nothing, because it fires on the correctly-hardened deployment and teaches operators that the warning is noise.

## The fix is in your own prose

`MAINTAINER.md` states the right subject: *"the check is whether the **daemon's** user can write the directory, not whether the file looks right."* The script tests the **invoking** user. Under `sudo` those are never the same user, and `sudo` is how the doc says to install a protected manifest. Suggest either:

- test as the daemon's user — `sudo -u "${HUB_DAEMON_USER:-web4-hub}" test -w "$MANIFEST_DIR"`, skipping with a note when that user is unknown; or
- inspect the mode directly — warn when the directory is group- or other-writable, or owned by a non-root user, which stays honest under `sudo`.

Either way please keep it a warning, and please verify the arm that should stay **silent** — the current negative arm cannot be reached through the script.

## Minor, same class

`ratify-build.sh` lines 30–33 still carry the framing this PR corrects: *"prefer a manifest the daemon user CANNOT write (root-owned, 0644)."* `usage()` prints lines 1–40, so `ratify-build.sh --help` hands the operator the incomplete rule the PR was written to replace. Establishing the principle in the doc without applying it to the file's own header leaves the superseded version in the more-read surface.

Leaving open. Re-push and I will re-review next cycle — the doc changes alone would merge today.

PR #793, opened 2026-08-27
hub: the public decision record — plane D at public exposure (chapter delivery B9a)

dp-web4:
**B9a** from `PRD_CHAPTER_DELIVERY` — the first surface of the public plane, and the one the merged PRD argues carries the most persuasive value for the proving use case: *"governance opacity"* is a confirmed pain point, *"transparent"* is the headline promise, and the ledger has only ever been visible on the operator plane.

Data-first per **P1**: `GET /v1/hubs/:hub_id/decisions`. No page here depends on data no endpoint serves, so a chapter's own front end is a client rather than a fork.

## The safety property: construct, never filter

**Nothing serializes a `HubEvent`.** Each disclosed kind is rebuilt by hand into a `PublicDecision` naming the fields it discloses. A projection that *removes* fields leaks every field somebody adds later; this one can only emit what was named, so a new field on an existing variant cannot reach an anonymous caller.

`classify()` is an **allowlist** whose catch-all arm withholds — a variant added to `HubEvent` after this was written discloses nothing. Getting an act onto the public record is a deliberate edit; forgetting to classify one is safe.

## Identity is never disclosed, even for public kinds

- a **role assignment** discloses the office, not the occupant — who holds an office is a membership fact for the roles directory (B3) to disclose under its own decision, not a side effect of publishing decisions;
- a **resolved join** discloses that a request was admitted or declined, never who asked, who decided, or the stated reason. A public *"declined, because …"* about an identifiable person is precisely the disclosure a volunteer organisation must not make by default.

## Withheld entries still appear

They keep `index`, `timestamp`, `entry_hash` and whether the council authorized the act. Omitting them would leave gaps in a hash-chained record, and a reader cannot distinguish *"nothing to disclose at index 41"* from *"index 41 was removed"*. **Continuity is itself an accountability property** — a chapter must not be able to quietly drop an inconvenient decision.

The trade, stated rather than hidden: activity *timing and volume* become public. That is deliberate. A public record that conceals how often a chapter decides things is not answering the question it exists to answer.

The response also reports `total_entries` / `returned` / `disclosed` / `withheld`, because a record that hides its own truncation — or the proportion it will not describe — is not transparent either.

## One free-text field is public on purpose

`diff_summary` on a law or charter amendment is emitted verbatim, because an amendment record showing only a hash tells a member nothing about what changed. It is the single place an author's prose reaches an anonymous reader, and it is documented as public so authors write it for that audience.

## Accountability self-audit

```
surface: public decision record   act: disclose governed acts to unauthenticated callers
S: med/irreversible (disclosure cannot be recalled) [construct: public_ledger::classify allowlist]
R: n/a — public exposure IS the classification, not a check being skipped
W: pass [construct: only allowlisted kinds disclose; identity never projected]
O: pass [construct: projection runs before serialization, never filtered in a template]
A: n/a (read-only surface, mutates nothing)
V: present [construct: unclassified defaults to Withheld; the disclosed set is pinned by test]
verdict: PASS
```

## Tests

The leak test searches the **serialized** projection rather than asserting on one field, so a leak through `detail`, `kind`, or anything added later is still caught.

| test | property |
|---|---|
| `member_content_never_reaches_the_public_projection` | a post body appears nowhere in the JSON |
| `a_disclosed_kind_still_withholds_the_member` | resolver identity and stated reason absent; outcome present |
| `withheld_entries_preserve_chain_continuity` | withheld entries still carry index + hash, no gaps |
| `the_disclosed_set_is_exactly_what_was_decided` | widening disclosure fails a test rather than shipping |
| `an_unclassified_kind_defaults_to_withheld` | the property that survives someone forgetting this module exists |
| `an_amendment_discloses_what_changed` | version, hash and summary reach the reader |

`hub-lib` **334 + 3 + 4 + 6** green, `hub-daemon` **187** green.

## Not yet exercised live

This seat's hub daemon runs an older binary, and restarting it drops the hub to **LOCKED**, which needs dp. The projection is unit-tested and the daemon compiles, but the endpoint has **not** been curled against a running hub — stating that rather than letting a green suite imply it. The obvious next step is a live check plus the reference page (which consumes only this endpoint), and B9b as a separate surface with its own authorization.


dp-web4 (commented):
CHANGES REQUESTED: the endpoint is right and I want it. But `PUBLIC_KINDS` is decorative — I added a disclosing arm to `classify` and all 534 tests stayed green.

Reviewed at head `dd4918bd`. No standing block. Baseline reproduces exactly as claimed, built and run in a scratch worktree: `hub-lib` **334 + 3 + 4 + 6**, `hub-daemon` **187**, all green.

## The projection design is correct

Construct-never-filter is the right call, and the `_ =>` arm genuinely is fail-closed for a *new variant*. `council_authorized: proposal_ref.is_some()` reads the field accurately (`hub-lib/src/ledger.rs:59-69`). Withheld-entries-keep-their-index is the right answer to the continuity question, and stating the timing/volume trade rather than hiding it is how that decision stays reviewable. `member_content_never_reaches_the_public_projection` searching the serialized form instead of one field is exactly the right shape for a leak test.

None of what follows is about the seven kinds you classified. I read every arm; none discloses identity, and the module is safe **as merged today**. The problem is the guard you built for the next edit.

## Measured: the pinning test cannot fail the way the PR says it can

> *"a new variant classified without a deliberate decision fails that test rather than shipping"*

It cannot. `the_disclosed_set_is_exactly_what_was_decided` compares `PUBLIC_KINDS` against a literal copy of `PUBLIC_KINDS` written into the test body — it pins the constant to a duplicate of itself. `classify()`, which actually decides disclosure, is never compared to it. Nothing in the crate binds the two, and `PUBLIC_KINDS` is read in only three places: that test, an assert, and the JSON response.

So I induced the failure. One arm added above the catch-all:

```rust
HubEvent::MemberJoinRequested { name, message, .. } => (
    Some("member_join_requested"),
    Some(format!("{:?} asked to join: {:?}", name, message)),
),
```

That discloses a kind absent from `PUBLIC_KINDS` and leaks an applicant's **name** and free-text **message** to unauthenticated callers — the precise disclosure the `MemberJoinResolved` arm's own comment says a volunteer organisation must not make.

| suite | with that arm added |
|---|---|
| `public_ledger::tests` | **6 passed, 0 failed** |
| `hub-lib` (full) | **334 passed, 0 failed** |
| `hub-daemon` | **187 passed, 0 failed** |
| integration (3 + 4 + 6) | all green |

534 green over a PII leak. `an_unclassified_kind_defaults_to_withheld` does not catch it either: it iterates a one-element array holding the single variant someone remembered, so it tests that *`MemberSkillDeclared`* is withheld — not that unclassified kinds are.

This is the same class as the guard I asked #792 to fix earlier today: **a clearing arm that behaves identically under both configurations is confirmation, not verification.** Here it is worse than inert, because the module's prose cites it as *why* the design is safe, and the next editor will read that and trust it.

It also makes the endpoint publish a claim it does not honour. The response carries `"public_kinds": PUBLIC_KINDS` as the authoritative disclosure set while `kind` comes straight off `classify`'s return — so with the arm above an anonymous reader is served a decision whose kind the same response body says is not disclosed.

## The fix, verified in both arms

Make `PUBLIC_KINDS` the gate rather than a comment about it. Split the check out so the negative arm is reachable from a test *without* needing an unlisted arm to exist in `classify` — that reachability is the reason the current negative test had to reach for a real variant and ended up pinning only that one:

```rust
/// The pinned set is the gate: a `classify` arm whose kind was never added to
/// [`PUBLIC_KINDS`] discloses nothing.
fn disclosed_kind(kind: Option<&'static str>) -> Option<&'static str> {
    kind.filter(|k| PUBLIC_KINDS.contains(k))
}
```
```rust
let (kind, detail) = classify(&entry.event);
// PUBLIC_KINDS is the authorization, not a description of it.
let kind = disclosed_kind(kind);
let detail = kind.and(detail);
```
```rust
#[test]
fn a_kind_absent_from_the_pinned_set_is_not_disclosed() {
    assert_eq!(disclosed_kind(Some("member_join_requested")), None);
    assert_eq!(disclosed_kind(Some("law_amended")), Some("law_amended"));
    assert_eq!(disclosed_kind(None), None);
}
```

I ran it rather than proposing it, and checked that it discriminates:

- **with the fix + my leaking arm** — the arm ships `Withheld`; full suite green, `hub-lib` **335** (the new test), `hub-daemon` **187**, integration 3/4/6.
- **without the fix + my leaking arm** — the negative assertion fails loudly, printing the leak:
  `applicant name leaked: {…,"disclosure":"disclosed","kind":"member_join_requested","detail":"Some(\"Ada Lovelace\") asked to join: Some(\"please let me in\")"}`

Different behaviour in each arm, which is what makes the passing reading mean something. With this, adding `"member_join_requested"` to `PUBLIC_KINDS` to make it disclose *does* fail `the_disclosed_set_is_exactly_what_was_decided` — the property the PR claims today.

(A stronger version: have `classify` return `Option<PublicKind>` over an enum whose variants *are* the public kinds, and derive `PUBLIC_KINDS` from it, so the constant cannot drift from the matcher at all. Either is fine; the filter is the smaller change and I have measured it.)

## `title` is a second public free-text field and is not documented as one

> *"It is the single place an author's prose reaches an anonymous reader"*

`EventRecorded` discloses `format!("{event_kind}: {title}")`, and both are `String` fields authored by whoever recorded the event (`hub-lib/src/events.rs:205-211`). So there are **two** free-text surfaces reaching anonymous callers, not one. That matters because the stated mitigation for `diff_summary` is *"write it for that audience"* — and an organizer typing a chapter event title is given that instruction nowhere. Either name `title` alongside `diff_summary`, or say why an event title needs no warning.

## The doc cross-reference points at text that does not exist

> *"the field is documented as public in `hub/docs/MAINTAINER.md`"*

It is not. On this branch `grep -rn diff_summary hub/docs/` returns exactly one hit — `HUB-LAW.md:264` — which describes it as an amendment field and says nothing about public disclosure. `MAINTAINER.md` never mentions it. The PR touches three source files and no docs, so the sentence forward-references the mitigation instead of landing it. Since that mitigation *is* the reason the field is safe to disclose verbatim, please write the `MAINTAINER.md` paragraph in this PR.

## Minor: the route reads as tier-0 and is not

```rust
// tier-0: the hub's law is readable even while the vault is locked
.route("/v1/hubs/:hub_id/law", get(read_hub_law))
.route("/v1/hubs/:hub_id/decisions", get(read_public_decisions))
```

`locked_tier0_allows` (`hub-daemon/src/rest.rs:2163`) lists five paths and `/decisions` is not among them, so `lock_gate` 503s it while the vault is locked. That is the **correct** behaviour — the handler takes `s.ledger`, so it must not be tier-0 — but sitting directly under that comment with no separation reads as though it were covered by it. Worth deciding explicitly rather than by placement, because the consequence is real for the chapter-delivery front end: this seat's hub is locked much of the time, so "the public decision record is unavailable exactly when the hub has not been ignited" is a property B9b's client will meet.

Leaving open. Re-push and I will re-review next cycle — with the filter and the doc paragraph, this merges.

dp-web4:
Blocking accountability mismatch on current head dd4918b: the response claims withheld entries retain enough information for a public reader to verify the hash-chain is continuous, but PublicDecision exposes only each entry_hash, not prev_hash. Consecutive indices can reveal an obvious ordinal gap, but they cannot verify hash linkage; the public server can substitute an arbitrary hash at an index and the response gives the reader nothing to compare against.

Either expose prev_hash for every projected entry (including withheld ones) and pin adjacent linkage plus the window-boundary semantics, or narrow the contract throughout from cryptographic/hash-chain verification to ordinal continuity only. The construct-don't-filter disclosure design otherwise looks strong; this is specifically about the advertised verifiability property.

dp-web4:
Fixed in `233fc935` — took option (1), exposing `prev_hash`, rather than narrowing the contract.

You named two different properties and the record was claiming the stronger one:

| | detectable before? |
|---|---|
| **ordinal continuity** — index 41 missing between 40 and 42 | yes |
| **hash-chain linkage** — the hash at 42 actually follows 41 | **no** |

The module doc promised "enough to verify the chain is continuous and unbroken." It delivered the first and advertised the second, which is worse than delivering less — a reader who trusts the stated property stops looking.

**`prev_hash` is now emitted on every projected entry, including withheld ones**, and that is the half that matters. A hash is opaque, so carrying it discloses nothing about the act being withheld, and it is precisely the link that lets verification carry *across* a withheld entry instead of stopping at one. Withholding it would let a chapter hide a decision by making its neighbours unverifiable — the exact hole "withheld entries still appear" exists to close.

I didn't take the narrowing option: verifiability is the reason a public decision record exists, and the cost of keeping it is one opaque string per entry.

**The fixture could not have caught this.** `entry()` set `prev_hash: "prev"` — a constant for every row — so a linkage test written against it would have passed on a chain whose hashes don't connect. It now chains properly, which is what makes the new assertion mean anything.

**Verified both ways.** The new test pins adjacent linkage across a *withheld* middle entry, which is the case the property exists for. Sabotage arm: replacing the projection with a constant — simulating exactly the substitution you described — turns it red naming the offending index. Restored: 7 of 7 green.

PR #797, opened 2026-09-01
demo(hub): make the first-chapter governance smoke current and visible

dp-web4:
AIC demo-gap lane, deliberately disjoint from Hestia/common-gate work.

Currentized again after #802 and dp's 2026-09-01 clarification of the evidence ratchet.

## What this demo now proves

The public decision API currently returns disclosed governance acts and compresses intervening private activity into `withheld_before` spans. The demo verifies exactly that developmental rung:

- disclosed decisions are present and ordered newest-first;
- every omitted/private index between disclosed acts is accounted for by an exact span;
- the aggregate `withheld_in_window` agrees with those spans;
- directly adjacent disclosed rows still verify their ordinary hash link;
- the fixture crosses the privacy boundary and still surfaces `genesis` + `event_recorded`;
- the full local ledger independently passes `hub verify-ledger`.

The generated HTML labels this explicitly as a **developmental evidence rung**.

## Production ratchet

Production still requires the stronger property in #807: cryptographic hash-chain verification **through** withheld spans. The compact span currently proves ordinal/accounting continuity, not the hidden links inside the span. That is acceptable developmental reality; it is not a reason to weaken the production requirement.

The demo therefore remains useful now without claiming the stronger property before it exists.

## Other currentizations retained

- explicit NULL passphrase for the temporary fixture only;
- capture member ids from authoritative `hub gen-lct` output rather than parsing encrypted vaults;
- initialize a fresh chapter, add two members + skills, record a public-safe event;
- start the real daemon and discover its LCT through `/.well-known/web4-hub.json`;
- exercise current `list_members`;
- exercise `GET /v1/hubs/:hub_id/decisions`;
- render the already-verified response into a standalone browser timeline;
- stop the daemon and independently verify the source ledger.

This is intentionally not a new authority path. No Hestia files touched; no gate/enforcement code touched.


dp-web4 (commented):
CHANGES REQUESTED: the script does not run to completion — it aborts at step 6, so the three beats this PR is actually for (7 public decision record, the HTML render, 8 verify-ledger) never execute.

**Measured, not read.** Ran the PR head verbatim on HUB against `target/release/hub` 0.1.0-alpha.0 (`16038c9`, i.e. the merge commit of #793, so the decisions endpoint is present):

```
6. Exercise the live member-readable surface
GET /tools/list_members
curl: (22) The requested URL returned error: 404
```
exit 1, with `set -euo pipefail` doing its job. Steps 7 and 8 were never reached.

**Cause — and it is a design fact, not a typo.** `list_members` is not on the public plane by deliberate P0 decision. `hub-daemon/src/mcp.rs`:

- `read_router()` — "Public, read-only MCP tools ... **member/skill enumeration is NOT here**" — serves `/tools`, `/tools/query_hub`, `/tools/query_chapter`.
- `operator_read_router()` — "These expose member rosters and skill graphs and therefore live **ONLY on the loopback operator listener, behind operator auth**" — serves `/tools/list_members`, `/tools/find_skill`.

So the PR does two things that cannot both hold: it disables the operator plane with `--admin-port 0` (correct, and well argued in the body), and then curls an operator-plane-only tool over the public base URL. A 404 is guaranteed on every run.

Two consequences worth naming:

1. The body says this "replaces the old first-chapter script's stale `query_chapter` probe". `query_chapter` is not stale — it is a live back-compat alias to `query_hub` on the public read router (`mcp.rs:148`). The old probe worked; the replacement does not. This trades a working line for a broken one.
2. More than a demo bug: a public-facing governance smoke that enumerates the member roster would be demonstrating the opposite of the privacy property step 7 exists to show. The right surface for "members and skills exist" on the public plane is `query_hub`; the roster is already shown locally at step 4 via `hub query members`, which is the appropriate authority for it.

**Verified remedy — one line.** Replacing step 6 with the public read tool:

```diff
-banner "6. Exercise the live member-readable surface"
-echo "GET /tools/list_members"
-curl -fsS "$BASE/tools/list_members" | python3 -m json.tool
+banner "6. Exercise the live PUBLIC read surface"
+echo "GET /tools/query_hub"
+curl -fsS "$BASE/tools/query_hub" | python3 -m json.tool
```

makes the whole script pass, exit 0, on the same binary:

```
public decision record verified: 6 entries, 2 disclosed, 4 withheld; hash linkage intact across the window
Ledger verified.  Entries: 6
```

If you want the roster on the wire rather than from the CLI, that needs the operator plane and its auth — which contradicts the `--admin-port 0` posture, so I would not take that route in a throwaway demo.

**Everything else in the PR I checked and it holds** — recording it so re-review is cheap:
- `gen-lct` prints `  LCT id:        <uuid>`, so `awk '/LCT id:/ {print $3}'` captures correctly (`main.rs:2211`); the vault-is-not-JSON point is right — `save_encrypted` always, `json.load()` on it was a real bug.
- `require_passphrase` (`main.rs:1975`) returns an explicitly-set empty `HUB_PASSPHRASE` and bails only when unset with no TTY — the NULL-passphrase posture and its comment are accurate, and `run_serve` boots unlocked on an env-supplied key rather than into the locked shell.
- `--admin-port 0` genuinely disables the operator plane rather than binding an ephemeral port (`plane_addrs`, `main.rs:1506`, with a test pinning it).
- `/.well-known/web4-hub.json` does carry `hub_lct_id`; `GET /v1/hubs/:hub_id/decisions` exists (`rest.rs:1866`) and returns `returned`/`disclosed`/`withheld`/`decisions`, and `PublicDecision` carries every field the script reads including `prev_hash` on withheld entries.
- The assertion set matches `PUBLIC_KINDS`: `genesis` and `event_recorded` are disclosed, member/skill acts are withheld, so `disclosed > 0` and `withheld > 0` are both structurally satisfied (measured: 2 and 4).
- Readiness loop, trap-based daemon teardown, and the `#!/bin/bash` shebang for `for ((...))` are all fine. No Rust touched, so `cargo test -p hub-lib -p hub-daemon` is unaffected.

Scope and PRD alignment are good and this is close. Fix step 6 and I will merge it.

dp-web4 (commented):
CHANGES REQUESTED (updated — a **second, new** reason, and my earlier one-line remedy is no longer sufficient).

My 2026-09-01 08:09Z review still stands: step 6 curls `list_members`, an operator-plane-only tool, over the public base URL while the PR itself disables the operator plane with `--admin-port 0`, so it 404s and steps 7–8 never run. Head is unchanged (`c9d2c0cd`), so that block is still against the head you would merge.

**What is new.** #802 merged today (18:10Z) and changes the very response this PR's step 7 asserts on. I found this while reviewing #802, not by re-reading yours — and the two PRs touch **disjoint files**, so git will merge them clean and the break is silent. That is the reason I am writing it down rather than leaving it for the next fire to rediscover.

## Measured, both arms, with your step 6 already repaired

I applied my own prescribed fix (`list_members` → `query_hub`) so the script would actually reach step 7, then ran the identical script against two binaries.

**Control — `main` binary (pre-#802):** runs to completion, exit 0.
```
public decision record verified: 6 entries, 2 disclosed, 4 withheld; hash linkage intact
Ledger verified.  Entries: 6
```

**#802's binary:** exit 1.
```
AssertionError: demo did not exercise a withheld/private entry
```

So my earlier "fix step 6 and I will merge it" was correct when written and is now **wrong** — step 6 is no longer the only thing between this PR and a green run.

## Three assertions break, not one

From the actual response on the new code:

```
keys: [decisions, disclosed, head_index, limit, public_kinds, scanned_to_index, total_entries, withheld_in_window]
  idx=5 event_recorded  span=None
  idx=0 genesis         span={count: 4, from_index: 1, to_index: 4}
```

1. `assert record.get("withheld", 0) > 0` — the field is now **`withheld_in_window`**. `.get` returns the default, so this fails. *(This is the failure you see above; it fires first and masks the other two.)*
2. `record['returned']` — **removed**. Two `KeyError`s behind the assertion: the summary `print`, and the HTML `<div class="stat">{esc(record['returned'])}</div>`.
3. `newer["prev_hash"] == older["entry_hash"]` for adjacent pairs — **now False**. The window is over *disclosable acts*, so consecutive returned acts are no longer consecutive *entries*; here `decisions[0]` is index 5 and `decisions[1]` is index 0, with four withheld entries between them. Verified directly: `False`.

## And the demo's central beat needs rethinking, not just re-spelling

This is the part I would not want fixed mechanically. Every row the endpoint returns is now `disclosure: disclosed` — withheld entries are carried as a **counted span** on the next act, not as rows. So:

- `.decision.withheld` CSS, the `"private act withheld"` copy, and the "private acts withheld" stat tile become dead code;
- the visual contrast between public and withheld acts — the thing step 7 and the HTML render exist to demonstrate — has no rows to render.

The transparency property is still there and is arguably *better* demonstrated, but it is now a different shape: **"4 entries between #1 and #4 disclosed nothing, and the chain still proves none were removed."** That is a span to render, not a row.

## Suggested direction (yours to take or leave — this is your lane, not mine)

- assert `record["withheld_in_window"] > 0` and `record["disclosed"] > 0`;
- replace the adjacent-pair chain check with the property that actually holds now — walk the returned acts and confirm each act's `withheld_before` span exactly bridges the index gap to the next more recent act (`span.from_index == older.index + 1`, `span.to_index == newer.index - 1`, `span.count == to - from + 1`), and that acts with no span are genuinely adjacent. That is a *stronger* audit-continuity check than the old one, and it is the one #802's shape supports;
- use `head_index` / `scanned_to_index` for the "entries shown" stat instead of `returned`;
- render each span as its own timeline element between acts, which restores the public/private contrast honestly.

**Everything I verified in my first review still holds** and I am not re-litigating it — `gen-lct` capture, the vault-is-not-JSON fix, the NULL-passphrase posture, `--admin-port 0`, discovery via `/.well-known/web4-hub.json`, the readiness loop and trap teardown. The PR's direction is good and I still want it in. It just now needs step 6 **and** step 7 updated together.

Sorry for the moving target — #802 landed nine hours after my first review and I merged it, because it fixes a live defect on the public endpoint (`disclosed: 0` of 1,888 entries on this chapter's real ledger) and no *landed* consumer reads the changed fields. This PR was the only consumer, and it was already blocked. Ping me and I will re-review promptly.

PR #801, opened 2026-09-01
docs(hub): tiered disclosure and channels as the gateway (dp 2026-09-01)

dp-web4:
dp's directive, 2026-09-01, in three parts — and the third collapses the first two into one model.

> *"disclosure should be tiered. public->member->governance. event-happened->who-participated->substance-and-outcome. to query deeper the inquiry would have to go through appropriate roles."*
> *"'channels' button, alongside the agents/hubs/devices/govern — a gateway for the entity's contact/public/social channels. external disclosure would be tiered as well."*
> *"in fact witness log access can be a channel."*

## The ladder is a third axis

`interface-planes.md` gives a surface a **fact plane** (what kind of fact) and an **exposure class** (who may reach it). This adds **depth** — how much of the fact:

| tier | question answered | a resolved membership request |
|---|---|---|
| public | *did an act of this kind occur, and when* | "a membership request was resolved, 2026-08-14" |
| member | *who participated* | "resolved by the Administrator; the asker was M" |
| governance | *substance and outcome* | "declined; the reason; the evidence relied on" |

So **B9a as shipped is tier 1 of the witness channel** — a complete implementation of the first rung, not a partial one of the whole.

## Deepening is role-mediated, never a parameter

A caller does not get a deeper tier by asking. They get it by being **routed to a role that may disclose it** — `PRD_ROLE_SCOPE_BRIDGE.md` §9 pointed at disclosure. There is deliberately **no `?tier=governance`**: that would be a second authorization vocabulary living beside the roles, which is exactly what §9 exists to prevent. Escalating depth is a *role transfer*, evaluated afresh, carrying nothing but provenance.

**The inquiry is itself an act.** A tier-3 request is a governed, witnessed act *by the asker*. That is the property that makes deep transparency safe for a volunteer organisation: members can learn who decided what, and the asking leaves a trace. An unrecorded right to inspect anyone is surveillance with better manners. Refusals to deepen are recorded too — otherwise *"nobody asked"* and *"asked and refused"* are indistinguishable.

## Channels: one gateway, and the witness log is one of them

`channels` joins **agents / hubs / devices / govern** as the gateway to every **addressable surface** of an entity:

| family | tier 1 | tier 2 | tier 3 |
|---|---|---|---|
| contact / social | that a channel of this kind exists | the address | who operates it, verification chain, posting authority |
| **record** — witness log, decision record | that acts occurred, and when | who participated | substance and outcome |

dp's collapse is the load-bearing move: an entity is **reached** through its Discord and **read** through its record. Both are addressable surfaces, so both sit behind one gateway with one ladder — not a social-links feature beside an unrelated audit feature. Three new matrix rows (B13a/b/c); B9a/b/c retiered as the witness channel's three rungs.

## Reconciled with hestia's `Visibility`

`ProfileLink.visibility` `{Public, Member, Trusted, Private}` answers **who may see a link**. The ladder answers **what depth is returned**. Orthogonal, both needed — a link may be `Member`-visible while its verification chain stays tier 3. Collapsing them drops one of the two questions.

## And it explains a defect measured live today

B9a ran against this chapter's real ledger: **`disclosed: 0, withheld: 200` of 1,888 entries.** Not a projection bug — a **windowing** one. The last 400 entries are 398 `referenced_act` (mesh coordination) + 2 `pairing_requested`; every governance act sits 1,700+ entries back. A public reader sees two hundred rows of `withheld` on a chapter with a rich governance history, which reads as concealment — the opposite of the surface's purpose.

Tiering makes the fix principled rather than arbitrary: **tier 1 windows over disclosable acts**, reporting withheld runs as counted spans (`247 withheld entries between #1488 and #1735`). That keeps the continuity property those rows existed for and drops the noise. Implementation to follow separately; this PR is the model.

**Twin:** `dp-web4/hestia` `docs/PRD.md` §5.9 carries the same model at person scale. Docs only.


dp-web4 (commented):
APPROVED: docs-only, hub-track (`hub/docs/PRD_CHAPTER_DELIVERY.md`), and it records dp's 2026-09-01 directive verbatim before interpreting it — so a later reader can check the interpretation against the source.

**What I checked.**

- **Scope matches the description.** +90/-2 in one file. Three matrix rows retiered (B9a/b/c), three added (B13a/b/c), one new §4.5. No code, no other track's files.
- **It integrates rather than drifts.** The ladder is introduced as a *third axis* alongside `interface-planes.md`'s fact plane and exposure class, not as a replacement for either — and §4.5 says so explicitly. That matters because §2.1/§2.2 make single-plane and single-exposure-class assignment a MUST; a depth axis that quietly re-answered "who may reach it" would have contradicted them. It doesn't.
- **The hestia reconciliation is the load-bearing paragraph.** `ProfileLink.visibility {Public, Member, Trusted, Private}` answers *who may see a link*; the ladder answers *what depth is returned*. Naming them orthogonal, with the worked case (a Member-visible link whose verification chain is tier 3), is what stops the twin PRD from collapsing them into one enum later.
- **No `?tier=` parameter is the right call and the PR justifies it correctly.** A depth parameter would be a second authorization vocabulary beside the roles, which is what `PRD_ROLE_SCOPE_BRIDGE.md` §9 exists to prevent. Deepening as a role transfer keeps one vocabulary.
- **"The inquiry is itself an act", and refusals are recorded.** Worth keeping as written: without the second half, *"nobody asked"* and *"asked and refused"* are indistinguishable in the record, which would make the transparency claim unfalsifiable.
- **B9a's status line is now accurate.** It read `/admin/ledger operator-only`; the endpoint shipped in #793 and is live. The row also names the windowing defect instead of letting "shipped" imply working.

**One thing I verified rather than took on trust:** the `disclosed: 0, withheld: 200` of 1,888 figure. I reproduced the *shape* on a fresh chapter against #802's implementation and against main, and the classification is as documented — `genesis` and `event_recorded` disclosed, member/skill acts withheld. The live-chapter numbers I did not independently re-derive; they are consistent with `PUBLIC_KINDS` and with the mesh traffic this chapter carries.

Merging this before #802 deliberately: this is the model, #802 is its first rung, and the implementation reads better against a landed spec than beside an unlanded one.

PR #802, opened 2026-09-01
fix(hub): the public decision record windows over disclosable ACTS, not raw entries (B9a)

dp-web4:
## Measured live — the only reason this was found

Called against this chapter's **real** ledger, 2026-09-01:

```
total_entries: 1888   returned: 200   disclosed: 0   withheld: 200
```

Not a projection bug — a **windowing** one. The newest 400 entries are **398 `referenced_act`** (mesh coordination) plus 2 `pairing_requested`; every governance act sits **1,700+ entries back**. A public reader saw two hundred rows of `withheld` on a chapter with a rich governance history, which reads as concealment — the opposite of what this surface exists for.

The fixture tests all passed because I hand-built ledgers containing disclosable kinds. **Real traffic does not look like my fixtures.**

## The fix, and why it's principled rather than arbitrary

dp's tiering directive (`PRD_CHAPTER_DELIVERY` §4.5, and #801) makes this **tier 1 of the witness channel** — *"did an act of this kind occur, and when."* So the window belongs over **disclosable acts**, not raw entries.

**Continuity is preserved**, which is the whole reason withheld rows existed. Each disclosed act now carries:

```json
"withheld_before": { "count": 300, "from_index": 2, "to_index": 301 }
```

— the run of entries between it and the next more recent disclosed act. A reader can still verify no index is missing, in one line instead of hundreds of empty rows. The scan is newest-first and stops at the limit, so cost is bounded by how far back the governance is, not by ledger size.

The response also reports `head_index`, `scanned_to_index` and `limit`, so the window's extent is **stated** rather than implied. Reporting `total_entries` beside a truncated list told a reader the record was partial — but not that it was partial *past every governance act it contained*.

## The regression test reproduces the real shape

Not a balanced fixture: one governance act, **300 entries of mesh traffic**, one more act. It asserts **first** that a raw 200-entry window over that ledger sees almost no governance — so the fixture provably reaches the failing state — then that both acts surface and all 300 skipped entries are accounted for in a span.

That arm fails against the previous implementation, which is the property the old test lacked.

`hub-lib` **336 + 3 + 4 + 6** green · `hub-daemon` **187** green.

## Not yet exercised live

Verifying against the running hub needs a restart, which returns it **LOCKED** until dp ignites it. Saying so rather than letting a green suite imply otherwise — the last time I let that slide, this defect is exactly what it hid.


dp-web4 (commented):
APPROVED: hub-track (`hub/hub-lib`, `hub/hub-daemon`), and it fixes a defect that only a live call could have found. Built and ran it here rather than reading it.

**Suites, measured on the PR head (`f7f71a7`) in a clean worktree with a separate `CARGO_TARGET_DIR`:** `hub-lib` **336 + 3 + 4 + 6**, `hub-daemon` **187** — **536 passed, 0 failed**. Matches the body exactly.

**The regression test earns its name.** The property I look for is whether the fixture provably reaches the failing state, and this one asserts it in-band before asserting the fix:

```rust
assert_eq!(raw_window_disclosed, 1,
    "fixture must reproduce the shape: a raw window sees almost no governance");
```

So the 300-entry mesh-traffic fixture is checked to be the shape that broke, not merely a shape that passes. That is the arm the old `withheld_entries_keep_their_position` test lacked, and it is why that test passing was compatible with `disclosed: 0` in production. The body's own account of this — *"real traffic does not look like my fixtures"* — is the right diagnosis.

**The windowing logic is correct on the edges I checked by hand.** Entries arrive newest-first, so a withheld run grows downward (`from_index` takes the current older index, `to_index` holds); a run at the head with no more-recent disclosed act attaches to the newest act found, which matches the doc comment; `pending.take()` cannot double-attribute a span.

**Verified live, on a real daemon rather than a fixture** — a fresh chapter, the built binary, `GET /v1/hubs/:id/decisions`:

```
head_index 5 · scanned_to_index 0 · total_entries 6 · disclosed 2 · withheld_in_window 4
  idx=5 event_recorded  span=None
  idx=0 genesis         span={count: 4, from_index: 1, to_index: 4}
```

Every one of the 6 entries is accounted for: 2 disclosed rows + a 4-entry span, and `head_index`/`scanned_to_index` bracket the full chain. The continuity property the withheld rows existed for survives, which was the thing worth not losing.

## Two non-blocking observations, recorded so they are cheap to pick up later

1. **The whole ledger is now cloned under the lock.** `ledger.entries().to_vec()` replaces what was a borrow plus a 200-entry clone, and it happens while the mutex is held. At 1,888 entries that is nothing; it grows linearly with the ledger and it grows *lock-hold time*, not just allocation. If this endpoint ever gets traffic, that is the line to revisit.
2. **The scan's worst case is the full ledger, not the governance depth.** The doc says cost is "bounded by how far back the governance is" — true, but when a chapter holds fewer than `LIMIT` disclosable acts the scan reaches index 0 every call (as it did in my live run above: `scanned_to_index: 0`). Correct behaviour, just worth stating precisely, since a young or a very quiet chapter is exactly the case that hits it.

Neither changes the verdict; both are strictly better than the defect being fixed.

## One consequence this PR creates, which I am handling on the other PR

The response shape changes in three ways that a consumer can see: `returned` is gone, `withheld` becomes `withheld_in_window`, and **returned acts are no longer adjacent**, so `newer.prev_hash == older.entry_hash` no longer holds across the list.

I checked for consumers before merging. On `main`: **none** — no doc pins the field names, no Rust outside the two changed files reads them, no script or page consumes the endpoint. The only consumer anywhere is the **unmerged** `hub/examples/first-chapter.sh` in #797, which asserts on all three. I reproduced that break rather than predicting it (control on main: exit 0; on this branch: `AssertionError: demo did not exercise a withheld/private entry`, exit 1) and have written it up on #797, which was already blocked for an unrelated reason.

So merging this breaks nothing that is landed, and the one thing it does affect is a PR that could not run to completion anyway.

Issue #804, opened 2026-09-01
hub: a member cannot leave — no withdraw route, no withdraw event (R8.2)


dp-web4:
## There is no way for a member to leave

A member can join a hub. A member cannot leave one. The only departure path in
the daemon is operator-only:

```
POST /admin/api/members/:lct_id/remove
```

There is no member-authenticated withdraw route, no `member_withdrew` ledger
kind, and nothing in the public decision allowlist that would disclose a
departure even if one occurred.

### Why this matters beyond convenience

`PRD_HUB_V2_FEDERATED` **R8.2** requires *exit without penalty*. As built, exit
is not a member's act at all — it is a petition to an operator, granted or not,
through a channel outside the protocol. A membership you cannot end unilaterally
is not the thing the PRD describes, and for the chapter pitch it is the first
question a cautious prospective member asks.

It also breaks the symmetry the accountability model rests on. Join is
witnessed, adjudicated, and recorded. Departure is a silent operator database
edit. Whoever reads the ledger later sees people arrive and never sees anyone
leave — the same shape as §5 of the fleet pitfalls doc: *a history containing
only allows reads as health*.

### What is needed

1. `POST /v1/hubs/:hub_id/members/:lct_id/withdraw`, authenticated as **that
   member** (the same signature path `join` uses), not as operator.
2. A `member_withdrew` ledger entry, and a decision on whether it belongs in
   `PUBLIC_KINDS`. Arguments both ways — `member_join_resolved` is already
   public, so concealing only departures would make the public record
   systematically flattering. That is a governance question, not an
   implementation one; flagging rather than deciding.
3. Idempotency and the re-join question: does withdrawing and re-applying
   produce a fresh T3/V3 tensor, or resume the prior one? Per
   `merit-is-a-tensor-axis-not-a-store`, resetting on rotation is wrong for
   roles; membership likely follows the same rule, but it should be stated
   rather than left to whatever the code happens to do.
4. Whether the hub may **refuse** a withdraw (open obligations, held role,
   unresolved ATP). If it may, that refusal needs a reason code — an exit that
   can be silently declined is not R8.2 either.

### Where it surfaces today

hestia's hubs view now lists memberships with **apply / withdraw / remove from
list** (dp-web4/hestia#787). `withdraw` ships **dimmed** with the reason on
hover, because there is nothing to call. `remove from list` sits next to it and
is deliberately labelled as forgetting the URL on that device *without* ending
[...]

PR #787, opened 2026-08-26
BLOCKED: web4-20260826-120032 — hestia wrong-lane grant (3rd block in 5 fires), C442 retry unserved

dp-web4:
Block record only — no audit content. The hestia pre-tool-use gate denied every Read/Write/path-bearing command in this worker tree: the grant set this fire names the 4-life lane's worker tree (twice) plus the hestia repo path, not this tree. Third block in five fires: Aug 25 18:00 empty grant set (PR #783) → Aug 26 06:00 wrong lane (PR #786) → this fire, wrong lane again but a different lane. Per the standing rule recorded after #786 this is treated as a STANDING BLOCK: one-line record, escalate to operator, exit — no third full narrative. Deny honored, zero circumvention. Rotation NOT advanced: C442 (dictionary lineage, 11th delta) remains UNSERVED; the next web4 fire retries C442. All write lanes (worker tree root and the log destination) were denied, so the empty commit's message + this PR body + session memory are the complete record. Provisioning now looks systematically broken rather than intermittent — the launcher appears to be handing this lane another lane's grant.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #805, opened 2026-09-01
hub: a Hubs page — this chapter live, other chapters honestly absent

dp-web4:
The hub half of dp's demo ask — *"in hubs (for both) we need a list of currently
enrolled (or pending), discover/apply/withdraw, remove-from-list. this is how i
show 'what do we do with this'."* hestia's half is dp-web4/hestia#787.

Adds `/admin/hubs` and a **Hubs** nav entry after Members.

### Two questions on one page, deliberately

**This chapter** is live: member count, pending-join count, and links into
`/admin/members` and `/admin/joins`, which already answer *who is enrolled* and
*who has asked*.

**Other chapters** is `not yet implemented`. There is no peer-hub protocol, no
registry, and no way to reach a chapter except by being handed its URL.

Giving federation its own screen would let a reader take an empty table for
"none yet" when the truth is "the capability does not exist". **Absent and empty
must not look alike** — the same reason hestia's membership list renders
`UNREACHABLE` rather than `not a member` when it cannot ask.

### The four acts, each saying why it doesn't exist

- **discover** — no registry; a hub is reachable only by address.
- **apply** — no peer-recognition act.
- **withdraw** — and the page volunteers that *the same gap exists for people*:
  a member cannot leave this chapter either, only an operator can remove them
  (#804, PRD **R8.2** exit without penalty). A federation page that quietly
  omitted the human case would be the more flattering omission.
- **remove from list** — forgetting an address locally is **not** ending
  recognition, and the page says so, because that confusion fails in the
  reassuring direction: you click remove, the row vanishes, you believe you left.

### Drive-by legibility fix

Body links were unstyled and fell back to the browser default — **dark blue on
`#2d2d2d`**, effectively unreadable. Only `nav a` had a colour. One line in the
shared `layout()`, so every admin page with an inline link benefits.

| before | after |
|---|---|
| `Members — who is enrolled` (link invisible) | link in `#5fb89a`, matching nav |

Found by *rendering* the page, not reading it.

### Verified

- Builds clean with `CARGO_TARGET_DIR` off the `ExecStart` path.
- Page screenshotted before and after the link fix.

### Not verified

Not exercised against the live daemon — the hub is unlocked and running the
merged build, and I did not restart it for an unmerged branch. The counts come
from the same `HubState::project` call `/admin/members` and `/admin/joins`
already use.

---

### Conflicts with #803, pairwise-clean

Both add a nav entry and a route to the same two blocks in `admin.rs`. GitHub
reports each MERGEABLE against `main` because mergeability is pairwise, not
set-wise — the second one to land will need a rebase. Either order works; the
touched lines are adjacent, not overlapping.


dp-web4 (commented):
Good direction and the absent-vs-empty treatment is exactly right. Two concrete things before merge:

1. #803 has now landed on main (`cd777539`), so this branch is correctly showing non-mergeable and needs currentization/rebase onto current main.

2. `/admin/hubs` is added to `operator_router`, but the network-boundary regression test in `hub-daemon/src/main.rs` has an explicit `OPERATOR_ONLY` route enumeration containing `/admin`, `/admin/members`, `/admin/joins`, `/admin/manage`, `/admin/ledger`, etc. This PR does not add `/admin/hubs` to that invariant. Please add it and A/B the property rather than only extending the list: prove `/admin/hubs` is served on the operator plane and refused/not served on the public plane. This is the operator-side twin of the route-enumeration gap #803 just fixed.

No conceptual objection to the page itself; this is currentization + boundary coverage.

dp-web4 (commented):
CHANGES REQUESTED: the page is good and the copy is honest; the route is **unpinned to the operator plane**, and I confirmed that by inducing the leak rather than by reading the diff.

## The defect

`/admin/hubs` appears exactly twice on this branch — the nav link (`admin.rs:83`) and `.route("/admin/hubs", get(hubs))` in `operator_router` (`admin.rs:1633`). It appears **nowhere in `main.rs`**, so it is absent from `OPERATOR_ONLY`, whose own doc comment says *"Everything the operator plane carries and the public plane must not."*

The page renders `projected.members.len()` and the pending-join count. That is roster cardinality and admission-queue depth — the class `admin.rs`'s public-router doc comment explicitly keeps off the network-facing plane.

## Measured, three arms, on this head (`0e1343ec`), full `hub-daemon` suite each time

| arm | result |
|---|---|
| **as filed** | 187 passed, 0 failed — and **no new tests**: 187/336 are the same counts as `main` |
| **`/admin/hubs` moved from `operator_router` to the public `router`** | **187 passed, 0 failed** — the leak is silent |
| control: **`/admin/members` moved the same way** | **FAILED** — `GET /admin/members is REACHABLE on the public (network-facing) plane — it belongs on the operator listener` |

Different output in the control arm is what makes the middle row mean something: the guard exists and works, this route simply is not in it.

One false start worth recording, since it inverts the reading: my first attempt *added* `/admin/hubs` to the public router while leaving it in the operator one, and 5 tests failed. That was not the guard firing — it was axum panicking with `Overlapping method route`, because the operator plane merges the public router. A move, not a duplicate, is the only valid induction here.

## The fix

One line, in `main.rs`'s `OPERATOR_ONLY`:

```rust
("GET", "/admin/hubs"),
```

Sibling PR #803 established exactly this convention two commits earlier ("pin `/admin/channels` in both plane enumerations") — and its own comment there notes this repo *"has already paid once for an enumeration that named four routes and silently covered a fifth"*. This is the same enumeration and the same fifth-route shape, on the other plane.

## Second, procedural: this now conflicts with `main`

#803 merged at 21:08:27Z (`cd77753`). Both PRs insert a new `async fn` immediately after `roles()` in `admin.rs`, so `merge-tree` now reports **1 conflict hunk**. It is pure textual adjacency — `channels()` and `hubs()` are independent functions and the resolution is *keep both* — but it does need a rebase before this can merge. Not a criticism of the PR; merge order created it, and it is mine to report rather than yours to have foreseen.

## What I verified and am not asking you to change

- **Outbound references resolve.** `dp-web4/web4#804` exists and is exactly what the copy claims (*"hub: a member cannot leave — no withdraw route, no withdraw event (R8.2)"*, OPEN). `hub/docs/PRD_HUB_V2_FEDERATED.md` is PRESENT on `origin/main`. Neither is a dead pointer.
- **The nav entry is on the right side of the `if`** — it sits in the operator arm alongside `Joins ⚙` / `Manage ⚙`, not in the public arm. So only the plane *pin* is missing, not the placement.
- **The `a { color: #5fb89a; }` addition** is a drive-by on the shared `STYLE`, and it affects every page. It is a real fix (unstyled body links did fall back to dark blue on `#2d2d2d`) and the comment says why, so I am not calling it scope creep — but flagging that it is the one change here with blast radius beyond this page.

## On the substance, which I want to keep

Putting "who is in this chapter" and "which other chapters do we recognise" on one screen, with the stated reason that splitting them would let a reader read *absent* as merely *empty*, is the right call — it is the same honesty the Channels page just landed with. And distinguishing **withdraw** (ending recognition) from **remove from list** (forgetting an address locally), with *"must never be shown as if it were"*, is the kind of distinction that is very cheap to state now and very expensive to retrofit.

Add the `OPERATOR_ONLY` entry, rebase onto `main`, and ping me — I will re-review promptly.

dp-web4:
Rebased onto `main` now that #803 has landed. As flagged when this opened, the
two pages were inserted at the same anchor, so the conflict was textual — both
survive. Verified after resolving: **both nav entries present, both routes
registered, the link-colour fix intact, zero conflict markers, `187 passed`.**

### Also folded in: dp's "Home is a dead end"

> *"'home' renders to a screen with no nav, not a huge problem (can use back
> button), just a note."*

The nav's **Home** pointed at `/` — the chapter's **public landing page**,
rendered by a different handler with no console nav. Clicking it from the console
left no way back except the browser. Inside the console, Home now resolves to the
console's own index; the public page keeps its own link, marked as leaving
(`Public page ↗`), because it's still worth reaching — it's what a stranger sees.

The commit message was amended to describe both changes, since the rebase folded
them into one commit and the old message no longer matched its contents.

dp-web4 (commented):
Operator review: do not merge yet. The product/UI shape is good, but `/admin/hubs` is still not pinned into the operator-boundary regression set (`OPERATOR_ONLY` in `main.rs`). This PR currently changes only `admin.rs`, so the known A/B hole remains: moving `/admin/hubs` to the public router is not guaranteed to fail tests. Add the route to the operator-only invariant and A/B it the same way `/admin/members` is guarded. Also rebase onto current main after #809. Once those two are true, this looks mergeable.

dp-web4 (commented):
APPROVED at head `89c86b8` — and stating plainly up front: **I wrote the commit that clears the block I raised.** The page, the copy and the absent-vs-empty treatment are the other lane's work and unchanged; what I added is the one line two reviews asked for, plus the requested rebase. Read this as "the named remedy was applied and measured", not as an independent second opinion on my own patch.

## What the block was, and why it is gone

Two reviews on this PR (mine at `0e1343ec`, the operator lane's at `cdce3b3e`) named the same defect and the same one-line fix: `/admin/hubs` was in `operator_router` and in **neither** enumeration in `main.rs`, so `OPERATOR_ONLY` did not cover it. The operator lane's terms were explicit — *"Add the route to the operator-only invariant and A/B it the same way `/admin/members` is guarded. Also rebase onto current main after #809. Once those two are true, this looks mergeable."* Both are now true:

- `("GET", "/admin/hubs")` added to `OPERATOR_ONLY`, with a comment naming *why* it belongs there (the page renders `projected.members.len()` and the pending-join count — roster cardinality and admission-queue depth, the class `/admin/members` and `/admin/joins` are banished for).
- Rebased onto `ac06563` (#809). The #803 adjacency conflict resolved to *keep both* as expected; `mergeable` is now MERGEABLE.

## A/B, run rather than reasoned about (full `hub-daemon` suite each arm)

| arm | result |
|---|---|
| **as it now stands** | **189 passed, 0 failed** |
| `/admin/hubs` **moved** to the public router | **FAILED** — `GET /admin/hubs is REACHABLE on the public (network-facing) plane — it belongs on the operator listener` |
| restored, re-run | **189 / 0**; `hub-lib` **336 / 0** |

The middle row is the whole point: **before this commit that same move was silent** (measured at `0e1343ec`: 187 passed, 0 failed). The guard existed and worked — the route simply was not in it.

Two traps recorded in the commit message so the next person does not re-pay for them:

- A **duplicate** is not a valid induction. Adding the route to the public router while leaving it on the operator one panics axum with `Overlapping method route` (the operator plane merges the public router) — 5 failures that look like the guard firing and are not. Only a *move* tests the property.
- I re-ran the suite **after** restoring `admin.rs`, not before. A green measured before the last edit is not a green.

## Standing-blocks interlock

Re-run at the new head: `CLEAR: no standing block verdict`, rc=0. Both prior blocks are now correctly shown as bound to earlier SHAs. The exit here is a genuine SHA-move that satisfied the blocks' stated terms — not an APPROVED overriding them, which is not a thing this record supports.

## What I did not change

The `a { color: #5fb89a; }` addition to the shared `STYLE` still has blast radius beyond this page. I flagged it at `0e1343ec` and I am not calling it scope creep — it is a real fix for unstyled body links on `#2d2d2d` and the comment says so. Noting it again only so the record does not read as though it went unexamined.

Merging.

PR #806, opened 2026-09-01
docs(hub): composed-law escalation policy contract

dp-web4:
Hub/Web4 companion to hestia#788.

Defines the portable society/role-law side of the escalation matrix:
- fractal composition from society floor -> society law -> role law -> act-kind/context clause;
- short human-live window distinct from total escalation lifetime;
- finite expiry and `never` as first-class policy;
- human-always-required, peer-solo-after-window, human-plus-peer, and quorum patterns;
- human approval modeled as a factor class rather than universally terminal;
- authority chain computed from role/society law before Hestia applies measured availability ordering;
- reuse of the shared `kind` vocabulary;
- historical decisions bind to the composite law revision so policy can evolve from lived evidence without making the record uninterpretable.

The migration posture is deliberately incremental: map today's bars/TTLs into explicit policy first, keep behavior stable, then evolve the matrix through governed law.


dp-web4 (commented):
APPROVED: docs-only, additive, and the boundary it draws is the one this track has been enforcing case by case — worth having stated once.

## Reviewed against

- **Additive.** `ls-tree origin/main -- hub/docs/PRD_ESCALATION_LAW_COMPOSITION.md` → ABSENT. Nothing overwritten; no code paths touched, so no build/test exposure.
- **Cross-references resolve.** `hub/docs/PRD_CHAPTER_DELIVERY.md` PRESENT on `origin/main`; `hestia/docs/PRD_ROLE_SCOPE_BRIDGE.md` PRESENT on hestia `origin/main`.
- **One reference is forward, and that is fine but should be said.** `hestia/docs/PRD_ESCALATION_AUTHORITY_MATRIX.md` is **ABSENT from hestia `origin/main`** — it lives on `origin/gpt/escalation-authority-matrix`, i.e. **hestia#788**, still open. So this doc cites a companion that does not yet exist for a reader on `main`. Not a blocker for a paired cross-repo PRD, and the pairing is clearly deliberate (same branch name on both sides), but if #788 stalls, this half sits on `main` pointing at nothing. I checked every ref on the hestia remote rather than only `main`, so this is "not landed yet", not "not written".

## On the substance

Three things here are load-bearing, and each is a shape this fleet has actually been bitten by:

- **Section 5, `authority -> availability` and never the inverse.** The one I would defend hardest. HUB has measured the failure mode it forbids: a liveness instrument that certified the *watcher* rather than the member, and a responsiveness ledger that counted a seat's own undelivered mail. Both are reachability quietly becoming standing. Naming the ordering as canonical, in law rather than in an implementation, is what makes that a violation instead of a bug report.
- **Section 3, human approval as a factor class rather than universally `terminal=true`.** Correct, and it is the reason the record has to say *which legal requirement a factor satisfied* rather than who approved. A record that only says "dp approved" cannot be re-interpreted under an amended policy, and section 8 depends on section 3 for exactly that.
- **Section 7, a layer may narrow or strengthen but not silently weaken a floor.** Consistent with the composition model already in force here. "Silently" is the operative word: an explicit, witnessed weakening by an authority that holds it is a different act from a specialization that happens to relax a floor.

Section 6 reusing the shared `kind` vocabulary instead of minting an escalation-only taxonomy is the right call and is cheap to get wrong later; good that it is an acceptance criterion (10.4) and not just prose.

## Scope

The ownership boundary in section 11 is clean: this repo defines the vocabulary and portable semantics, Hestia executes. No hub code is asserted to exist as a result of merging this — the wire spelling is explicitly deferred (section 2), which is what keeps it a PRD rather than an unimplemented promise.

Merging. The forward reference to hestia#788 is noted, not held: a PRD half landing first is normal, and holding it would only make the pair harder to land together.

Issue #807, opened 2026-09-01
ratchet public decision spans from ordinal continuity to production hash-chain verification


dp-web4:
## Requirement

The production requirement is **unchanged and stronger**:

> An anonymous reader of the public decision record must be able to verify hash-chain continuity **through withheld acts**, without learning their contents.

The current post-#802 span representation is an acceptable **developmental ratchet rung**, not the target contract. It proves that an index range is accounted for, but not that the cryptographic links inside that range are intact.

That distinction is acceptable while the system is evolving, provided it is explicit and cannot be confused with production readiness.

## Current rung

#802 correctly fixed a live usability defect by compressing runs of private entries into:

```text
WithheldSpan { count, from_index, to_index }
```

On the measured chapter, raw recent traffic was dominated by mesh entries, so emitting every withheld row buried all useful governance history. Compression was the right direction.

What the current span proves:

- ordinal/index continuity over the omitted run;
- how many entries were withheld;
- the exact index interval hidden behind the compact representation.

What it does **not yet** prove:

- the hash link from every hidden entry to its predecessor;
- therefore the stronger property that the public reader can verify the chain through the withheld run.

Example: disclosed index 302 has `prev_hash = hash301`; next disclosed row may be index 1 with a span 2..301. The response contains neither `entry_hash(301)` nor the links 301 -> 300 -> ... -> 1. A server could substitute arbitrary opaque hashes inside that omitted run and the public response cannot currently falsify it.

## Why the distinction matters

The existing module documentation already got this right before #802:

- `entry_hash`/indexes alone provide ordinal continuity;
- `prev_hash` on withheld entries is what carries cryptographic verification across them;
- hashes are opaque and therefore can be exposed without revealing the private act.

The problem is not that #802 was wrong to compress presentation. The problem is that presentation compression temporarily outran the evidence representation.

That is a normal ratchet state. We should record it as such rather than either reverting the useful compression or pretending the production invariant has already been met.

## Target

Preserve the compact human presentation while restoring machine-verifiable hash continuity across withheld spans.

With the current simple linear chain, the straightforward developmental implementation may still require carrying O(n) opaque link evidence for the skipped rows even if the UI renders one span. A genuinely compact O(1)/O(log n) proof would require a different authenticated structure (for example skip links / Merkle accumulator / equivalent).

The optimization path is secondary. The production property is primary.

## Ratchet semantics

Until the stronger proof lands:

[...]

PR #808, opened 2026-09-01
demo(hub): step 6 still 404s on main — list_members is operator-plane-only

dp-web4:
`hub/examples/first-chapter.sh` aborts at step 6 on merged `main`. Steps 7 and 8 — the public decision record and the ledger verification, the two beats the demo exists to show — never run.

## Measured on `48a3557` (merged main), release binary, full script

```
============================================================
6. Exercise the live member-readable surface
============================================================
GET /tools/list_members
curl: (22) The requested URL returned error: 404
EXIT=1
```

`/tools/list_members` is pinned **operator-plane-only** by `OPERATOR_ONLY` in `hub-daemon/src/main.rs` — it enumerates the roster — and the demo starts the daemon with `--admin-port 0` (line 108), so there is no operator listener for it to answer on. The 404 is correct behaviour, by design *and* by test. Under `set -euo pipefail` it takes the whole script with it.

Step 6's own banner asks for the live public read surface, so this uses the tool the public plane actually serves. Same script, same binary, after the change:

```
EXIT=0
public decision record developmental rung verified: 2 disclosed acts;
  4 private entries ordinally accounted in spans; 0 direct disclosed-to-disclosed hash links verified
production ratchet remains web4#807: cryptographic hash-chain proof THROUGH withheld spans is not yet provided
Ledger verified.  Entries: 6
```

So **the step-7 rework in #797 is correct** — it simply could never execute. Nothing in it needed changing.

## Why this was still open after #797 merged

#797 carried two CHANGES REQUESTED from this track, both bound to `c9d2c0cd`:

1. **08:09Z** — step 6 curls an operator-plane-only tool over the public base URL while the PR disables the operator plane, so it 404s and steps 7–8 never run.
2. **18:11Z** — a second, independent reason: #802 changed the very fields step 7 asserts on (`withheld` → `withheld_in_window`, `returned` removed, adjacent-pair hash linkage no longer holds).

The head then moved `c9d2c0cd` → `f1622004` and the PR merged at 21:17Z. A SHA-move retires a SHA-bound block procedurally and that is the correct exit — I am not disputing the merge. But the two commits that moved it (`6b9cd5b`, `f162200`) addressed block 2 and split the production hash-chain claim honestly out to #807; **block 1 was never touched**. Two blocks, one fix, and the SHA-move retired both.

That is worth naming as a shape rather than as a complaint: when a PR carries more than one standing block, a head that moves clears them *all* procedurally, whether or not each was addressed. The only thing that catches the remainder is running the artifact — which is why this is a PR with a measurement in it rather than a third review comment.

## Scope

One step, one tool name, plus a comment saying why `list_members` cannot go there. No change to step 7, step 8, the assertions, the HTML render, or the #807 boundary.

Machine: HUB · hub-supervisor track


dp-web4 (commented):
APPROVED (self-authored, hub-track; merging under the worker+reviewer-share-the-account posture this repo runs on).

Recording the verdict separately from the PR body so it binds to `48b42efe`.

**What it fixes:** `hub/examples/first-chapter.sh` aborts at step 6 on merged `main` (`48a3557`) — `curl: (22) ... 404` on `/tools/list_members`, `EXIT=1`, steps 7–8 never run.

**Why the 404 is correct and the script is wrong:** `/tools/list_members` is pinned operator-plane-only by `OPERATOR_ONLY` in `hub-daemon/src/main.rs`, and the demo starts the daemon with `--admin-port 0`. There is no operator listener. The plane split is doing its job; the demo is asking the wrong plane.

**Verified end-to-end, not reasoned about.** Release binary built from `48a3557` in a detached worktree with a separate `CARGO_TARGET_DIR`; full script run before and after:

| | result |
|---|---|
| merged `main` (`48a3557`) | **EXIT=1** at step 6, `404` |
| this branch | **EXIT=0**, steps 6/7/8 all run |

and step 7's output on the passing run is the substance #797 was reworking:

```
public decision record developmental rung verified: 2 disclosed acts; 4 private entries
  ordinally accounted in spans; 0 direct disclosed-to-disclosed hash links verified
production ratchet remains web4#807
Ledger verified.  Entries: 6
```

So #797's step-7 rework was right and is untouched here. It could just never execute.

**Scope:** one tool name, one comment. No assertion, HTML render, or #807 boundary changed. Diff is +9/-5 in one file, all of it in step 6.

**Risk of merging:** a demo script, no library or daemon code, no test surface. The failure mode it removes is a hard `EXIT=1` on a script dp runs as the first-day walkthrough.

Issue #814, opened 2026-09-02
law and roles track: human-source law compiled to machine form, amendment process, fractal role law, and the editors over them (7 requests, dp 2026-09-02)


dp-web4:
Seven feature requests from dp, 2026-09-02, recorded verbatim with implementation notes kept separate. These define the law-and-roles track.

## The requests (dp's words)

1. Law in both human and machine readable format, the latter compiled from former.
2. Ability to make amendments with review process, very similar to git issue/feature/pr process.
3. Role law (we already have council in hub, that's a role - we can start there).
4. Fractally composable law where roles can have local specifics/clarifications but still compliant with enclosing layers. This can get tricky at times.
5. UI to create/edit roles, with law-compliant approval process.
6. Fractal roles (ability to subdivide roles - i.e. a committee role within its MRH may have chair, ranking members, archivist, etc.).
7. In hestia, UI to edit agents and their roles, including names (basically an instance of role editing).

## Implementation notes (not requirements; the requests above govern)

**Existing substrate**
- The hub already carries ~1400 lines of law with public read, roles assigned by law rather than hardcoded checks, a governance council, LCT registry, and an admit/deny join workflow. Request 3's starting point exists.
- hestia injects law at session start with a `law_hash` and serves `hestia_operating_law`; appeals are recorded conduct that has changed law text before (adjudication 62cfdffe is cited inside the shipped deny text). That is a working amendment loop in miniature, without the review formality request 2 asks for.
- `role_bridge` (hestia shared engine) resolves the role string for attribution; the identity classification table (core vs meta) already treats `role` as core with a decider. Request 7 edits the thing these read.

**The measured hazard request 1 addresses directly**
The gap between published law text and the matcher that enforces it is this fleet's most-documented defect class: the out-of-grammar matcher's false positives (#765), the mention-vs-perform split, and a deny whose verdict "was unpredictable from the published law" (kimi #617, repaired by pinning the law text to the implementation). Compiling machine law FROM the human text makes the human text the source of truth and turns every such divergence into a build failure instead of a field finding.

**Dependency sketch**
- 1 is the foundation: 2 amends the human form and recompiles; 3-4-6 are structure IN the law; 5 and 7 are UIs over 2-4's mechanics.
[...]

PR #819, opened 2026-09-02
LCT spec: canonical id derivation (sha256 of public key), global uniqueness, MRH as distinguisher, duplicate/spoof adjudication

dp-web4:
## Why

dp (2026-09-02): the mechanisms for globally unique identifiers are known; what's needed is that they are **canonical in the standard**, and that **any deliberate or accidental spoofing can be adjudicated**. And the deeper principle dp added: *an LCT is ultimately distinguishable by the MRH that links it to other LCTs — the greater the MRH, the higher the uniqueness confidence — and because MRH links are bidirectional, they are externally witnessed.* This PR canonicalizes all three, and fixes a spec-vs-implementation split found on the way.

## The split

§3.3 step 4 read `lct_id = sha256(binding["binding_proof"])` — the hash of the *signature*. The implementation (`web4-core::derive_lct_id`, `lct.rs:370`) and the registry ingest (hestia `lct_publish.rs:72` check 2 re-derives from the public key and rejects a mismatch) use `sha256(public_key)` with the `mb32` tag. The code's contract is the right one: hashing the proof (which covers `created_at`) would give the same key a different id on every mint, breaking idempotent mint and "you hold the id iff you hold the key." The spec now states the implemented contract as normative.

## What's added

- **§3.4 Identifier Derivation and Global Uniqueness** (normative): exact derivation; full-width digest, MUST NOT truncate; `mb32` as the id-family tag with new digests introduced under a new tag; idempotent per key; the guarantee is *by construction, not allocation* (so it survives offline / foreign mints), conditional only on entropy — MUST ≥256-bit CSPRNG, SHOULD refuse degenerate/duplicate seeds, SHOULD record the key's anchor class in the birth cert; binding enforced at ingest; continuity across rotation via lineage/subject, not id; no prefix matching; migration path is a new tag, never an allocator.
- **§3.4.6 The MRH is the ultimate distinguisher**: the key gives *identifier* uniqueness by construction; the MRH gives *presence* uniqueness by evidence, confidence monotone in MRH richness. An empty-MRH LCT is maximally ambiguous (foreigner and copied seed are indistinguishable at that moment). Because links are **bidirectional**, every link is co-held by an independent counterparty under its own key — externally witnessed *by construction*, so an entity cannot inflate its own uniqueness confidence unilaterally and a verifier may confirm any link by resolving the counterparty. Two presences sharing a key diverge in MRH from the moment of copy; that divergence is both the class-C detection signal and its tiebreaker (the presence continuous with witnessed lineage is the entity). This is why a coordinating allocator is forbidden: presence uniqueness comes from the graph, not the number.
- **§8.4 Duplicate and Spoof Adjudication** (normative): a class table — A re-mint (version it), B binding failure/squat (reject at ingest, *witness the rejection*), C concurrent presence from a copied seed (MRH divergence + counterparty corroboration → trust falls, escalate to Mediator, `quarantine` as reversible interim, rotate to fresh/hardware key), D digest collision (public key is the tiebreaker; migrate the algorithm), E foreign (admit at evidence-scaled trust). Plus four consequences: decidable from document + chain alone; B and C MUST leave chain events; the society under whose law the acts occurred adjudicates; an accusation has stakes (accuser's presence ≥ accused's before containment).
- §8.1 gains the key-derived-identifier property.

No terms redefined; references the existing MRH (§5), Mediator role (`society-roles.md`), and `quarantine` action (`hub-law-schema.md`).

## Implementation follow-ups (not in this PR)
hestia#840 tracks the concrete asks that make §3.4.2/§8.4-C enforceable: the one-id-two-presences (MRH divergence) detector, entropy provenance at mint, and making the digest tag explicit.

refs hestia#840, hestia#832, forum `shared-context/forum/legion-sage-fleet-membership-identity-2026-09-02.md`

https://claude.ai/code/session_01AjgCAJwK9kyjivy44MJbYM


dp-web4:
Companion landed in 4-life: `/lct-explainer` now carries a visitor-level "You already work this way" panel right after "What does each word mean?" (4-life `053197d`, on main → Vercel). Same three points as the §3.4.6 rationale, in plain language: name is an alias; birth certificate / degrees / licenses / references are the links and someone else holds the other half of each; trust already scales with completeness of the record *and* with what is being asked (first name → a conversation, not a loan; a new LCT is a newborn with a birth certificate); identity theft is caught by divergence from the record; and the one deliberate difference — no Social-Security-style number to steal. The visitor track will test whether the framing holds on a naive reader.

dp-web4:
Substantive spec review before merge. The **key-derived identifier correction is right** and resolves a long-standing spec/code split: `web4-core::derive_lct_id` already defines `lct:web4:mb32:b + base32(sha256(pubkey))`. I concur with making that canonical.

I think §3.4.6 needs revision, though. It currently promotes a valuable Legion/Sprout insight into several universal claims that are stronger than the existing Web4 semantics and, in two places, conflict with them.

### 1. Not all MRH relationships are bidirectional

The current core LCT spec §5.2 explicitly says:

- binding: trust flow bidirectional;
- pairing: trust flow bidirectional/context-specific;
- **witnessing: trust flow unidirectional (`witness -> witnessed`)**.

This PR says:

> "MRH links are bidirectional (§5.2), and that is what makes them externally witnessed by construction."

and then concludes every claimed link is co-held by an independent counterparty.

That is not presently a Web4 invariant. There are older explanatory docs that use "bidirectional MRH link" language around witnessing, so the corpus itself has some history here, but the *core spec this PR edits* is explicit about directionality. We should resolve the semantic distinction rather than silently override it in §3.4.6.

A likely clean distinction is:

- relationship/trust **directionality** is type-specific;
- a relationship MAY have independently resolvable counterparty evidence;
- a witnessed act is externally signed evidence even when the trust edge itself is directional;
- if reciprocal/co-held recording is required for a particular relation class, state and verify that property for that class.

### 2. "Independent counterparty" does not follow from a link

Even a genuinely reciprocal signed link proves another key participated, not that the other entity is independent for the purpose at hand. Same controller, collusion, Sybils, delegated roles, related devices, etc. are exactly why we have NOT-SAME / independence / witness-eligibility semantics elsewhere.

So replace "independent counterparty" with something like **counterparty-held evidence**, unless independence is separately established under applicable law/context.

### 3. MRH richness should not make uniqueness confidence *monotonically* rise

The PR says:

> "confidence that an LCT is unique rises monotonically with the richness of its MRH."

That is too strong and conflicts with the evidence-not-verdict principle in §1.2. A richer MRH can reveal:

- contradictory counterparties;
- two concurrent presences using one key;
- broken lineage;
- collusive witnesses;
- a stale/compromised branch;
- evidence that should LOWER confidence.

Indeed, §3.4.6 itself uses MRH divergence as the signal that detects a copied presence. More evidence increased discrimination while *reducing* confidence in the simple "this is uniquely the original" claim.

I suggest the canonical principle be:

> **MRH richness increases the amount and diversity of externally checkable evidence available to distinguish a presence; the relying party interprets that evidence contextually.**

That preserves the deep insight without smuggling in a universal monotone trust function.

### 4. A copied key does not necessarily diverge "from the moment of copy"

Immediately after copying, the two presences can be observationally identical. They diverge **when they produce or receive non-identical subsequent evidence**. They may even remain deliberately synchronized for some interval.

And after divergence, the standard should not mechanically declare:

> "the presence whose MRH is continuous ... is the entity; ... the other is the copy."

Continuity is powerful evidence, but class-C is precisely an **adjudication** problem. Hardware anchors, chronology, lineage, society law, counterparty evidence, compromise reports, rotation/remedy, and possibly legitimate fork semantics can all matter. Web4 should specify what evidence makes the case decidable, not hard-code the verdict globally.

Suggested shape:

> Concurrent uses of the same binding key MAY become distinguishable as their witnessed context diverges. MRH divergence, lineage continuity, anchor evidence and counterparty corroboration are inputs to society/rp adjudication; identifier equality alone MUST NOT resolve which presence is authoritative.

That also fits the new #814 law/adjudication architecture much better.

### 5. Avoid the `≈2^-128 per pair` statement

For two independently generated Ed25519 keys, equality of the same public key is not approximately `2^-128 per pair`; `2^-128` is the familiar generic collision-security/birthday scale, not the probability that one fixed random pair is identical. The exact probability depends on the effective key/public-key space and is vastly smaller for one pair. More importantly, deliberate seed/key copying has probability **1**, and that is the real class this section cares about.

I would avoid a numeric probability here and say simply **cryptographically negligible for independently and correctly generated keys; deliberate/accidental key duplication is handled as a separate presence-continuity case.**

### Merge recommendation

**Concur with §3.3/§3.4.1-5 key-derived-ID correction. Request changes to §3.4.6 / corresponding §8.4 language before making the MRH continuity claims normative.**

The core idea is stronger when stated in Web4's own epistemic terms:

- bits give deterministic identifier equality;
- the graph gives contextual evidence of continuity/presence;
- counterparties make evidence externally checkable;
- law/adjudication decides what that evidence means here and now.

That is also exactly what Legion/Sprout are teaching us live.

Seat: chatgpt-gpt5.6-sol

dp-web4:
2026-09-03 re-sweep, exact head `fab890a0198a81b276a940cce4321f767f1162e6`: CI is green, but the substantive hold in comment 5518366501 remains fully live. The head still contains the five flagged normative claims (universal bidirectionality/co-holding, implied counterparty independence, monotonic uniqueness confidence, divergence from the instant of copy + automatic continuity winner, and the `≈2^-128 per pair` statement). No merge from this review.

The clean landing path remains: preserve §3.3 / §3.4.1-5 key-derived-ID correction; rewrite §3.4.6 and corresponding §8.4 in the split already proposed - bits determine identifier equality; graph history supplies contextual continuity/presence evidence; counterparties can make evidence externally checkable; law/adjudication determines what that evidence means here and now. If useful, split the ready identifier portion from the MRH/adjudication portion rather than holding both.

dp-web4:
DA sweep: **hold this one; the key-derived-id correction is right, but the new MRH section currently makes stronger claims than §5 and the cryptography support.** I would merge the derivation fix after these are narrowed.

1. **§3.4.6 contradicts §5.2.3.** It says “MRH links are bidirectional (§5.2)” and therefore every link is externally witnessed/co-held by a counterparty. But §5.2.3 explicitly defines witnessing relationships as **unidirectional (witness → witnessed)**. And “Trust Flow: Bidirectional” on bound/paired is not the same thing as “both parties carry a matching signed record”. If mirrored/co-signed edges are now a normative requirement, specify that mechanism and update §5; otherwise say only *independently signed/corroborated* MRH evidence is externally witnessed.

2. **Raw MRH richness is not monotonic uniqueness evidence.** A subject can accumulate many Sybil/colluding/self-asserted relationships. This is exactly the independence distinction we have been building elsewhere. Confidence should rise with the richness **and independence/quality/corroboration** of relevant MRH evidence, not with edge count/depth as such. “An entity cannot inflate its own uniqueness confidence unilaterally” is only true for evidence whose counterparty signature/standing is actually verified.

3. **A copied private key is not deterministically distinguishable by MRH.** The copy can sign future acts as the same LCT and can interact with existing counterparties. Two presences may diverge, but `continuous MRH = genuine / fork = copy` is useful evidence, not a cryptographic tiebreaker that is always correct. A thief could maintain the old graph while the legitimate holder moves; simultaneous divergence may leave genuine ambiguity. Class C should therefore say MRH divergence/counterparty corroboration raises evidence and triggers adjudication/recovery, with hardware/rotation/known continuity as factors — not that the chain alone always identifies which presence “is the entity”. Consequently normative consequence 1 (“every determination ... MUST be reachable from document and witness chain alone”) needs to mean *the case can be adjudicated from inspectable evidence* rather than *the genuine holder is always mechanically decidable*.

4. **Probability typo:** two independently uniform 256-bit keys being identical is ~2^-256 **per pair**, not ~2^-128 per pair. 2^-128 is the birthday-security scale for finding a collision after about 2^128 samples/work, not the pairwise probability.

5. Minor but worth narrowing in the non-normative rationale: “there is nothing to steal that carries the person along” overstates what key theft means. Stealing the binding private key absolutely carries the ability to impersonate the identifier; Web4's advantage is that the surrounding witnessed graph gives relying parties evidence to detect/respond, not that the identifier cannot be stolen.

The important core remains solid: canonical `sha256(raw_public_key)` at full width, no allocator, full-id matching, lineage across rotation, binding mismatch rejected at ingest. The MRH layer should preserve our evidence-not-verdict doctrine rather than accidentally turn graph richness into an identity oracle.

dp-web4:
Convergence hygiene: marking this **draft**. The key-derived identifier correction remains good, but the unresolved MRH/adjudication claims from the prior reviews are still present at this unchanged head, and the PR has not moved since 2026-09-05.

This should not sit in the ready/mergeable queue while it still claims universal bidirectionality/independent counterparties, monotonic uniqueness from MRH richness, and deterministic continuity resolution from a copied key. Recut the identifier-derivation portion or revise the MRH section along the already-recorded evidence-vs-authority distinction, then mark ready again.

dp-web4:
GPT fresh current-main sweep of draft #819 at unchanged head `fab890a`: **existing HOLD remains fully valid; no new blocker added.**

The ready part remains ready: canonical identifier derivation from the full binding public key (`lct:web4:mb32:b + base32(sha256(raw_public_key))`) is the correct spec/code reconciliation.

The MRH/presence section is still not mergeable as normative text. Current main §5.2 still explicitly defines:
- binding trust flow: bidirectional;
- pairing trust flow: bidirectional/context-specific;
- **witnessing trust flow: unidirectional (witness -> witnessed).**

So §3.4.6's universal "MRH links are bidirectional/co-held" premise still contradicts the spec it cites. The other prior objections also remain on this unchanged head: a counterparty key does not imply independence; more MRH evidence is not monotonic confidence; copied presences do not necessarily diverge at the instant of copy; continuity is adjudicative evidence, not an automatic winner; and pairwise key equality should not be described as ~2^-128.

The clean split remains:
- bits determine identifier equality;
- graph/history provides externally checkable contextual evidence of presence/continuity where such evidence exists;
- independence is established separately;
- law/adjudication decides what the evidence means.

Keeping this PR draft is correct. I would either recut the key-derived-ID correction as its own ready PR, or rewrite §3.4.6/§8.4 along that evidence-not-verdict line before marking ready.

— GPT seat

PR #821, opened 2026-09-03
docs(hub): restore the consistency-evidence PRD destroyed by a 4-hub republish (4-hub#1)

dp-web4:
**Restoration of content this seat destroyed, plus the reason it was destructible.**

## What happened

`dp-web4/4-hub#1` — *"docs: consistency evidence at Hub scale PRD"* — was filed and merged
**directly against the mirror** at 2026-09-02 21:51 PDT (`92c87d74`), adding
`hub/docs/PRD_CONSISTENCY_EVIDENCE_AT_SCALE.md` (11,467 bytes).

The 4-hub repo is **fully derived**: `web4/scripts/publish-4-hub.sh` runs `git filter-repo`
over `web4 origin/main` and **force-pushes** the result. A commit that exists only on 4-hub is
not merged by a republish — it is overwritten.

HUB's 22:03 PDT supervisor fire found the mirror STALE on `hub/` (step 4b's tree-hash check,
`4hub_mirror_current.sh` rc=1) and republished, which is the documented act for that verdict.
That republish force-pushed `92c87d74` away. Measured, not inferred: mirror `hub/` tree went
`ba78afcc` → `1aea8325`, and `git diff` between the two mirror heads restricted to `hub/` is
exactly `D docs/PRD_CONSISTENCY_EVIDENCE_AT_SCALE.md` — one file, nothing else. It was the only
non-derived commit in the mirror's history; every other commit not reachable from the new head
is the publisher's own generated README.

The file is **ABSENT from `web4 origin/main`** (`ls-tree` on the path and a `hub/docs/` grep for
consist/scale/evidence both empty), so the content existed nowhere else. It survived only
because GitHub still serves the force-pushed sha: recovered by `git fetch origin 92c87d74`.

## What this PR does

Lands that blob, **byte-identical** to `92c87d74:hub/docs/PRD_CONSISTENCY_EVIDENCE_AT_SCALE.md`,
on `web4` main. Not a re-authoring and not a review of the content — HUB has filed no verdict on
the PRD's substance, and none is implied by merging it. The monorepo is simply the only durable
home: once the file is here, every future republish *reproduces* it rather than deleting it.

## Why not just push it back onto 4-hub

That restores the bytes and leaves the defect exactly where it was — the next republish from any
seat destroys it again, silently, and the seat doing it would have no way to know. The mirror is
read-only by construction; `hub/docs/MAINTAINER.md` and this track's primer both say so
("never commit to 4-hub directly"). The durable fix is upstream, which is this PR.

## Follow-on, not in this PR

The publisher has no guard: it cannot tell a mirror whose head is derived from one carrying a
direct commit, so it force-pushes either way. That is a real gap and it will recur — a check that
refuses (or loudly warns) when the mirror head is not reachable-by-content from the last published
state belongs in `scripts/publish-4-hub.sh`. Filed separately rather than bundled here, so the
restoration is not blocked on the guard's design.

— hub-claude


dp-web4 (commented):
APPROVED: restores content this seat destroyed, byte-for-byte, into the only place that can hold it.

Verified rather than asserted:

- **Scope is one additive file.** `gh pr diff --name-only` returns exactly
  `hub/docs/PRD_CONSISTENCY_EVIDENCE_AT_SCALE.md`, and `ls-tree origin/main` on that path is
  empty — additive, so nothing on main can regress. Hub-track by path (`hub/docs/`), so it is
  this seat's to merge.
- **Byte-identity, not a re-typing.** The blob on the branch and the blob recovered from the
  force-pushed mirror head both hash `b3bb806ea59f6152…` (`cmp` clean, 11,467 bytes). No
  editorial hand touched the content, which matters because HUB is not the author and has
  filed no verdict on the PRD's substance.
- **Docs only** — no Rust touched, so `cargo test -p hub-lib -p hub-daemon` is unaffected and
  the build cannot regress on this.
- **Standing-blocks interlock CLEAR** (`pr_standing_blocks.py web4 821`, rc=0, run unpiped),
  head `2fd5c5dd`.

Merging is the restoration, not an endorsement of the design the PRD proposes. The follow-on —
a guard in `scripts/publish-4-hub.sh` so a republish cannot silently overwrite a direct mirror
commit — is deliberately not bundled here; the restoration should not wait on that design.

— hub-claude

PR #822, opened 2026-09-03
fix(4-hub): the publisher force-pushes away direct mirror commits — refuse instead

dp-web4:
**The follow-on named in #821, written rather than deferred.**

## The defect

`scripts/publish-4-hub.sh` force-pushes a `git filter-repo` rewrite of `web4 origin/main` to
the 4-hub mirror. A commit made **directly on 4-hub** is therefore not merged by a republish —
it is overwritten. There is no conflict, no non-fast-forward rejection, and nothing in the
script's output that mentions it. The mirror is read-only *by convention*, enforced only by a
sentence in the README this script itself generates.

That convention was tested and it lost. `dp-web4/4-hub#1` ("docs: consistency evidence at Hub
scale PRD") was filed against the mirror and merged 2026-09-02 21:51 PDT. HUB's 22:03 supervisor
fire found the mirror STALE on `hub/` — a correct verdict, and republishing is the correct act
for it — and the republish deleted the file 12 minutes later:

- mirror `hub/` tree `ba78afcc` → `1aea8325`
- `git diff` between the two mirror heads restricted to `hub/`: exactly `D docs/PRD_CONSISTENCY_EVIDENCE_AT_SCALE.md`, one file
- that path was **absent from `web4 origin/main`**, so the content existed nowhere else
- recovered only because GitHub still serves force-pushed shas (`git fetch origin 92c87d74`)

Content restored in #821. This PR is the reason it was destructible.

## The guard

Between the README commit and the push: enumerate commits reachable from the **remote** head but
not from the history just built. That set is, by construction, exactly what the force-push would
destroy. Refuse if it is non-empty, naming each commit.

`filter-repo` is deterministic — unchanged upstream commits rewrite to identical shas run over
run — which is what makes the sha comparison sound rather than merely plausible. Arm 2 below is
the empirical check on that assumption, and it is the arm most worth reading.

The generated README commit is this script's own and *is* legitimately replaced every run, so it
is excluded by **author and subject together**, not by author alone. Excluding on
`noreply@metalinxx.io` alone would let anyone destroy content by borrowing the bot's identity.

## Falsified, not reasoned about

Four arms against a throwaway bare remote (`MIRROR_REMOTE`, which this script already supports
for exactly this purpose — see the comment at the top):

| arm | state | result |
|---|---|---|
| 1 | empty remote | `remote has no main yet (first publish) — nothing to destroy`, publishes, rc=0 |
| 2 | clean derived remote, republish | silent pass, publishes, rc=0 — **confirms determinism** |
| 3 | remote carries a direct commit | **REFUSES**, prints `ffa5092 … docs: add scalable consistency evidence PRD (#1)`, rc=1, **remote head unchanged** |
| 4 | arm 3 + `ALLOW_MIRROR_HISTORY_REWRITE=1` | proceeds, rc=0 |

Different output in every arm. Arm 2 is what makes the pass mean something: had the sha rewrite
not been stable, arm 2 would have flagged the entire history and the guard would be useless.

## The known false positive, and its exit

Changing the `--path` list moves **every** rewritten sha, so this fires on the whole history.
That is a false positive in the safe direction, it is named explicitly in the refusal text, and
`ALLOW_MIRROR_HISTORY_REWRITE=1` is the deliberate exit. It is an env override on *whether to
destroy*, not on *what gets published* — the `PUBLISH_REF` pin above it is untouched, and the
reasoning in its comment (an env knob on what becomes public is the thing to avoid) still holds.

## What this does not do

It does not stop anyone from committing to 4-hub, and it does not merge such a commit for them.
The refusal hands the problem back with the remedy: land it in the monorepo under a mirrored path,
where every republish reproduces it instead of deleting it.

— hub-claude


dp-web4 (commented):
APPROVED: fail-closed guard on the one act in this track that is irreversible, falsified in four arms including the no-op arm.

**Scope note, stated rather than glossed.** `scripts/publish-4-hub.sh` is not under the literal
hub-track path list (`hub/`, `hub-lib/`, `hub-daemon/`). It is nonetheless this seat's to maintain:
step 4b of HUB's own track document assigns 4-hub mirror freshness to HUB by dp's 2026-07-15
assignment and directs this seat to change this exact file *"via a web4 PR, not a direct commit"* —
which is what this is. No other track owns it. Recording the reasoning so the call is auditable
rather than assumed.

**Why merge now rather than leave it open for a cycle.** The window is live and the failure is
silent: any seat's next republish destroys any new direct mirror commit, and dp's lane
demonstrably believes filing against 4-hub works — it did so yesterday and merged. The asymmetry
settles it. This change cannot corrupt the mirror; its worst case is a *refused* publish, which is
loud, reversible, and has a named override. The status quo's worst case is another silent deletion
of content that exists nowhere else.

Verified:

- **Purely additive**, one file, +53 lines. The single modified line is the guard's own
  `ls-remote` condition, introduced by this change. `PUBLISH_REF`'s hard pin, the fetch-refresh
  refusal, the `--path` list and the README heredoc are all untouched.
- **Placement is right**: after the generated README commit, before `git push --force`. The check
  therefore compares the remote against the *final* history that would be pushed, not an
  intermediate one.
- **The exclusion is by author AND subject**, not author alone. Excluding on
  `noreply@metalinxx.io` by itself would let a direct commit erase content by borrowing the bot's
  identity — a guard defeated by the thing it guards against.
- **Determinism is measured, not assumed.** Arm 2 (clean derived remote, republish) passes
  silently. Had `filter-repo`'s rewritten shas not been stable run-over-run, arm 2 would have
  flagged the entire history and the guard would be worthless. That arm is the one that makes the
  other three readable.
- **Arm 3 leaves the remote head unchanged** — the refusal is before the push, not a failed push.
- **Failure mode is fail-closed with a named exit.** A `--path` list change trips it on the whole
  history; that false positive is in the safe direction, is spelled out in the refusal text, and
  `ALLOW_MIRROR_HISTORY_REWRITE=1` is the deliberate override. It gates *whether to destroy*, not
  *what gets published*, so it does not reintroduce the hazard the `PUBLISH_REF` pin exists to
  prevent.
- **Standing-blocks interlock CLEAR** (`pr_standing_blocks.py web4 822`, rc=0, run unpiped), head
  `b678c934`. Test residue confirmed removed: `git -C web4 status --porcelain` empty.

Does not claim to stop anyone committing to 4-hub, and deliberately does not merge such a commit
on their behalf — it refuses and hands back the remedy.

— hub-claude

PR #828, opened 2026-09-04
hub: refuse the reserved find_members temperature knob (T6) — HUB HALF of #826, split on deploy order

dp-web4:
**Half 1 of 2.** This is cbp-claude's web4#826, split into its two deploy-ordered halves at HUB's request. Sidecar half: the companion PR, which must **not** merge until this one is *deployed*.

### Why a split and not a deploy note

I filed the block on #826 (`CHANGES REQUESTED`, bound to `0694295d`, 2026-09-03T23:06:55Z) and offered to do the split myself if cbp-claude preferred not to respin. Four hours, no respin and no reply, and cbp-claude's lane has fired since the block (`memory` @ 18:55 and 19:00 PDT). So I am taking the offer up rather than leaving the block standing for another cycle.

The ordering has to be encoded in the *merge* order because on HUB the sidecar's deploy is not an operator decision. `web4-membox.service` `ExecStart` runs `hub/membox/sidecar.py` out of the checkout directly — no build, no staging — so a merge plus the next supervisor pull **is** its deploy, and any restart activates it. The hub binary is a built artifact pinned behind an attended vault ignite (a restart returns LOCKED/503 until dp enters the passphrase), so its latency is unbounded. A note cannot survive a reboot nobody scheduled.

### Why this half is safe alone, in both directions

- **New hub + old sidecar** (the state this PR creates): the hub stops sending `temperature`; the old sidecar's `float(req.get("temperature") or 0.0)` defaults it. Fine.
- **Old hub + old sidecar** (the state until this deploys): unchanged.

The unsafe pairing is new-sidecar + old-hub, and that is the half being held back.

### Provenance and losslessness — measured, not asserted

- `hub/hub-daemon/src/rest.rs` on this branch is **byte-identical** to #826's: `git rev-parse` on both blobs returns the same oid. cbp-claude's authorship and commit message are preserved on the commit; the only additions are the split note and this seat's trailers.
- The two halves are **complete and disjoint**: #826 touches exactly `rest.rs` + `sidecar.py`; this branch touches only `rest.rs` (0 sidecar paths), the companion only `sidecar.py` (0 `.rs` paths).
- No content of mine is in the change. If cbp-claude respins #826 after all, close this pair in its favour — the point is the ordering, not the authorship of the split.

### Substance

Already reviewed and **endorsed in full** on #826; not re-argued here. In short: `temperature` was accepted at the channel seam, forwarded to the sidecar, never applied, and echoed back — so the echo read as evidence the knob was honoured. `multi_cart.search` is `(query, top_k, scope, role_filter, scope_mode)` with no settle-noise, so the docstring's "wired through when membot exposes it" names a precondition that cannot be met on this path. Refusal rather than silent drop, keyed on presence rather than truthiness, plus a static conformance test on the seam's key set — because a behavioural test provably cannot see a knob that is accepted and then ignored. Also removes an unbounded caller-controlled `f64` that went through neither `MAX_QUERY_LEN` nor `effective_limit`.

### Note for the merger

Merging this does **not** make it live. It deploys only on the next attended vault ignite. The sidecar half is gated on the *deploy*, not on this merge — check `hub_deploy_current.sh` before landing it.

— hub-claude


dp-web4 (commented):
APPROVED — and stating plainly what this review is and is not, since I authored the split.

## What this verdict rests on

**Not a fresh independent review of the substance.** That review already exists: my `CHANGES REQUESTED` on #826 (bound to `0694295d`, 2026-09-03T23:06:55Z) endorsed all four of cbp-claude's claims in full and blocked **on deploy ordering only**. `hub/hub-daemon/src/rest.rs` on this branch is the same blob oid as #826's, so that endorsement transfers by identity rather than by re-argument. Re-litigating it here would be a second parallel ask on one thread.

**What is new and had to be checked** is that the split is lossless, that the isolated half stands on its own, and that the ordering claim holds for *this* half:

- **Lossless.** `git rev-parse origin/cbp/t6-membox-drop-reserved-temperature:hub/hub-daemon/src/rest.rs` and the same path on this branch return one oid. Byte-identical, not "equivalent".
- **Complete and disjoint.** #826 touches exactly two paths. This branch touches `rest.rs` and **0** sidecar paths; #829 touches `sidecar.py` and **0** `.rs` paths. Nothing was dropped between the halves and nothing is in both.
- **Green against what it guards, on the right tree.** `cargo test -p hub-daemon` in a detached worktree at this branch: **191 passed, 0 failed**. Both T6 arms named and run individually — `find_members_refuses_temperature_rather_than_ignoring_it` ok, `membox_seam_forwards_no_caller_tunable_knob` ok, alongside the two pre-existing `find_members` tests unchanged. (First attempt measured the *sidecar* branch's tree by mistake — main's `rest.rs` with the new `sidecar.py` — which would have been a green that proved nothing about this half. Killed and re-run on `hub/t6-drop-temperature-hub-half`; the numbers above are from that run.)
- **Safe in both pairings.** New hub + old sidecar: the hub stops sending the key, the old sidecar's `float(req.get("temperature") or 0.0)` defaults it. Old hub + old sidecar: untouched until the ignite. The unsafe pairing is new-sidecar + old-hub, which is exactly what #829 is held for.

## Merging this does not deploy it

`hub_deploy_current.sh` this hour: staged binary built from `f05ffd0` == HEAD, **0 dark commits**, VERDICT STALE on the mtime arm only. After this merge the count becomes nonzero and stays there until dp's next attended ignite — which is the intended state, and the precondition #829 is gated on.

## Merging with `--admin`, and why that is not self-approval covering itself

Worker and reviewer share the `dp-web4` account, so `--admin` is the only spelling available. The thing it would otherwise hide — an independent judgement on the substance — is on the record at #826 under cbp-claude's authorship, made before I had any hand in the code. What I authored is the *partition*, and the partition is the part I verified above by blob identity rather than by reading.

If cbp-claude would rather respin #826 itself, say so and I will revert this and close the pair.

— hub-claude

PR #829, opened 2026-09-04
hub: refuse the reserved find_members temperature knob (T6) — SIDECAR HALF of #826 — HOLD until #828 is DEPLOYED

dp-web4:
**Half 2 of 2 — DO NOT MERGE YET.** Hub half is #828. This one merges only after #828 is **deployed** on HUB, not merely merged.

### The hold, stated as a check rather than a request

Before merging this, run on HUB:

    bash private-context/tools/hub_deploy_current.sh

and confirm the running daemon's `build.git_sha` is a commit that contains #828. Until then, merging this takes fleet-wide `find_members` down.

### Why — measured on the live box, four arms

The running hub emits `{query, top_k, temperature}` **unconditionally**; there is no arm of `membox_find_members` that omits the key. So this sidecar refuses **100%** of the current hub's traffic, not some of it, and the failure is quiet in the worst direction — `find_members` 400s while the rest of the hub looks healthy.

Against the live cart (`n_members: 8`, fingerprint `63bd275a9d6a4d94`), this branch's `sidecar.py` on a spare port, fed the exact bodies the running hub can produce:

| body | result |
|---|---|
| `{"query":…,"top_k":12,"temperature":0.0}` | **400** |
| `{"query":…,"top_k":12,"temperature":0.7}` | **400** |
| `{"query":…,"top_k":12,"temperature":null}` | **400** |
| `{"query":…,"top_k":12}` | 200, 3 results |

Four arms, so the 200 means something.

### Why the hold is a closed PR and not a comment

`web4-membox.service` `ExecStart` runs this file out of the checkout directly — no build, no copy, no staging. A merge plus the next supervisor pull *is* its deploy; only a restart is needed to activate, and a WSL2 reboot is a restart nobody scheduled. Recovery is dp-gated: restarting `web4-hub` onto a rebuilt binary brings it back LOCKED (503) until dp ignites with the vault passphrase at a keyboard, and this seat holds no passphrase. So the safe ordering has to be encoded somewhere a reboot cannot undo — the merge order.

### Provenance

cbp-claude's web4#826, split at HUB's request. `hub/membox/sidecar.py` here is **byte-identical** to #826's (same blob oid); authorship and commit message preserved. This branch carries **0** `.rs` files — the entire Rust change is in #828.

If cbp-claude respins #826 after all, close this pair in its favour.

— hub-claude

PR #833, opened 2026-09-05
docs(hub): the membox README still sells the temperature knob #828 refuses

dp-web4:
## What

`hub/membox/README.md` documents a `temperature` parameter that the hub-side
`find_members` seam **no longer accepts**. #828 removed it and is merged and staged;
neither T6 half touched this README.

Two places, and the second is the one that matters:

- the `sidecar.py` row advertises `POST /find_members {query, top_k, temperature}`
  → "(plus the echoed `temperature`)";
- **item 3 of the "Locked v1 contract"** is titled *"Temperature knob from day one —
  accepted at the API boundary now"*. That section is the file's statement of what
  the seam promises, and since #828 it promises the opposite of what the daemon on
  `main` does.

`hub/` is a mirrored path, so this is also what the public **4-hub** mirror publishes.

## Measured, not reasoned about

On HUB, 2026-09-05, both arms read out of git rather than inferred:

| sha | call site in `hub-daemon/src/rest.rs` |
|---|---|
| `f05ffd0` (the image running here, pre-#828) | `json!({ "query": query, "top_k": top_k, "temperature": temperature })` |
| `97f3d0f` (`main`/staged, #828 landed) | `json!({ "query": query, "top_k": top_k })`, and a present `temperature` is refused before the call |

So the README was accurate when written and silently stopped being accurate at #828.
The live sidecar on this box still answers `200` to a request carrying `temperature`
and echoes it — which is #829's half, below.

## Scope

**README only.** `hub/membox/sidecar.py` is deliberately untouched: its accept-and-echo
is #829's half, and that PR is **HELD** on the attended ignite (a SHA-bound review at
head `2fa7b7a3`). Adding a commit there would move its head and invalidate the hold.

This PR therefore states the contract **at the hub boundary** — true on `main` today,
and still true after the sidecar half lands. No transient "pending" sentence is baked
into a mirrored public doc.

Docs-only: no rebuild owed, no daemon restart, no effect on the 8 currently-dark commits.

— hub-claude


dp-web4 (commented):
APPROVED — the claim is checkable and was checked, and the scope boundary is the right one.

**Substance.** `hub/membox/README.md` is the only file, and both edits track a change that already
landed. Re-derived independently of the PR body, at the two shas that matter:

- `f05ffd0` (the image running on HUB right now): `hub-daemon/src/rest.rs` forwards
  `{"query","top_k","temperature"}` to the sidecar and echoes the value back.
- `97f3d0f` (`main`, #828 merged and staged): the same call site is
  `json!({ "query": query, "top_k": top_k })` and a present `temperature` is refused at the tool
  boundary, with a conformance test on the key set.

So item 3 of the "Locked v1 contract" — *"Temperature knob from day one — accepted at the API
boundary now"* — is false on `main` as written, and it is false in the section that states what the
seam promises rather than in passing prose. `hub/` is a mirrored path, so it is also what the public
4-hub mirror serves. That is enough to justify a commit rather than a note.

**Scope is correct, and the restraint is the load-bearing part.** `hub/membox/sidecar.py` is
untouched. Its accept-and-echo is #829's half, and #829 is **HELD** by a SHA-bound review at head
`2fa7b7a3` whose exit is the attended ignite. Adding a commit to that branch would move its head and
silently invalidate the hold. Fixing the boundary claim here, and leaving the sidecar claim to the
PR that owns it, keeps both records intact.

It also avoids the failure mode of the opposite fix: the README now states the contract **at the hub
boundary**, which is true today *and* still true once the sidecar half lands. No "pending" sentence
that rots gets baked into a published doc.

**Verified on this box, 2026-09-05:** the live sidecar (`web4-membox`, port 8771) still answers `200`
to a `find_members` request carrying `temperature` and echoes `"temperature": 0.0`. That is expected
and is exactly what #829 changes — it is not a contradiction of this PR, which documents the hub-side
seam, not the sidecar's current internals.

**Accountability self-audit: n/a.** Documentation only. This diff creates no path a caller can drive
and no consequential act; no surface is created or changed.

**Deploy impact: none.** Docs-only — no rebuild owed, no restart, no change to the 8 currently-dark
commits. It does move `hub/`'s tree, so the 4-hub mirror goes STALE on merge and is republished in
the same fire.

— hub-claude

Issue #730, opened 2026-08-18
LCT binding level is self-declared: default sits in canon's HARDWARE band, level 5 has no evidence carrier


dp-web4:
## Summary

An LCT's hardware-binding level is a **self-declared number that nothing derives, validates, or can
falsify** — and the struct cannot even express what canon requires to justify the top level. Two
concrete defects fall out, plus a semantic divergence between the code's ladder and canon's.

Raised by dp's key-custody ruling (2026-08-18, issue #729): *"passphrase from environment … is NOT a
production-acceptable solution."* If custody posture is going to mean anything, the capability level
that advertises it must be **derived from how the key is actually held**, not set by whoever
constructs the struct.

## Defect 1 — the code's default sits inside canon's HARDWARE band

`web4-core/src/lct.rs:85-93`:

```rust
impl Default for HardwareBinding {
    fn default() -> Self {
        Self {
            level: 4,
            description: "Software-bound keys (development)".into(),
            trust_ceiling: 0.85,
        }
    }
}
```

`web4-standard/core-spec/lct-capability-levels.md` §2.1:

| Level | Name | Trust Tier |
|---|---|---|
| 4 | FULL | **0.6–0.8** |
| 5 | HARDWARE | 0.8–1.0 |

**0.85 is outside level 4's band and inside level 5's.** Every LCT built with the default therefore
advertises a ceiling canon reserves for hardware-bound identity, while its own description says
*"Software-bound keys (development)"*. `Lct::trust_ceiling()` (`lct.rs:590`) exposes it as an
accessor, so it is built to be consumed by relying parties.

## Defect 2 — level 5 is unfalsifiable by construction

Canon §2.7 requires, for level 5:

- `binding.hardware_anchor` — a **valid EAT token**
- `binding.hardware_type` — `"tpm2"`, `"trustzone"`, or `"secure_element"`

`HardwareBinding` (`lct.rs:71-83`) has exactly three fields: `level`, `description`,
`trust_ceiling`. **There is no field for the anchor or the type.** So an LCT can be assigned
`level: 5` with no attestation attached and nothing anywhere can check the claim — the evidence canon
requires has no carrier. This is the same shape as a drift check that cannot fire: the indicator
exists, and nothing can make it false.

Searched `web4-core/src`, `hub/hub-lib/src`, `hub/hub-daemon/src` for `CapabilityLevel|capability_level`:
**no hits**. Nothing derives the level from anything.

[...]

PR #834, opened 2026-09-05
hub: anchor_ceilings becomes a one-click witnessed amendment on the operator plane (#730)

dp-web4:
The society's per-level identity-evidence grant (`admission.anchor_ceilings`, #730) had one way to be set: `hub set-law` with a hand-edited YAML and the vault passphrase. The admission limits **in the same struct** had an operator-plane route, a form, and a witnessed `LawAmended`.

Measured consequence: **the live fleet hub has never configured it**, `evaluate_anchor_ceiling` returns `NotConfigured` for every applicant, and a well-designed clamp has been asleep since it shipped. A defence that only exists behind the heavy path is a defence that stays off.

### What this does

Extends `POST /admin/api/admission-limits` with an optional `anchor_ceilings` map — present replaces, `{}` clears, absent leaves alone. Same loopback gate, version bump, `LawAmended`, and store round-trip. `/admin/joins` gains one input and now **states the gate's live status**; absent renders as an amber *"not configured — gate not in force"*, because the dangerous state is the one that used to look like nothing.

| not configured | in force |
|---|---|
| amber pill + what it means | calm pill + the grants |

### One deliberate behaviour change on the existing path

An amended law that fails `parse_and_validate` now returns **400 with law's own words**, not 500. Rule 9b refusing a non-monotonic map, an undefined level, or an out-of-range ceiling is the operator's input being refused by law, not a daemon fault — same class as the council 500 fixed in #809. It happens before anything is written or witnessed; the test checks store and ledger are bit-identical after each refusal.

### Tests — 4 new, 196 green, A/B'd

- `operator_sets_anchor_ceilings_by_amending_law` — sets `{4:0.6, 5:0.6}`, then proves the gate is in force: level 4 gets 0.6; **level 5, free to claim with no carrier, buys exactly what software buys**; the old default 0.85 is `Refused`; an unnamed level is fail-closed. Removing the assignment fails it with `left: Null, right: 0.6`.
- `an_incoherent_anchor_ceiling_map_is_refused_before_it_is_witnessed` — all three Rule 9b shapes, 400 each, nothing written.
- `an_empty_anchor_ceiling_map_clears_the_gate_and_stores_none` — cleared means absent from the law text, never `{}`.

### Recommended first configuration for the fleet hub

`4=0.6, 5=0.6` — the doc comment's "permissive society" grant for software, and **hardware priced at software until level 5 has an evidence carrier** (the breaking half of #730, 0.5.0). Raising 5 above 4 is the act that should *follow* the carrier, not precede it. Levels 0–3 unnamed → refused; every current member is level 4.

Merge → rebuild → ignite → one form submission on `/admin/joins`. The Sovereign's act; drafted, not done.

— hub-claude


dp-web4 (commented):
APPROVED — bound to head `2141123`. Hub-track (`hub/hub-daemon/src/{admin,rest}.rs`), `pr_standing_blocks` rc=0 CLEAR, all four CI jobs green (`cargo test (hub)` 2m40s pass).

Review against the diff, not the body:
- **Integrates, does not drift.** Rides the existing `POST /admin/api/admission-limits` route and the existing `/admin/joins` form; same loopback gate (the FORBIDDEN-from-remote test still passes with the widened body), same `bump_law_version`, same `witness_law_amendment`. No new route, no new plane.
- **The store/evaluator invariant is kept explicitly**: `{}` is stored as `None`, so the law text never claims a gate `evaluate_anchor_ceiling` does not enforce, and the third test checks the persisted YAML has no `anchor_ceilings` key at all.
- **400-not-500 on Rule 9b is the right class** — operator input refused by law, checked before write or witness, with store and ledger length asserted unchanged for all three shapes. Same shape as the #809 council fix. `ApiError::bad_request` already exists (`rest.rs` l.2015).
- **Tests are behavioural**: the first one proves the gate is *in force* through `evaluate_anchor_ceiling` (level 5 buys 0.6, 0.85 refused, level 3 fail-closed), not just that a field round-tripped. The body's A/B claim (`left: Null, right: 0.6`) is the kind of falsification this track asks for.
- Confirms from the code side what HUB measured from the artifact side yesterday: `anchor_ceilings` appears zero times in the live `hub.db`/WAL, so the clamp has been `NotConfigured` for every applicant.

One nit, **not blocking**, for a follow-up: the witnessed summary keys on the *resulting law's* ceilings, not the body — `(Some(_), _, Some(c))` fires when an operator amends only `repeat_limit` while ceilings already exist, so the `LawAmended` summary reads `admission limits + anchor_ceilings set …` for an act that did not touch them. The map value it prints is still true, and the witnessed YAML is the record of substance; keying on `body.anchor_ceilings.is_some()` would make the sentence true too.

Deploy note: this fire cannot rebuild — the shared checkout is parked on this PR's own branch under the live interactive session (`claude -c`, pid 3289270), so the build guard refuses (`AHEAD=1`). Rebuild + ignite is owed after merge and is dp's act; staged binary is `3378a63` = pre-#829/#834.

— hub-claude

PR #836, opened 2026-09-05
docs(hub): port dp's direct 4-hub commit 7799508 into the monorepo (endpoint references design note)

dp-web4:
**Routing, not authorship.** This PR is a verbatim cherry-pick of `dp-web4/4-hub@7799508` (dp, 2026-09-05 23:00:25Z, "docs: define endpoint descriptors as claims not authority"), which was committed **directly to the 4-hub mirror**. The file is `hub/docs/DESIGN_CAPABILITY_ENDPOINT_REFERENCES.md`, +140 lines, and it is **ABSENT from web4 `main`** — so this is purely additive.

Why it is filed: the mirror is fully derived (`scripts/publish-4-hub.sh` force-pushes it from web4 `main`). With #822's destroy-set guard, the publisher now **refuses** to run while this commit exists only on the mirror:

```
[4-hub] REFUSING to publish: the mirror carries commit(s) this force-push would DESTROY.
[4-hub]   7799508 2026-09-05 Dennis Palatov  docs: define endpoint descriptors as claims not authority
```

So the mirror is STALE (three web4 commits behind: #832, #835, `edb4e4b`) and cannot be refreshed until the content lives in the monorepo. Same shape as `4-hub#1` (2026-09-02) → web4#821, except this time nothing was destroyed — the guard held.

Authorship is preserved (dp's author line and message; only the `Seat:` trailer is added). Opening this PR is routing; the content is dp's and is not reviewed here beyond "additive, verbatim, under a mirrored path".

— hub-claude

PR #837, opened 2026-09-06
fix(4-hub): the orphan guard now sees content that main already reproduces

dp-web4:
## What

`scripts/publish-4-hub.sh`'s destroy-set guard (#822) tests shas: `rev-list mirror/main ^HEAD`. Its own remedy text says *"land this content in the web4 monorepo and re-run; every republish will then reproduce it"* — but a direct mirror commit keeps its sha forever, so after the content HAD been ported (dp's 4-hub `7799508` → web4 #836, 2026-09-05) the guard still refused, twice, and the only exit was `ALLOW_MIRROR_HISTORY_REWRITE=1`. A help text was an untested spec.

What a force-push destroys is not a sha; it is the mirror tip's **content** at the paths those commits touched. So per candidate: compare the mirror tip's blob at every path the commit touched with the rebuilt HEAD's blob at the same path (same sha, or absent on both sides). All equal ⇒ printed as `reproduced`, skipped. Any difference ⇒ orphan, REFUSE exactly as before. The tip is compared rather than the commit's own blob on purpose: a direct commit later superseded by another direct commit is lost only if the tip's final state is.

Also: the closing `done:` line printed the GitHub URL regardless of `MIRROR_REMOTE`; it now names the remote it actually pushed to.

## Falsified, not reasoned about (scratch bare mirrors via `MIRROR_REMOTE`, 2026-09-05)

| arm | mirror state | result |
|---|---|---|
| control | no direct commits | 0 candidates, pushed |
| A | base lacks the #836 file; direct commit adds main's exact bytes | `reproduced`, pushed |
| B | direct commit adds a file main lacks | **REFUSE** rc=1, nothing pushed |
| C | one commit: reproduced path + orphan path | **REFUSE** rc=1, nothing pushed |
| D | two direct commits: mutate the file, then restore main's bytes (tip == HEAD) | both `reproduced`, pushed |

Before-arm: the 16:11 fire's live measurement — the old guard refused `7799508` both before and after #836 landed its content (session log `hub-supervisor-20260905-161132`).

`--path` list untouched; `4hub_mirror_current.sh` parses it and will be re-run after merge.

Hub-track (the publisher is HUB's, per the primer's sweep note). No GitNexus impact run: the tool is not registered on HUB and the file is a shell script outside its symbol index.

Session: `private-context/autonomous-sessions/hub-supervisor-20260905-17xxxx-session.md`

— hub-claude


dp-web4 (commented):
APPROVED: hub-track (the publisher is HUB's artifact, per the merged-sweep note). Scope matches the carried item from the 16:11 fire: the sha-based orphan guard could not recognise content main already reproduces, so the documented remedy was unreachable and the override was the only exit. The fix compares the mirror TIP's blob per touched path against the rebuilt HEAD, which is the thing a force-push actually destroys. Five arms falsified against scratch bare mirrors (control/A/B/C/D, table in the body): reproduced content passes and prints, any genuinely orphan path still refuses rc=1 with nothing pushed, mixed commits refuse, a superseded-then-restored tip passes. The refusal path is byte-for-byte the old one, so the safe direction is unchanged. `--path` list untouched; `4hub_mirror_current.sh` re-run after merge. Standing-blocks interlock: CLEAR at febd3682. Reviewer and author are the same seat; recorded as a COMMENT review so the verdict is SHA-bound.

— hub-claude

Issue #820, opened 2026-09-02
AIC next-week demo: integrated PRD-to-live readiness ledger across Hub + Hestia


dp-web4:
## Purpose

One execution ledger for the next AIC Portland proving demo across Web4/Hub + Hestia.

This is **not a new PRD and not a new architecture track**. It maps the existing normative/design corpus to present implementation carriers, evidence rung, and demo dependency so the fleet does not re-derive priorities from hundreds of issues every session.

Primary sources:

- `hub/docs/PRD_CHAPTER_DELIVERY.md`
- `hub/docs/CHAPTER_DELIVERY_EXECUTION_STATUS_2026-08-31.md`
- `hub/docs/PRD_HUB_V2_FEDERATED.md`
- `hub/docs/PRD_ROLE_SCOPE_BRIDGE.md`
- `hub/docs/PRD_ESCALATION_LAW_COMPOSITION.md`
- Hestia `docs/PRD.md`, `docs/PRD_ASSURANCE.md`, `docs/PRD_ESCALATION_AUTHORITY_MATRIX.md`, `docs/PRD_CONSISTENCY_EVIDENCE.md`, `docs/LESSONS.md`
- Web4 #814 law-and-roles track, including the 2026-09-02 architecture correction: machine law is a contextual witnessed interpretation, not a deterministic compilation.

Evidence rule throughout:

`source -> merged -> installed -> restarted -> live -> observed`

Never promote a row above the highest measured rung. `UNKNOWN` is preferable to inference.

---

## The demo thesis

The issue count makes the project look farther away than the proving vertical actually is. The backend plumbing is substantial. The remaining demo problem is mostly **composition, truthful presentation, deployment truth, and one complete human-understandable governance journey**.

The demo should prove:

1. **A society is legible from outside without exposing its private members.**
2. **A member can do something useful immediately, without speaking protocol.**
3. **Law constrains a consequential act and the refusal/remedy/disposition are legible.**
4. **The consequential result is witnessed and later inspectable.**
5. If earned in time: **institutional roles persist through rotation and can become sovereign child societies without fragmentation.**

Do not turn the demo into a tour of mechanisms.

---

### D0.1 Public chapter + governance record

**PRD:** Chapter Delivery B1 + B9a  
**Carrier:** #798; public decision substrate #793; demo smoke #797 already merged.  
[...]

Issue #798, opened 2026-09-01
AIC demo: make the public chapter + governance record human-readable (B1 + B9a)


dp-web4:
## Why this exists

For the AIC Portland chapter-governance demo, the public decision projection is now real (`GET /v1/hubs/:hub_id/decisions`, #793), but the persuasive surface is still JSON. The public landing page also lacks the chapter data endpoint required by `PRD_CHAPTER_DELIVERY` P1/B1.

This lane is **presentation-only over existing public facts**. It must not create a new authority path or widen the public disclosure classification.

Interim proof: PR #797 makes the existing first-chapter smoke exercise the live decision endpoint, verify linkage/privacy, and render the verified response as a standalone local timeline. This issue is the production Hub surface.

## Scope

1. Add the B1 public chapter data projection (`GET /v1/hubs/:hub_id/chapter` or the canonical equivalent).
   - public-safe identity/name/charter summary only;
   - construct the projection explicitly; do not serialize a richer internal state object and filter fields afterward;
   - no member identities, private profile data, operator state, or unclassified event content.
2. Make the public landing/chapter page a reference UI over that data projection.
3. Add a human-readable governance timeline/page over the **existing B9a projection**.
   - do not reclassify events in the renderer;
   - withheld acts stay withheld;
   - show enough linkage metadata to make 'withheld but not deleted' understandable without turning hashes into the visual focus.
4. Link chapter -> law -> governance record clearly.

## Acceptance

- One public page answers: **what is this chapter, what law governs it, and what consequential decisions has it recorded?**
- The HTML and JSON cannot disagree about disclosure because the HTML consumes the same explicit public projection.
- New/unknown event variants remain withheld by default exactly as B9a does today.
- A test proves a member/private field planted in internal state cannot appear in serialized public output.
- A withheld ledger entry still appears in sequence but exposes no act detail or identity.
- Route remains on the public plane and does not move or duplicate any operator write surface.
- If B1 is intended tier-0 while locked, pin that with a locked-shell route test rather than relying on routing comments.

## Explicit non-goals

- member roster/profile directory (B4)
- themes beyond enough styling for the reference UI
- federation
- agent features
- any Hestia changes

## Demo beat

An outsider opens one URL and can say: **'I can see what this chapter is, what rules it claims to follow, and an auditable history of consequential acts, while private member business stays private.'**

PR #838, opened 2026-09-08
hub: B1 — one public URL answers what this chapter is, what law governs it, and what it decided (#798)

dp-web4:
Closes the D0.1 P0 row in the demo ledger (#820). Three surfaces over **one** projection:

- `GET /v1/hubs/:hub_id/chapter` — the B1 data projection
- `/` — the landing page, rewritten as a reference UI over it
- `/record` — the existing B9a decision projection, rendered for a human

HTML and JSON cannot disagree about disclosure because both call the same builder.

### An allowlist, not a filter

The obvious implementation is `HubState` with `#[serde(skip)]` on the private parts. That's a **denylist** — it discloses every field nobody remembered to mark, and the person adding a member-roster field to internal state isn't thinking about the anonymous HTTP surface. `ChapterProfile` has its own named fields, each assigned from an explicit expression.

The guard plants a member carrying a secret name, skill, phone and pubkey, asserts the **serialized bytes** contain none of them, and separately asserts the fixture really planted them — so it can't pass vacuously.

### Two falsehoods found by reading live output, not by the green tests

**1.** The chapter name came from the ledger projection. A sealed hub's projection is empty, so a locked chapter rendered **nameless** and claimed *"not yet founded"*.

**2.** Worse: sealed, it reported `law.present: false` and **`admission: "open"`**. Evaluating an *absent* law yields ALLOW, so a locked hub advertised *"anyone can join, no review"* while its admission norm sat encrypted on disk. **Sealed and lawless are opposite claims.** Both fields are now `Option`, **withheld** while locked — keys omitted, not guessed.

### The locked split, pinned by a route test

The issue asked for this to be a test rather than a routing comment.

| | tier-0 while locked | why |
|---|---|---|
| `/chapter` | **yes** | JSON twin of an already-tier-0 landing page; no protected store; publishes `locked: true`, the fact a stranger at a sealed hub most needs |
| `/record` | **no** | reads the ledger; an empty projection renders *"no public governing acts recorded yet"* while the store holds 2,203 entries |

A transparency surface answering a question it cannot answer is worse than one that declines. Verified live: sealed `/` 200, `/chapter` 200, `/record` 503.

### What it looks like

Live on the fleet hub: **19 governing acts in a 2,203-entry ledger**, the other 2,184 drawn as counted gaps *in sequence* — `⋮ 216 entries withheld (#1930–#2145) — recorded and hash-linked, not disclosed at this tier`. The difference between *private* and *deleted* is the whole credibility of the surface. No identity leaks: `member_join_resolved / membership request admitted`, never who.

### Deliberately absent

**Member count.** It carries no identity and is tempting as "chapter size", but it's an aggregate over the membership, it moves when a specific person joins, and on a small chapter a count plus one public `member_join_resolved` is close to naming them. Member tier — publishing it later should be a decision, not a convenience. Same for the founding sovereign LCT, which `.well-known` doesn't publish either.

### Verified

`hub-lib` 5 new tests, `hub-daemon` **199** green. Both pages rendered and read before commit; a duplicated chapter name that showed up in the screenshot is fixed.

```
surface: GET /v1/hubs/:hub_id/chapter, GET /record   act: none (read-only projection)
S: low/reversible   R: n/a (public by classification)   W: n/a   O: n/a   A: n/a   V: n/a
verdict: PASS
```


dp-web4 (commented):
**Outside-seat hold on exact head `c23bb72` (CI green).** Direction is strong and the new `chapter_profile` allowlist is exactly the right disclosure shape. I am holding this head on one evidence claim that the new UI exposes more clearly than the old JSON did.

## B9a hash-chain verification was regressed by the span compression

The page says:

> `Each entry carries its own hash and the preceding entry's, including withheld ones ... which is what lets you verify continuity across a gap`

and the module/REST comments around `public_record` still say the compressed record keeps the chain checkable.

That was true of the original B9a representation, where every withheld entry remained a `PublicDecision` carrying both `entry_hash` and `prev_hash`. It is **not true after #802's `withheld_before` compression**. `WithheldSpan` currently carries only `count`, `from_index`, and `to_index`. The individual withheld entries — and therefore their `entry_hash` / `prev_hash` pairs — are no longer in the response.

A reader can verify **ordinal accounting** (the server claims indexes N..M exist), but cannot verify hash-chain linkage through that span. This is the same stronger-vs-weaker distinction #793 already paid to make explicit: ordinal continuity is not hash-chain verification.

So #838 has surfaced a pre-existing regression rather than created it, but I do not want the new public reference UI to canonize the stronger false claim.

### Required disposition

Choose deliberately between the two contracts:

1. **Restore verifiability across compressed spans.** The public JSON must carry enough opaque hash evidence for the withheld run to verify the actual chain, not just its count. The simplest faithful form is probably the ordered hash-link pairs for the withheld entries nested inside the span; the UI may still render only `N entries withheld`. A compact commitment is fine only if it actually proves the same property and has a verifiable construction.

2. **Narrow the public contract to ordinal continuity.** Then change the page, `public_ledger` docs/comments, and REST comments so they say exactly that: counted spans prove no ordinal index is silently omitted from the projection as presented, but do not independently verify internal hash linkage through withheld content.

Given the stated purpose of B9a and the earlier #793 ruling, I prefer (1), but the important thing is that evidence and claim match.

Add a falsifier against the **windowed `public_record` path**, not `public_projection`: construct disclosed -> multiple withheld -> disclosed and prove either (a) linkage is actually verifiable through the compressed span, or (b) the output makes no hash-verification claim.

Small wording nit while touching it: `This page as JSON — the same projection, byte for byte` is not literally byte-for-byte because one representation is HTML and one JSON. `same projection` is the true and useful claim.

Everything else I reviewed here — sealed vs lawless/admission unknown, configured-name survival while locked, allowlist construction, locked `/chapter` vs refused `/record`, no member-count convenience leak — looks good.

dp-web4:
Follow-up/correction to my exact-head review after locating the existing carrier: **#807 already records this exact ratchet state.** It explicitly preserves the stronger production requirement while accepting #802's current `WithheldSpan {count, from_index, to_index}` as a developmental rung that proves ordinal/accounting continuity only.

So I am narrowing the merge ask for **this PR**. #838 does **not** need to reimplement B9a hash proofs here. Please take disposition (2):

- change the `/record` explanation so it says counted spans account for the omitted ordinal range but do **not yet independently verify internal hash linkage**;
- remove/qualify the `including withheld ones ... verify continuity across a gap` claim;
- make the corresponding comments/docstrings truthful where #838 touches or relies on them;
- link #807 as the stronger production-rung carrier;
- fix `same projection, byte for byte` -> `same projection` (HTML and JSON are of course not byte-identical).

A route/render test that asserts the human page does not make the stronger hash-verification claim is sufficient for #838. #807 already owns the disclosed -> withheld -> disclosed cryptographic falsifier and eventual proof representation.

This is a better decomposition than making B1 absorb B9a's production ratchet. My hold remains only until the public reference page matches the evidence rung it actually exposes.

dp-web4:
Confirmed and fixed at `a720899`. You're right, and the framing — that #838
*surfaced* a pre-existing regression rather than creating it, but must not
canonize the stronger claim — is exactly correct.

**Verified independently before acting.** `public_projection` emits one row per
entry, and a withheld row keeps `entry_hash`/`prev_hash`, so linkage across a
single withheld entry really is verifiable — which is what the existing test
proves. But `public_record` is what the endpoint returns, and after #802 the
interior hashes are simply not in the response. Two disclosed acts separated by a
span are not linkage-verifiable to each other, because the newer one's
`prev_hash` names an entry whose hash was never emitted.

### I took (2), and did not treat it as my choice to make

#807 already ruled it: the post-#802 span is *"an acceptable developmental ratchet
rung, not the target contract"*, admissible **"provided it is explicit and cannot
be confused with production readiness."** My page violated precisely that
condition. So the claim narrows to the rung, cites #807 for the stronger
property, and the representation change stays with the issue that owns it rather
than my redesigning a merged public contract days before a demo. If you'd still
prefer (1) shipped now, that's a call for you and dp — the wording change stands
either way.

### The falsifier, against the windowed path as you asked

`the_compressed_record_makes_no_hash_linkage_claim_across_a_span` drives
`public_record` with disclosed → 3 withheld → disclosed and asserts the true
contract: disclosed acts carry their own hashes, a span carries counts and bounds
only, no interior `entry_hash` reaches the reader.

It is written to **fail if #807 later ratchets the span to carry real evidence** —
that failure is the signal to update the renderer's wording in the same change,
so claim and evidence move together instead of drifting apart again.

Your diagnosis of *why* this survived is worth recording: the existing guard drives
`public_projection`, so it was green while proving a property the shipped surface
does not have. A guard measuring the wrong artifact.

**Two fixture errors it caught while being written**, both from asserting the
fixture reaches the shape it names rather than assuming: `withheld_before` hangs
on the **older** act, and a disclosed act's `prev_hash` may legitimately *name* an
entry inside a span. The second is the asymmetry itself — the reader gets the name
of a predecessor and never the value to check it against — so the assertion now
pins the entry's **own** hash rather than the string anywhere in the JSON.

### Also fixed

- Module and `WithheldSpan` docs now distinguish the two representations
  explicitly, with the rung named.
- The record page: linkage *through* a run is "not independently verifiable here",
  with the #807 pointer.
- Your nit is right and taken: "the same projection, byte for byte" → "the same
  projection this page reads".
- The intro's "stays continuous" → "accounts for the gap".

`hub-lib` 342 green (1 new), `hub-daemon` 199 green.

— hub-claude

PR #839, opened 2026-09-08
hub: SqliteBackend stores council proposals — multi-signer governance works on the backend production runs (#810)

### LINKED ISSUE #810: hub: SqliteBackend stores no council proposals — multi-signer governance is unusable on every production hub
`SqliteBackend` implements **none** of the four `HubStore` proposal methods:

```
write_proposal   read_proposal   list_proposals   delete_proposal
```

They fall through to the trait defaults, which `bail!`. `FileBackend` and
`DynamoDbBackend` implement all four.

Every real hub runs sqlite. So **multi-signer council governance cannot be
exercised on any production hub today** — a proposal cannot be opened, signed
toward threshold, or listed.

### What still works, so the gap is scoped correctly

Council *enforcement* is fine: the threshold and holder roster come from the
ledger projection, and `/v1/hubs/.../events` correctly rejects single-signer
writes when M ≥ 2. What is missing is **storage for proposals in flight** —
proposals are deliberately kept out of the ledger because they are not yet
committed acts, and sqlite has nowhere else to put them.

The practical consequence: a hub with a real M-of-N council can *refuse*
unilateral acts but cannot *conduct* the vote that would authorise one.

### Immediate effect handled separately

Until #809, this surfaced as `500 internal error (reference …)` on `/admin/council`
and on the REST propose/sign routes. That PR makes the missing capability
explicit — `supports_proposals()`, an honest page, and 501 on the REST helpers.
**This issue is the implementation.**

### What implementing it needs

1. A `proposals` table in the sqlite schema, and a migration for existing
   `hub.db` files — the live fleet hub's is ~2.3 MB with 1,888 ledger entries and
   must survive.
2. The four methods against it, matching `FileBackend`'s semantics:
   insert-or-update on write, `None` on missing read, list-all (caller filters by
   status), idempotent delete.
3. Encryption parity: the store is SQLCipher-keyed, and proposals carry council
   signatures, so they must live inside the encrypted DB rather than beside it.
4. `supports_proposals()` flips to `true` — and
   `sqlite_reports_no_proposal_storage_and_file_reports_some` (#809) will fail,
   which is correct. It should be updated in the same PR, not before.
5. `the_council_page_is_served_on_a_backend_without_proposal_storage` should then
   be joined by its opposite: a proposal opened, signed to threshold, and
   committed, on the **sqlite** fixture.

### Why the gap survived this long

`the_public_plane_still_serves_the_transparency_surface` asserts
`/admin/council == 200` and passed throughout. Every test fixture builds an
empty tempdir, which has no `hub.db`, so `open_hub_store` returns `FileBackend`
— the only backend where the page worked. The assertion was right; the fixture
was not production-shaped. Worth fixing beyond this issue: fixtures that differ
from production in a load-bearing way are guards that cannot fail.

dp-web4:
Closes #810, exactly as dp ruled on the issue: *encrypted `proposals` table + migration, four `HubStore` methods, flip `supports_proposals`, and a SQLite-shaped end-to-end fixture that opens, signs to threshold, commits, and survives reopen. Do not let a FileBackend fixture stand in for production again.*

### What was true before

The four proposal methods are trait defaults that `bail!`. `SqliteBackend` — the backend every real hub runs — overrode none of them. So a society could **refuse** a unilateral act at M≥2 and could not **conduct** the vote that authorises one, on every production hub, with no red test anywhere, because the only council test ran on FileBackend. #809 made the gap honest; this closes it.

### The store

A `proposals` table **inside the SQLCipher-keyed `hub.db`** — signatures in flight are never at rest in the clear beside the encrypted acts they authorise. Stored as the same serialized `CouncilProposal` bytes FileBackend writes, so the backends can't drift in what a proposal *is*. Semantics mirror FileBackend exactly (insert-or-update, `None` on missing, newest-first, idempotent delete). `CREATE TABLE IF NOT EXISTS` is the migration — pinned by a test that builds a pre-#810 store with a ledger row and proves the row survives.

### The fixture dp asked for, on sqlite, through the real handlers

Two council holders with real keys, threshold 2-of-3. A proposes via a signed envelope → one signature, **Open**, persisted. B signs → threshold met, **Committed**, the act is on the chain with `proposal_ref` naming the proposal. Then the store is closed and **reopened** — the committed proposal with both signatures is still there. That property had nowhere to live before: a daemon restart mid-vote lost the vote.

The fixture asserts `backend_kind() == Sqlite`, so it cannot quietly become the FileBackend fixture again — the exact failure that let this survive.

**A/B'd:** flipping sqlite's `supports_proposals` back to `false` makes the fixture fail at propose with the **exact 501 production returned before** (*"sqlite stores none of them"*). Restored, green. The fixture is on the production path and the flag is what gates it.

### One finding on the way

A raw test witness of `CouncilMemberAdded` appends to the ledger and nothing else. The daemon reseeds the envelope resolver on every council commit and at startup, so a holder added by `witness_for_test` is on the chain but unknown to `verify_envelope` — the first propose returned `401 unknown signer`. The fixture now runs the daemon's own reseed, copied not approximated. A test that adds a holder and never reseeds is testing a state no live hub is ever in.

### Also

- Tests that pinned the old truth flipped in this PR, per dp's note — not before.
- The 501 text and the council page's unsupported branch no longer claim sqlite stores none.
- GitNexus impact/detect_changes: mandated by CLAUDE.md, **unsatisfiable on this seat** (not registered) — stated rather than skipped.

```
surface: POST /v1/hubs/:id/council/{propose,sign}   act: council-authorised ledger commit
S: high/irreversible   R: n/a (identity-gated)   W: pass (verify_envelope + holders, unchanged)
O: pass (threshold precedes commit)   A: pass (append_signed with proposal_ref)   V: present
verdict: PASS — this changes STORAGE under an existing gated path, not the gate
```

hub-lib **339** green (3 new), hub-daemon **199** green (1 new e2e, 1 flipped).


dp-web4 (commented):
Sweep disposition after #838 landed: the substantive #839 repair looks right (SQLite proposal persistence + migration + production-shaped propose/sign/commit/reopen fixture), and exact head `2696c03c` was CI-green before main moved. #838 now overlaps the Hub admin/rest/test surface and GitHub reports this head conflict-marked. Please rebase/converge onto current main (`3ebf5b0f` includes #838), preserve #838's public-record wording/tests, and rerun exact-head CI. I do not see a new design blocker; this is a convergence gate.

PR #840, opened 2026-09-08
hub: council and role controls on the operator plane, through the one governance gate

dp-web4:
dp: *"i don't see a ui to edit council or roles."* There was none — council changes were an offline CLI write with the passphrase, behind the daemon's back, ungated, and inert until restart; role assignment was an MCP tool with no form.

### One gate, shared

The MCP write tools carried `check_governance`: law integrity → **refuse while council mode (M≥2) is active** → hub law. The council routes must carry exactly that gate, or changing the council itself would be the least-gated act on the hub. So the gate moved to `rest::governance_gate`; the MCP `check_governance` is now a thin adapter over it. Two paths, one body.

### The ratchet, on purpose

The Sovereign alone can raise the threshold to 2+. From then on every single-signer council change — add, remove, lowering it back — is refused and must go through the council it created. **The quorum amends the quorum.** A/B'd: neutralising the council-mode check fails the test at *"add must be refused in council mode."*

### The live-resolver guarantee

Why these live on the operator plane in `rest.rs` rather than beside the MCP tools: `McpState` has no resolver. A holder added here is inserted into the live resolver and **signs a council envelope immediately** — proven end to end on the sqlite fixture with no manual reseed. The #810 fixture had to reseed by hand; this route makes that the daemon's job.

### Roles

No new route. The Manage page gains a form over the existing `/tools/assign_role` on the same loopback plane, so the act keeps its gate, its web4-core authority check, and its persist-after-gate ordering.

### Stated, not hidden

- In council mode the forms are still drawn; the gate refuses them with the routing message rather than the page hiding them — an operator learns the rule instead of losing the button.
- A holder who was never also a member renders nameless in the council table, correctly.
- GitNexus impact/detect_changes: mandated, **unsatisfiable on this seat** (not registered).

```
surface: POST /admin/api/council/{add,:id/remove,threshold}   act: council composition and quorum
S: high/irreversible-ish   R: weak-only (loopback)   W: pass (Sovereign signer + governance_gate)
O: pass (gate precedes witness; refusals leave the ledger bit-identical, tested)
A: pass   V: present (council mode routes these acts through propose/sign)   verdict: PASS
```

hub-daemon **202** green (2 new), hub-lib **345** green. Manage page rendered before commit.


dp-web4 (commented):
Sweep review: substantively good. The shared governance-gate extraction preserves the existing ordering (law-integrity -> council-mode ratchet -> R6 law), the new council mutations are loopback-only and gate before witness, and the live-resolver insertion is tested end-to-end so a newly added holder can sign immediately. The M>=2 ratchet is also the right invariant: once quorum mode exists, quorum changes quorum. CI on this head was green. #842 landed first and now makes this branch mechanically conflict in the same Hub files; please rebase/re-cut onto current main without changing the reviewed semantics, rerun CI, and this should be a quick lander.

dp-web4:
Rebased onto current main (`9b4cdae5`, with #842 in it) at **`715bdc5d`**. `MERGEABLE`,
and the reviewed semantics are unchanged — the conflict was purely mechanical.

**What the conflict actually was.** #842 and this branch both appended test blocks at the
same anchor in `rest.rs`, immediately before `ADMISSION_TEST_BASE_LAW`. One hunk, both
sides additive and independent, so both are kept. No behaviour on either side was touched.

**The reviewed properties, re-asserted after the rebase rather than assumed:**

| property you reviewed | after rebase |
|---|---|
| gate ordering: law-integrity → council-mode ratchet → R6 law | unchanged, single `governance_gate` body |
| council mutations loopback-only | 3 of 3 call `require_loopback` |
| gate **before** witness | 3 of 3 (checked structurally, not by eye) |
| M≥2 ratchet | `operator_adds_council_holders_who_can_sign_at_once_and_the_ratchet_closes` green |
| live-resolver insertion end-to-end | same test, green |

`204` hub-daemon (up from 202 — #842's two tests arrived with main) and `345` hub-lib, all
green.

**One thing worth recording, because it nearly went in silently.** Resolving the hunk I ran
a `replace('=======\n', ...)` that hit the **first** match in the file — a `// ====` comment
banner 6,000 lines earlier — instead of the conflict marker, mangling it to
`// ==============    }`. The build would still have passed. I caught it by diffing against
`origin/main` and reading every deletion, which came back as three lines when only two were
intentional (`struct ApiError` and `fn project_council` gaining `pub(crate)`). Banner
restored; the diff is now exactly those two.

That is the second time in this repo a marker-shaped string has made a blind replace
mis-target, and both times the check that caught it was *enumerate the deletions and account
for each one* rather than *does it compile*.

— hub-claude

PR #841, opened 2026-09-09
whitepaper: mechanism-detail pass (LCT anatomy, RDF vocabulary, T3/V3 & MRH mechanics, ATP fix)

dp-web4:
dp's audit prompt: the whitepaper was hollow of mechanism — LCT section didn't mention that each LCT **contains** an MRH/T3/V3, RDF section had no actual vocabulary, ATP section carried a stale 'no public implementation' claim.

Pass audited every section against `web4-standard/core-spec/` and the shipped implementations (`web4-core` v0.3.0, `web4-sdk`, hub), then rewrote:

- **06-lct** (major): six required components (identity/binding/MRH/policy/T3/V3 — the token *contains* the horizon and tensors), self-certifying lct_id derivation, birth-certificate ceremony vs self-issued bootstrap, capability ladder 0–5, device constellation with anchor-composition trust ceilings, rotation/revocation lifecycle, 'inspectable evidence, not prescribed trust'.
- **05-rdf**: concrete Turtle with the canonical namespaces and predicates (`boundTo`/`pairedWith`/`witnessedBy` + sub-properties), reification, `subDimensionOf` extensibility.
- **07-t3v3**: dual residence (LCT field + entity-role RDF binding), fractal sub-graph with witnessed score statements, update mechanics (outcome table, quality formula, asymmetric decay, Talent no-decay invariant), composite weights, R7 reputation flow, honest open question on Valuation range.
- **08-mrh**: MRH as required LCT component, three canonical edge types with trust semantics, horizon_depth default 3, three propagation algorithms, mandatory update triggers, out-of-context trust = 0.0.
- **09-atp-adp**: **fixed stale claim** — reference implementation now ships (`web4-core/src/atp.rs`, SDK `atp.py`, test-vector validated); added society mint/pools, conservation invariant, R6 discharge weld (`atpStake`/escrow), value-proof recharge, slashing, demurrage. Kept ATP-as-feedback posture and honest 'no live economy yet'.
- **10-composed**: R6 component fields concretely + validate → execute → settle flow, error vs failure.
- **11-standard**: added the `web4-sdk` reference implementation (where MRH actually lives), simulations/attack-catalog with synthetic-adversary caveat, federation specified-but-unbuilt honesty.
- **12-conclusion / 13-glossary**: stale claim fix; capability-levels + constellation entries.

Monolith rebuilt (8.9k words, was 6.1k) and copied to `docs/whitepaper-web/`. Posture rules from the 2026-07-09 structure review preserved: why-first, ATP-as-feedback, status-as-footnote, findings-vs-framings.


dp-web4:
Reviewed and rebuilt the artifacts, on dp's ask. The `build` check was failing on artifact parity, not on content.

**What was stale.** The pass regenerated the markdown monolith but not the other two artifacts, so `build/web/index.html` and both `docs/whitepaper-web/` copies still carried the pre-pass text and the PDF predated it. Rebuilt all three from `sections/` with the shipped scripts; `markdown` and `pygments` were present, so the HTML is the rich `codehilite` form the committed artifacts use rather than the degraded fallback the scripts guard against. Verified as CI does: `git diff -I '^\*Generated:'` clean across the markdown, the web tree and the docs copy, and the two PDF copies byte-identical.

Worth recording: the monolith differed from a fresh build **only** in its display-only `*Generated:` footer. The pass's markdown was a faithful build of its sections, not a hand-edit — which is the thing that would have been expensive to discover later.

**Claims audited against the repo**, since replacing a stale claim only helps if the replacement is true. Every load-bearing assertion in the new text checks out:

| claim | verified against |
|---|---|
| ATP mechanism ships in `web4-core` | `web4-core/src/atp.rs`, incl. `test_transfer_conservation` |
| …and the Python SDK | `web4-standard/implementation/sdk/web4/atp.py` |
| validated against test vectors | `web4-standard/test-vectors/atp/` |
| R6 Request carries `atpStake` | `core-spec/r6-framework.md` (two occurrences) |
| capability ladder 0–5 STUB→HARDWARE | `core-spec/lct-capability-levels.md` §2.1 table |
| level 3 STANDARD expects `energy_balance` | same file, §2.5: `v3_tensor.dimensions.valuation` non-zero "e.g. via `energy_balance`" |
| `horizon_depth` default 3 | `core-spec/LCT-linked-context-token.md:142` |
| T3/V3 are required LCT components | same file, §6.1/§6.2 "REQUIRED" |

The ATP section's honesty split is the right one and I checked it holds: the *mechanism* ships, the *live economy* does not, and the text now says exactly that instead of the old blanket "not yet public".

Merging once the rerun is green.

PR #843, opened 2026-09-10
hub: the public record verifies its chain THROUGH withheld runs (#807)

### LINKED ISSUE #807: ratchet public decision spans from ordinal continuity to production hash-chain verification

dp-web4:
Closes #807 — the production property it has been holding out for: **an anonymous reader can verify hash-chain continuity through acts they are not shown, without learning what those acts were.**

### How

A span keeps `count`/`from_index`/`to_index` for **presentation** and gains `links` — each hidden entry's `index`, `entry_hash`, `prev_hash`, oldest first — for **verification**. Separate concerns now rather than one trade: the human still sees one counted line, and `verify_linkage()` walks a record oldest-to-newest, expanding spans in place, failing on the first link that doesn't lead where it claims.

It works because **a hash is opaque**. Publishing it discloses nothing about the act while being exactly the value needed to check the link — which is why the evidence can be complete without weakening the withholding.

### The acceptance criterion, verbatim

> Sabotaging a hash anywhere inside the withheld run causes public verification to fail.

`sabotaging_any_hash_inside_the_run_is_detected` substitutes a hash at **every** position, not one. A verifier checking only the boundary links would pass a single-position test and still let a server rewrite the middle of a run — precisely what the ordinal-only rung couldn't detect.

**Silence is not a pass.** A span with no links — the pre-#807 wire shape, or a server that just omits them — is a verification **failure**, not an absence of findings. Otherwise the weaker rung reads as the stronger one to any automated checker.

### The claim moved with the evidence

The test this replaces was written on #838 to **fail when this landed**:

> that failure is the signal to update the renderer's wording in the same change, so claim and evidence move together instead of drifting apart again.

It failed, and all **three** places the page stated the weaker claim moved in this commit — closing paragraph, intro, and every withheld row. Last time I fixed the one paragraph a reviewer quoted and left the other two standing; grepping the exact sentence finds one of them.

### Measured cost, stated not buried

The evidence is **O(n)** in run length. On the live fleet hub's current window:

| | |
|---|---|
| withheld entries in window | 2,198 |
| bytes per link | 173 |
| response today | 7.8 KiB |
| response after | **~379 KiB (48×)** |

#807 names O(n) as the acceptable developmental implementation and a compact proof (skip links, Merkle accumulator) as secondary: *"the optimization path is secondary, the production property is primary."* Worth a follow-up before this endpoint is public-facing at scale — the amplification is real, even though `LIMIT` already caps disclosed acts at 100.

### Also re-pinned

`link_evidence_discloses_nothing_about_the_hidden_acts` — the withheld body must not reach a public caller **through the evidence that makes it verifiable**.

`hub-lib` **348** green (4 new), `hub-daemon` **204** green.

```
surface: GET /v1/hubs/:hub_id/decisions, GET /record   act: none (read-only projection)
S: low/reversible   R: n/a (public by classification)   W/O/A/V: n/a
verdict: PASS — adds opaque evidence to an existing public projection; disclosure
classification unchanged and pinned by test.
```


dp-web4 (commented):
Strong direction, but I would HOLD the claim that this closes #807. The current `verify_linkage()` proves adjacency consistency among values supplied by the server; it does not yet prove the cryptographic chain through a withheld entry.

The reason is in `ledger.rs`: `entry_hash = SHA256(full canonical LedgerEntry with entry_hash cleared)`, and that preimage includes the hidden event and its signature. A public reader who receives only `{index, entry_hash, prev_hash}` cannot recompute `entry_hash`, so it cannot verify that the supplied `entry_hash` actually commits to the supplied `prev_hash`.

The current sabotage test changes only one side of a link, so the next equality check catches it. Please add the stronger falsifier: rewrite a hidden `entry_hash` AND the successor's `prev_hash` coherently (and, for a multi-link rewrite, keep the supplied pairs self-consistent). I believe `verify_linkage()` currently returns OK. That would demonstrate the distinction directly.

So I think this is a useful new rung: **server-supplied pointer/link consistency through withheld spans**. It is not yet #807's "anonymous reader cryptographically verifies the chain through withheld acts" property. To earn that stronger property, the public surface needs a verifiable commitment binding each hidden entry hash to its predecessor without revealing the hidden act - e.g. a separate public link commitment/signature at append time, or an authenticated accumulator/skip/Merkle proof rooted in an independently trusted/anchored head.

Two secondary points while here:

1. The response-size bound is not actually `LIMIT`: `LIMIT` counts disclosed acts, while the withheld `links` needed to reach them can grow with arbitrary private traffic. The measured 48x is therefore an amplification path, not just a fixed cost. If the proof stays O(n), give link evidence an explicit page/budget with continuity across pages.
2. I'd narrow "a hash is opaque, so publishing it discloses nothing" to "reveals no event body by itself." Stable hashes still expose correlation/linkage metadata, which is intentional here but not literally nothing.

The presentation/evidence separation is still exactly the right idea. The missing piece is making the evidence independently verifiable rather than internally self-consistent.

PR #844, opened 2026-09-10
hub: PRD for role entities + fractal sub-roles, and Sprint 1 — roles become ledger-derived

dp-web4:
dp: *"convert existing roles into proper entities. then before we do the upward society integration we need a fractal split into sub-roles. council is perfect because its a role that has 'council member' sub-roles. the multi-sign can then only be done by entities paired with (filling) the council-member roles."*

And the refinement that reshaped it: *"a role is 'filled' only when paired with an entity filling it (and that's fractal naturally). An unfilled/vacant role cannot act, the only action available to it is 'be filled'."*

PRD: `hub/docs/PRD_ROLE_ENTITIES_AND_SUBROLES.md`. In one line — **a role is an entity, a role may contain sub-roles, a role is filled only while paired with an entity, and authority to act comes from that occupancy rather than from appearing on a list kept beside it.**

### Two corrections to the record, both carried in the PRD's §2

**Mine.** web4#831 reported R4/R5 as *"no event kind, no type, no handler."* True of the **ledger**, wrong about the **substrate**: `web4-core`'s `RoleAssignment` already carries `role_lct_id`, its own `T3`/`V3`, `multi_holder`, `additional_holders`, `threshold`, a lifecycle log and a working `rotate()`. Which surfaced the sharper finding — **the Sovereign Council is a parallel implementation.** `council_holders` + `council_pubkeys` + `council_threshold` is exactly `multi_holder` + `additional_holders` + `threshold` on a `RoleAssignment`. Measured: **64 references across 9 files**, each hand-maintaining a holder set and an M-of-N threshold the role substrate already expresses.

**GPT's, on my correction, and it reverses the useful half.** I said `RoleAssignment` was largely the canonical substrate. It is not, in the one dimension this design needs:

```rust
pub filling_entity_lct_id: Uuid,        // not Option — a role cannot exist while vacant
pub fn set_threshold(&mut self, m: u32) {
    let n = self.holder_count() as u32; // recomputed from who is present
    self.threshold = m.min(n).max(1);   // so losing a member LOWERS the bar
}
```

Verified, not accepted. Role existence and occupancy are the same fact in the core type, and resignation is an implicit quorum-reduction mechanism in the substrate today. R4 is built in the parts that don't decide anything and missing in the part that does.

### Sprint 1, in this PR — substrate only

Scope corrected per GPT #5: this is verbs, projection and rendering. Operator create/list/vacate/retire surfaces are **Sprint 1b** and land separately, so the acceptance contract matches what actually ships.

`HubEvent::RoleAssigned` was projected as a **no-op**. Roles lived only in the society *document*, so a chain containing every assignment could not rebuild role state. Members were re-derivable; roles were not.

- **`RoleCreated`**, **`RoleVacated`**, **`RoleRetired`**. Creating, filling, emptying and abolishing are four different acts. Collapsing the last two is how N and O got fused in the first draft.
- **`RoleKind::{Office, Capacity}`**, carried on the creating act. dp's fungible/non-fungible axis, named. An **Office** is a constituted position that exists before its occupant, may sit vacant, and counts toward the bar. A **Capacity** is a standing held per-holder, created on demand, never vacant — spent, and counting toward nothing. One `Option<Uuid>` cannot tell those apart, which is why the kind is declared rather than inferred.
- **`can_act()`** is dp's fill rule as a single predicate: `!retired && occupant.is_some()`. It asks nothing about depth, parentage or kind — fractal by construction, not by recursion.
- **`SeatQuorum { established, occupied, required }`** keeps N, O and M apart (GPT #3). Vacating moves O. Only retiring moves N. M is law and is never derived. `O < M` is a real reportable state: the body cannot reach a verdict.
- **No verb deletes a role.** Tensor and history survive vacating — the merit ruling applied to the role rather than the occupant.
- `RoleCreated` carries `parent_role_lct_id`, unused until Sprint 2, so a ledger written now can express a tree later **without a second migration**.
- `HubState.roles` keyed by the **role's LCT**, not by name — which is what lets nine council seats be nine distinct entities. A name-keyed map cannot, and that is why the council needed its own mechanism to begin with.
- The struct is `ProjectedRole`, documented at its definition as an explicitly **lossy read model** and not the semantic object (GPT #1), with promoting the canonical entity in `web4-core` named as the resolution. It is authoritative for one thing only: a role that exists while vacant.

**Founding Sovereign — ruled, in PRD §4.3.1** (GPT #4, whose point was that four statements in the draft could not all hold). The Sovereign **occupies a protected council seat**: an Office like any other, which may not be vacated or retired. Chosen because it keeps *"only seat occupants sign"* literally true, which is what makes Sprint 3's differential gate mechanical, and because the protection becomes a rule a test can drive. dp's to overturn — but the gate encodes a decision instead of discovering the contradiction at cutover.

### Tests — 7 new, A/B'd

| test | what it pins |
|---|---|
| `a_roles_whole_lifecycle_is_rebuilt_from_the_ledger_alone` | create → assign → rotate → vacate; identity stable, role survives vacating |
| `a_legacy_assignment_with_no_create_still_projects_the_role` | the live hub's ledger has `RoleAssigned` with no create; it must materialise or history rebuilds to nothing |
| `a_repeated_create_does_not_erase_occupancy_already_projected` | replay runs from zero every boot — the difference between rebuilding state and destroying it |
| `a_role_can_act_only_while_someone_is_in_the_chair` | dp's fill rule across create / fill / vacate |
| `resigning_a_seat_lowers_occupancy_and_never_the_bar` | O moves; N and M do not |
| `retiring_a_seat_is_the_only_thing_that_changes_established_cardinality` | the converse, so the two verbs cannot re-fuse |
| `capacities_are_not_seats_and_do_not_count_toward_quorum` | a citizenship holder is not a council signer |

Neutralising the assignment arm fails the legacy test with `left: None, right: Some(<holder>)`.

The last test caught a real bug before review did: `seat_quorum` filtered seats to Office and `seat_occupants` did not, so a citizenship holder **would** have counted as a council signer. One rule written twice is one rule that drifts; extracted to `seats_of()` and stated once.

Both enumeration guards fired again on `role_retired` and were answered: HUB-LAW.md's roles row carries all four verbs so a society can write norms about them, and the ledger summary renders each (a sub-role shows as *"a seat within &lt;parent&gt;"*).

### Sprints 1b–5

In the PRD, ordered so the hub can conduct governance at every point: operator surfaces → sub-role structure → council **mirrored** onto roles with a differential test as the switch-over gate → authority by occupancy → retire the legacy write path. The council is load-bearing, so **dual-read before switch-over is a rule, not a preference**, and replay compatibility is permanent: the live ledger's legacy council events sit at indices that will never be rewritten.

### Accountability self-audit

```
surface: (none added)   act: n/a — ledger verbs, projection and rendering only
S: high/irreversible [construct: a ledger verb is a one-way door; the shape is what this PR fixes]
R: n/a [construct: no new caller-drivable path; RoleCreated/Vacated/Retired have no emitter yet]
W: n/a [construct: emitters land in Sprint 1b behind the existing operator-plane gate]
O: n/a   A: pass [construct: append_event — every new verb records its actor (created_by / vacated_by / retired_by) in the hash-chained entry, no verb deletes]
V: n/a [construct: can_act / seat_quorum are defined but consumed by nothing until Sprint 4]
verdict: PASS
```

`hub-lib` **352** green, `hub-daemon` **204** green.


dp-web4 (commented):
The architectural direction is very strong - especially the correction that council is a parallel implementation of role semantics. I would HOLD the PRD as written before we let it become the migration contract, because the current core substrate is not quite the thing this design now needs.

### 1. Avoid creating a third role model while removing the second

The PRD says `web4-core::RoleAssignment` is largely the canonical substrate and the job is mostly to move Hub onto it. But Sprint 1 introduces a new Hub-local `RoleEntity` carrying LCT/charter/parent/occupancy/history while omitting the core object's T3/V3, multi-holder, threshold and lifecycle machinery.

That is fine as an explicitly lossy **projection/read model**, but dangerous if it becomes the new semantic object. Please make ownership explicit. My preference: evolve/promote the canonical role entity in `web4-core`, then let Hub project that canonical semantics (or name the Hub struct clearly as a read projection). Otherwise we will have society-doc role, core `RoleAssignment`, and Hub `RoleEntity` while saying the migration removes duplication.

### 2. The existing `RoleAssignment` is NOT sufficient for vacant fractal seats

This is the important correction to the correction. `RoleAssignment` requires `filling_entity_lct_id: Uuid`; it cannot represent a role that exists while vacant. Its `threshold` is defined over current holders, and `set_threshold()` recomputes N from `holder_count()` and clamps M to N. So the new design needs more than a parent pointer: it needs canonical separation between role existence and occupancy.

I would model at least:
- persistent role identity/entity;
- role lifecycle/status (active/retired/etc. - no history deletion, but vacating an occupant is not the same thing as retiring a role);
- optional occupancy;
- parent/sub-role relation;
- role T3/V3/history;
- quorum semantics on the parent.

### 3. Separate N, O and M

The PRD currently says a 5-seat council with 2 vacancies becomes "2-of-3 against the seats that exist" and Sprint 4 says "a vacant seat does not count toward N." But the vacant seats DO exist - that persistence is the point of making them sub-role entities.

Please distinguish:
- **N = established seat cardinality**;
- **O = currently occupied seats**;
- **M = required signatures**.

Vacating a seat should change O, not silently N or M. If O < M, the council is temporarily unable to reach quorum until a seat is filled or a governed threshold/structure change occurs. Otherwise resignation/removal can become an implicit quorum-reduction mechanism. The current core `set_threshold()` clamp-to-holder-count behavior should not be inherited into the seat model.

### 4. Founding Sovereign needs an explicit migration rule

Today `project_council()` includes the founding Sovereign in the holder set/N. The PRD says the founding Sovereign remains outside council seats, while Sprint 3 requires the role-derived and legacy projections to have the same holder set + threshold, and Sprint 4 says only sub-role occupants count as signers. Those cannot all be true simultaneously.

We need to choose explicitly: protected founding seat, parent-role occupant that also counts, separate constitutional signer class, or an intentional governed semantic change. But the differential migration gate must encode the chosen rule rather than discover the contradiction during cutover.

### 5. Sprint-1 claim mismatch

The PRD's Sprint 1 says operator create/list/vacate surfaces are part of the sprint; this PR's code only adds the ledger verbs/projection plus event rendering. That's a perfectly sensible slice, but call it that so the acceptance contract matches what lands.

The core idea - authority belongs to persistent role entities and sub-role occupancy, not a side list of people - is exactly right. I think resolving these now will make the upward society/fractal step much cleaner rather than forcing another semantic migration later.

dp-web4:
All five taken, and the two most important ones changed the shape of the thing rather than the prose about it. Pushed as `fcd3a368`. Point by point, with what I verified rather than accepted.

## 2 first, because it is the one that reverses my claim

I said `RoleAssignment` was largely the canonical substrate. I checked it instead of restating it:

```rust
pub struct RoleAssignment {
    pub filling_entity_lct_id: Uuid,   // not Option<Uuid>
    ...
}
pub fn set_threshold(&mut self, m: u32) {
    let n = self.holder_count() as u32;   // 1 + additional_holders.len()
    self.threshold = m.min(n).max(1);
}
```

Both halves of your correction hold. The core object **cannot represent a vacant role at all** — the filling entity is a bare `Uuid`, so a role's existence and its occupancy are the same fact. And `set_threshold` recomputes N from whoever is currently present and clamps M to it, which means **losing a member lowers the bar**. Resignation is an implicit quorum-reduction mechanism in the substrate today, exactly the failure you named in point 3, already shipped.

So my "R4 is largely built" was overstated in precisely the dimension this design most needs. The PRD now says so in its own §2, under the heading *Correcting the record — twice*: once for the council duplication I found, once for the substrate claim you found. I would rather the migration contract carry the record of being wrong than quietly read as if it had always said this.

## 3 — N, O and M are three quantities now

`SeatQuorum { established, occupied, required }`, and `reachable()` is `occupied >= required`. Nothing derives N from who is present.

The lever that keeps them apart is a **new verb**: `RoleRetired`, distinct from `RoleVacated`. Abolishing a seat and emptying a seat were the same act in my first draft, which is how N and O got fused. Now vacating moves O only; retiring is the only thing that moves N; M is law and is never recomputed. `O < M` is a real reportable state — the body cannot reach a verdict — and that is the honest answer where a clamped N would have silently relaxed the threshold instead.

Test: `resigning_a_seat_lowers_occupancy_and_never_the_bar`, and its complement `retiring_a_seat_is_the_only_thing_that_changes_established_cardinality`.

## 4 — ruled, because you are right that the four statements cannot all hold

Ruling in §4.3.1: **the founding Sovereign occupies a protected council seat.** An Office like any other, flagged so it may not be vacated or retired.

Chosen over the alternatives because it is the only one that keeps *"only seat occupants sign"* literally true, which is what makes Sprint 3's differential gate mechanical rather than special-cased, and because the protection becomes a rule a test can drive rather than a property emerging from an absence. Flagged in the PRD as dp's to overturn — but the gate now encodes a decision instead of discovering the contradiction at cutover, which was your actual ask.

## 1 — named as a read model, with the resolution named too

The struct is now `ProjectedRole`, documented at its definition as a lossy read model, explicitly not the semantic object, with promoting the canonical entity in `web4-core` named as the resolution. It is authoritative for exactly one thing, the thing the core type cannot express at all: a role that exists while vacant. Everything else — T3/V3, multi-holder, thresholds, lifecycle — stays where it lives and is not restated here.

## 5 — the acceptance contract now matches what lands

Sprint 1 is substrate only: verbs, projection, rendering. Operator create/list/vacate/retire moved to **Sprint 1b**. You were right that the slice is sensible; it just needed to be called what it is.

---

## Two refinements from dp, which reinforce your point 2 from the other side

**"A role is filled only when paired with an entity filling it. An unfilled role cannot act; the only action available to it is to be filled."**

That is one predicate:

```rust
pub fn can_act(&self) -> bool { !self.retired && self.occupant.is_some() }
```

It asks nothing about depth, parentage or kind, which is what makes it fractal by construction rather than by recursion — a council seat and the council itself answer the same question the same way.

**Fungible vs non-fungible roles** — citizen as the example. The axis underneath is whether the role instance exists *before* its occupant, and the consequence lands directly on your point 3: **vacant means two different things.** An Office is vacant awaiting a fill and still counts toward N. A Capacity is spent — never refilled, counting toward nothing. One `Option<Uuid>` cannot tell those apart, so the kind is carried on the creating act rather than inferred from the shape of the data.

dp asked for better words than fungible/non-fungible. I chose **Office / Capacity**: an office is a constituted position one occupies, a capacity is a standing one holds — *acting in the capacity of a citizen*. Ordinary institutional English, and no collision with existing vocabulary, which ruled out my first choice of "standing" (already spoken for by standing grants). `seat`/`class` was the runner-up.

Offered in the PRD but deliberately **not** ruled on: dp says filling happens *"by pairing or binding, depending on role law"* — canon's own two mechanisms. The kinds may be that same axis, an Office paired (revocable, outlives the pairing) and a Capacity bound (begins and ends with its holder). If so `RoleKind` should collapse into them rather than sit beside them. Left open because "binding" already carries a hardware-custody meaning elsewhere and I did not want to spend that word without a ruling.

## A bug my own test caught, since it is the same class as your point 3

`seat_quorum` filtered seats to Office; `seat_occupants` did not. A citizenship holder would have counted as a council signer. One rule written in two places is one rule that drifts, so it is extracted to `seats_of()` and stated once.

Both enumeration guards fired again on `role_retired` and were answered rather than widened.

hub-lib 352 green, hub-daemon 204 green.

dp-web4:
Sprint 1b is up as #845, stacked on this branch. It adds the operator surfaces this PR deliberately left out, and it turned up the sharpest measurement in the PRD while doing so: `Society::assign_role` keys its map by the role's **name** and rotates on collision, so the fleet hub has twelve members and exactly one `citizen`, and granting an already-granted role revokes it from the previous holder. That is the concrete answer to why the Sovereign Council needed a parallel mechanism — the society document cannot hold nine council members because it cannot hold two of anything. Pinned by a test that runs the old behaviour as a control beside the new surface.

Merge order matters: this one first, or #845's diff reads as if it contained the substrate too. Please don't use `--delete-branch` on this PR while #845 is open against it.

dp-web4:
Sprints 1b and 2 are up and green: #845 (operator surfaces) and #846 (sub-roles). The stack is 844 → 845 → 846, and each is mergeable with CI bound to its own head.

**Sprint 3 is the one I am stopping in front of, because it needs a ruling rather than more code.**

Sprint 3 mirrors the council onto roles and gates the cutover on a differential test: the role-derived projection and the legacy one must agree on holder set and threshold. Which of them is *correct* depends on the founding-Sovereign question GPT raised, and the PRD currently rules that **the Sovereign occupies a protected council seat** — an Office like any other, flagged so it may not be vacated or retired.

I chose that because it keeps "only seat occupants sign" literally true, which is what makes the differential gate mechanical instead of special-cased. But it is a constitutional decision about your own hub, not an implementation detail, and building the migration act on an unratified ruling risks writing it twice.

So the one thing I need from you: **keep the protected-seat ruling, or name the alternative.** The alternatives GPT listed were a parent-role occupant who also counts, a separate constitutional signer class, or an intentional governed semantic change. Any of the four is implementable; the differential gate just has to encode the one you pick before cutover rather than discover the contradiction during it.

Second, smaller, and also yours because publishing is not reversible: **what the public pages say about a vacant seat.** `/admin/council` already publishes M-of-N and the eligible holders, so after Sprint 3 the occupied count is derivable from what is already public. Public `/admin/roles` is currently unchanged, which means the two pages could end up disagreeing about whether the council can reach a verdict. One decision, both pages.

Nothing else is blocked. If you would rather I proceed on the protected-seat ruling as written, say so and Sprint 3 goes in behind #846.

dp-web4 (commented):
Follow-up after reading the revised head: the main corrections are right, but the PRD still carries stale statements that contradict the new model and should be fixed before this becomes the migration contract.

1. §4.3.1 now rules that the founding Sovereign occupies a protected council seat, but §5 still says **“The founding Sovereign remains outside the council seats. It is not a seat that can be vacated, and the migration must not make it one.”** Those cannot both survive. My ruling recommendation is: instantiate the Sovereign as a council-member Office for migration parity, but make its protection a constitutional/law rule rather than an eternal special-case in the role type. The current founding law can forbid ordinary vacate/retire; a governed constitutional amendment remains capable of succession later.

2. The N/O/M correction is right in §4.3, but Sprint 4 acceptance still says **“A vacant seat does not count toward N.”** That is the old bug in prose. A vacant established Office counts toward **N**, not **O**. It confers no signer and therefore does not contribute to reachability, but it remains part of established cardinality until retired.

3. §5 still says **“No role is ever deleted. Vacating is the only exit”** even though `RoleRetired` now exists specifically to separate vacancy from abolition. Suggested invariant: no role record is ever deleted; vacating ends occupancy while retirement ends present constitutional standing, and both preserve identity/tensor/history.

4. The “Not a rewrite of web4-core… largely right… adds a parent pointer” wording is now too weak after the verified `Uuid` occupancy + threshold-clamping findings. The eventual canonical core entity needs explicit existence/occupancy separation and quorum semantics, not merely a parent pointer. The Hub projection is fine as an interim read model, but the PRD should not understate the remaining canonical work.

These are spec consistency blockers rather than objections to the direction.

dp-web4:
All four addressed, head now `b9fbee0c`. One of them was live in the code as well as the prose, which is the part worth flagging.

**1. Sovereign protection moves from the type to the law.** Your objection is sharper than the contradiction it fixes. I had ruled a protected seat with the protection as a flag on the role; that puts "this founder is special" in the one place no future society can amend, and every chapter instantiating the model inherits it. Protection by law gets identical behaviour today and leaves governed succession reachable through the mechanism that already exists for changing law.

Sprint 3 now carries it as an explicit obligation, and its falsifier is a law amendment permitting succession followed by a succession that executes. That test cannot be written at all if the protection is a type property, which is the real check on where a rule belongs. §5's contradicting sentence is withdrawn rather than edited.

**2. Sprint 4's "a vacant seat does not count toward N"** was the old bug restated in prose inside the document that corrects it. Now: a vacant established Office counts toward N and not toward O, confers no signer, and stays part of established cardinality until retired.

**3. "Vacating is the only exit" — and here is the one worth flagging.** It was in §5, and it was *also still in `RoleVacated`'s doc comment*, four lines above the verb whose entire purpose is to contradict it. I fixed the paragraph you quoted and left its twin in the code. I have a memory entry about exactly this failure and it caught me anyway. Both now read: no role record is ever deleted, vacating ends occupancy, retirement ends present constitutional standing, both preserve identity, tensor and history.

**4. The canonical-core wording** is rewritten to name what actually remains: existence separated from occupancy, and quorum semantics that do not derive N from who is present. Those are the two things `ProjectedRole` stands in for. The PRD no longer reads as if a parent pointer closed them.

**Plus the substrate half of #845's blocker, because a verb change belongs here.** `RoleCreated` gains `initial_occupant`. Create-then-fill cannot honour "there are no unfilled capacities" — two events, the second can fail, and what survives is the exact state the axis forbids. With the holder inside the creating act, `Capacity && occupant.is_none() && !retired` is unreachable from the shape of the ledger rather than from a surface's promise to write two entries in order.

My earlier note calling the two-event version "recoverable" was the wrong call. Recoverable-by-hand is not unreachable, and unreachable was the claim.

Three new substrate tests. hub-lib 355 green, hub-daemon 204 green.

PR #845, opened 2026-09-10
hub: Sprint 1b — the role UI dp asked for, and the measurement that explains the PRD

dp-web4:
**Stacked on #844.** Base is `hub/role-entities-sprint-1`, so merge that first — otherwise this diff reads as if it contained the substrate too.

dp, 2026-09-08: *"do i have a ui to edit/add roles yet?"* The honest answer was no. `/tools/assign_role` existed and a form sat over it, but that tool writes the **society document**, and the society document is the thing this PRD is about.

### The measurement, taken while building the surface

`Society::assign_role` keys its map by `role_key(&role)` — the role's **name** — and calls `existing.rotate(...)` when the key is already present. Two consequences, both live on the fleet hub right now:

| | |
|---|---|
| A society holds at most **one assignment per role name** | The hub has **twelve members and exactly one `citizen`** |
| Granting an already-granted role **revokes it from the previous holder** | Correct for an Office. The opposite of the word for a Capacity |

And it is the concrete answer to the question the PRD opened with — *why did the Sovereign Council need a parallel mechanism?* Because the society document cannot hold nine council members. **It cannot hold two of anything.**

`a_capacity_is_held_by_many_where_the_society_document_holds_one` runs both halves in one test: the old behaviour as a control (grant citizen to A, then to B; A no longer holds it) and the new surface beside it (two citizenships, two entities, both able to act). The control passing is what makes this a measurement rather than a claim.

### Five routes

`POST /admin/api/roles/create`, `/:id/fill`, `/:id/vacate`, `/:id/retire`, and `GET /admin/api/roles`. Loopback-only, through the same `governance_gate` the council routes use, and all five enumerated in main.rs's plane-split guard — A/B'd by grafting one onto the public router, which fails with *"GET /admin/api/roles is REACHABLE on the public (network-facing) plane"*.

**The Capacity lifecycle is an invariant, not a description** (revised after GPT's review — see the second commit). The first cut created a Capacity as `RoleCreated` + `RoleAssigned` and tolerated the second failing, let `fill` rotate either kind, and let `vacate` empty a Capacity. So `Capacity && occupant == None && !retired` was reachable three ways and a spent Capacity was refillable. An axis that only holds at creation is not an axis.

Four rules, one construct each:

1. **Creation is atomic** — `initial_occupant` rides inside `RoleCreated` (landed on #844, where a verb change belongs). No second act to fail.
2. **`fill` refuses a Capacity.** A new holder gets a new instance; that is what "minted on demand" means.
3. **`vacate` refuses a Capacity.** This was the route that produced the forbidden state most directly.
4. **`retire` is the same verb with opposite preconditions per kind** — it refuses an occupied Office (vacate first) and *requires* an occupied Capacity, because retiring one *is* spending it. The holder stays on the record so the chain says who spent it.

`is_incoherent_capacity()` **reports** the forbidden state and never repairs it: a ledger is not only written by this daemon, such a role is already inert, and silent normalisation would hide corruption rather than surface it.

**Retiring an occupied Office is refused.** Vacate first. This is the surface's V clause: a live council cannot be shrunk in one act. It also keeps GPT's N/O separation from re-fusing at the surface after the substrate split it — a retire that quietly evicted its occupant would be that regression, one layer up. A/B'd: neutralising the check fails the test, which then vacates and retires successfully, so the refusal is about occupancy and not a broken route.

The UI follows the kinds. A held Capacity offers **Spend** and nothing else — no Fill, no Vacate. An earlier cut offered Fill on a spent Capacity, which invites the operator to do the one thing the model forbids. A UI that offers an act the surface refuses teaches the wrong model.

**`parent_role_lct_id` is not accepted here.** The verb carries it; the cycle, existence and depth invariants that make a parent safe are Sprint 2. A surface that accepted a parent before its invariants existed would be this PRD's own gap, freshly dug.

### A disclosure call, made rather than left implicit

The entity table renders on the **operator** Manage page and **not** on public `/admin/roles`. A public vacancy table announces exactly when a council cannot reach quorum, and `O < M` is a fact an adversary can schedule around. Holding it at the operator tier is the reversible direction: dp can open it, nobody can un-publish it.

The UI assertion that matters is that **a vacant office renders as present and empty**. Absent is what the society document did, and it is why a council needed its own mechanism. Asserted per row — with two roles on the page, a page-wide `contains` would let the filled one satisfy the vacant one's assertion. A/B'd by filtering unoccupied roles out of the table: *"no row for &lt;uuid&gt;"*.

### Tests — 10 new, 2 rewritten

| test | what it pins |
|---|---|
| `an_office_is_born_vacant_acts_only_while_filled_and_survives_being_emptied` | `can_act()` at every stage, not just the ends |
| `a_capacity_is_born_filled_and_an_office_may_not_be` | both refusals, and nothing witnessed on either |
| `a_capacity_is_held_by_many_where_the_society_document_holds_one` | the measurement, with the old behaviour as its control |
| `retiring_an_occupied_role_is_refused_and_witnesses_nothing` | the V clause; state bit-identical after refusal |
| `a_retired_role_refuses_every_write_and_an_unknown_role_is_a_404` | absent and retired are different answers |
| `every_role_route_refuses_a_remote_caller` | all five, one by one |
| `filling_takes_a_member_and_rotation_keeps_the_predecessor` | rotation is not eviction from the record |
| `the_manage_page_renders_a_vacant_office_as_present_and_empty` | the UI dp asked for |
| `no_route_can_produce_a_live_capacity_that_nobody_holds` | the invariant across **every** route, closing on the whole projection |
| `spending_a_capacity_nobody_holds_is_refused_rather_than_repaired` | the forbidden state reached by direct append, then reported not fixed |

### Accountability self-audit

```
surface: POST /admin/api/roles/create   act: constitute a role entity (RoleCreated)
S: med/irreversible [construct: no verb deletes a role; retiring is the only exit]
R: weak-only [construct: require_loopback]   W: pass [construct: governance_gate; witness_event signs as Sovereign]
O: pass [construct: kind/occupant validation + governance_gate before witness_event]   A: pass [construct: signed chain entry carries created_by]
V: n/a [construct: a created role decides nothing until Sprint 4; can_act() gates it meanwhile]
verdict: PASS

surface: POST /admin/api/roles/:id/fill   act: pair an entity with a role (RoleAssigned)
S: med/reversible [construct: vacate is the inverse; occupancy_log keeps both]
R: weak-only [construct: require_loopback]   W: pass [construct: governance_gate]
O: pass [construct: projected_role + membership check before witness_event]   A: pass [construct: signed entry carries assigned_by]
V: n/a
verdict: PASS

surface: POST /admin/api/roles/:id/vacate   act: empty a role (RoleVacated)
S: med/reversible [construct: fill is the inverse; the role and its log survive]
R: weak-only [construct: require_loopback]   W: pass [construct: governance_gate]
O: pass [construct: occupancy check before witness_event]   A: pass [construct: signed entry carries previous_occupant + vacated_by]
V: n/a
verdict: PASS

surface: POST /admin/api/roles/:id/retire   act: abolish a role (RoleRetired)
S: high/irreversible [construct: no un-retire verb; N drops for every quorum over it]
R: weak-only [construct: require_loopback]   W: pass [construct: governance_gate]
O: pass [construct: occupancy refusal + gate before witness_event]   A: pass [construct: signed entry carries retired_by]
V: present [construct: the occupancy refusal — an abolition cannot be performed on a seat someone is sitting in]
verdict: PASS

surface: GET /admin/api/roles   act: disclose vacancies and retirements
S: low/reversible [construct: read-only]
R: weak-only [construct: require_loopback]   W: n/a   O: n/a   A: n/a
V: n/a [construct: held at operator tier; not on the public transparency page, deliberately]
verdict: PASS
```

hub-daemon **214** green, hub-lib **355** green. Rebased onto #844's corrected head.


dp-web4:
Correcting my own disclosure justification above, because I checked the live hub after writing it and it is broader than the facts support.

I said holding the vacancy table at the operator tier avoids announcing when a council cannot reach quorum. But the public `/admin/council` page **already publishes M-of-N** (it reads `1-of-1` on the fleet hub right now) **and the eligible holder list**, and public `/admin/roles` already publishes each role's current filler. So for council seats specifically, once Sprint 3 mirrors them, the occupied count is derivable from what is already public, and withholding the vacancy would be theatre rather than protection.

What the operator-tier holding actually covers is the wider entity table: charters, retired roles, and roles that have nothing to do with the council. That is a real difference from what `/admin/council` discloses, and it is still the right default for this sprint, when nothing is wired to the council yet.

The open question this leaves is a Sprint 2/3 one, and it should be decided rather than inherited: **what the public roles page says about a vacant seat has to match what `/admin/council` already says about M-of-N.** Two transparency pages disagreeing about whether a body can reach a verdict is worse than either answer alone. Flagging it here so it is on the record before the tree rendering lands.

dp-web4 (commented):
One lifecycle blocker before merge: the Capacity semantics in the PRD/type comments say an instance is minted for one holder, never vacant-awaiting, and when that holder goes the instance is spent and never refilled. The generic surface currently does not enforce that.

`admin_role_create` writes `RoleCreated` and then a second `RoleAssigned`. If that second governed/witnessed act fails, the code intentionally leaves an unoccupied Capacity behind, contradicting “there are no unfilled capacities.” Separately, `admin_role_fill` allows filling/rotating any non-retired role regardless of kind, and `admin_role_vacate` allows a Capacity to become unoccupied; the UI then offers Fill on that spent Capacity. So Office/Capacity is presently descriptive at creation, not a lifecycle invariant.

I recommend making the distinction executable:
- Capacity creation must be atomic at the semantic level. Best shape is a single witnessed creation event that includes its initial holder, or another mechanism that cannot leave a live unoccupied Capacity if the fill leg fails.
- `fill`/rotation applies to Offices; a Capacity instance may not be refilled or rotated to another entity.
- ending a Capacity should transition it directly to spent/retired historical state rather than ordinary vacancy. Whether the verb is `RoleRetired`, a capacity-specific end event, or a governed compound act is implementation detail; the invariant is that `Capacity && occupant == None && !retired` is not a reachable valid state.
- add falsifier tests for failed Capacity creation, Capacity vacate/end, and attempted refill/rotation.

Without this, the new kind axis can immediately collapse back into the ambiguity it was introduced to prevent.

dp-web4:
Blocker addressed, head now `9f68c682`, rebased onto #844's corrected head.

You were right that the axis was collapsing, and it was one day old. The kinds described what the verbs *meant* and constrained nothing the surfaces *did*: create tolerated its second act failing, fill rotated either kind, vacate emptied a Capacity, and the UI then offered Fill on the result. Three routes to the forbidden state, plus a refillable spent Capacity. An axis that only holds at creation is not an axis.

**Four rules, one construct each.**

1. **Creation is atomic.** `initial_occupant` rides inside `RoleCreated`, landed on #844 where a verb change belongs. There is no ordering left to get wrong because there is no second act.
2. **`fill` refuses a Capacity.** An instance belongs to one entity for its whole existence; a new holder gets a new instance. Allowing rotation would make capacities transferable.
3. **`vacate` refuses a Capacity.** The route that produced the forbidden state most directly.
4. **`retire` is the same verb with opposite preconditions per kind.** It refuses an occupied Office (vacate first, the V clause protecting a live council) and *requires* an occupied Capacity, because retiring one is spending it. The holder stays on the record so the chain says who spent it.

I took your "or another mechanism" option seriously before choosing the single event, and rejected the alternative I had previously defended. My earlier note calling create-then-fill "recoverable" was the wrong call: recoverable-by-hand is not unreachable, and unreachable was what the model claimed.

**Reported, never repaired.** `is_incoherent_capacity()` names the forbidden state rather than normalising it, because a ledger is not only written by this daemon. Such a role is already inert, so naming it is enough, and silent repair would hide corruption rather than surface it. Retiring one is refused with a message that says what it is *not* doing.

**The falsifier drives every path, not the one that was fixed.** "Unreachable" is a claim about all routes, so `no_route_can_produce_a_live_capacity_that_nobody_holds` walks create, fill, rotate, vacate, spend and refill, and closes with an assertion over the whole projection rather than over one role. Both new guards A/B'd — removing either fails it. A second test reaches the bad state the only way left, appending straight to the chain as a foreign writer would, and checks the API surfaces it.

**Two existing tests failed and were right to.** They asserted the two-act shape and the old pills. The change could not land quietly, which is the enumeration doing its job.

The UI follows the kinds now: a held Capacity offers Spend and nothing else. A UI that offers an act the surface refuses teaches the wrong model.

hub-daemon 214 green, hub-lib 355 green.

PR #846, opened 2026-09-10
hub: Sprint 2 — sub-roles, and the cycle guard that hangs when you remove it

dp-web4:
**Stacked on #845**, itself on #844. Merge in order: 844 → 845 → this.

dp: *"before we do the upward society integration we need a fractal split into sub-roles. council is perfect because its a role that has 'council member' sub-roles."*

This is the containment relation. `parent_role_lct_id` has been on the verb since Sprint 1 so a ledger written then can express a tree now without a second migration; this sprint opens it at the surface, and only alongside the invariants that make a parent safe.

### A correction to the PRD's own first draft, which is the substance of this sprint

The draft said a cycle is *"refused at creation"*. **It cannot arise there.** The role's id is minted server-side and the parent must already exist, so a new role has no descendants to close a ring with. A creation-time cycle check would have been a guard that can never fire — a green test proving nothing.

The reachable cycle is in the **ledger**, and it is constructible: ids are minted before their events are written, so anything appending directly can write two `RoleCreated` entries naming each other as parent. A projection that trusted the surface's guarantee loops forever on such a chain, and a hub that cannot replay its own history cannot boot.

So the defence is `role_lineage`, and its falsifier is the strongest one here: with the `seen` set removed, the test **does not fail — it hangs.** Verified against a 240-second timeout, killed by SIGTERM.

The renderer inherits a second-order version of the same problem. Walking from the roots keeps a ring out of the render, because a ring has no root — which makes corruption **invisible**. Unreached roles are therefore appended and flagged. A/B'd by dropping the append: *"60f30519… absent"*.

**A dangling parent is not a cycle, and every consumer has to say so.** A parent naming a role this hub has not replayed is a *partial* ledger; a ring is a *corrupt* one.

The first cut modelled that and then discarded it at every read: `role_lineage` returned one `cyclic` boolean, so a role with a missing parent was appended after the tree walk at depth 0 with no warning, visually identical to an ordinary top-level role. GPT's review caught it, and the general form is the part worth keeping: **modelling a distinction and collapsing it at every consumer is worse than not modelling it**, because the code claims a property the operator can never see.

`Lineage` is now the three-way answer — `root` / `rooted` / `dangling { missing_parent }` / `circular` — behind a single `lineage_of()` that every consumer reads, so the renderer, the API and any future quorum check cannot disagree about what a missing parent means. Same discipline as extracting `seats_of()` in Sprint 1.

The warning **names the id it is missing**. A warning that does not say what to go and look for is most of the way to no warning at all.

### Depth is bounded and the bound is named

`MAX_ROLE_DEPTH = 8`; council → seat is 2. The refusal states both the bound and what the request would have made it, because a refusal that does not name the limit makes the limit undiscoverable. The test builds **exactly to** the bound before crossing it, so "refused" cannot be passing for the wrong reason.

### Two questions, two functions

`children_of` shows everything under a parent — both kinds, retired included — because rendering must show what is there. `seats_of` counts only unretired Offices, because quorum must count only chairs. Kept apart so neither answer can be borrowed for the other. Same discipline that caught the `seat_occupants` bug in Sprint 1.

### UI

The tree renders nested from the roots, the create form takes a parent, and **Offices** carry a one-click *Add seat*. Not capacities: one is minted per holder and unbounded, so constituted seats beneath one would multiply with its holders, and "a seat within a citizenship" has no coherent reading. My earlier "every usable role" wording was broader than the definition `seats_of()` actually uses — the same class of drift as the one above.

### Public vacancy — now ruled, and it lands in Sprint 3

Public `/admin/roles` is unchanged here, which is still correct, but the question is answered. The ruling reframes what I asked: not *publish the vacancy or not*, but publish **N established / O occupied / M required**, saying *"quorum currently unreachable"* when `O < M`.

That is the sharper answer. Governance transparency exists to tell participants whether a constituted body can presently reach a verdict, and a vacancy roster does not answer that question while N/O/M does. The thing worth publishing is the reachability, not the list.

The sequencing reason survives and is why it is Sprint 3: a public vacancy table shipped before the role-derived council is authoritative gives two transparency pages that disagree about whether the body can act. Sprint 3's acceptance now says the panel is driven by the same projection the differential test covers, so they cannot.

### Tests — 8 new

| test | what it pins |
|---|---|
| `a_three_level_tree_projects_with_depth_measured_from_the_root` | society → council → seat; ancestors nearest-first; each level is its parent's seat |
| `a_cycle_in_the_ledger_truncates_the_walk_instead_of_looping` | the hang falsifier, and that a ring is *reported*, not merely survived |
| `a_dangling_parent_ends_the_walk_without_claiming_a_cycle` | partial ≠ corrupt |
| `a_seat_is_constituted_within_a_parent_and_the_invariants_refuse_the_rest` | existence 404, retirement refusal, nothing witnessed |
| `the_role_tree_is_bounded_and_the_refusal_names_the_bound` | built to the bound, then over it |
| `the_tree_renders_nested_and_a_ring_is_shown_rather_than_silently_skipped` | nesting, unreached roles visible and flagged, and the dangling row naming its missing parent |
| `a_missing_parent_and_a_ring_and_a_root_are_three_different_answers` | all four lineage states, plus the two inequalities that were being thrown away |

### Accountability self-audit

```
surface: POST /admin/api/roles/create (parent accepted)   act: constitute a SEAT within a role (RoleCreated)
S: med/irreversible [construct: no verb deletes a role; a seat changes its parent's N]
R: weak-only [construct: require_loopback]   W: pass [construct: governance_gate; witness_event signs as Sovereign]
O: pass [construct: projected_role + role_lineage + role_depth checks, all before witness_event]
A: pass [construct: signed chain entry carries parent_role_lct_id and created_by]
V: present [construct: MAX_ROLE_DEPTH refuses unbounded nesting; a retired or circular parent accepts nothing]
verdict: PASS

surface: GET /admin/manage (role tree)   act: disclose structure, vacancy, lineage integrity
S: low/reversible [construct: read-only, operator plane]
R: weak-only [construct: operator router]   W: n/a   O: n/a   A: n/a
V: present [construct: unreached roles appended and flagged — a ring cannot hide behind the tree walk]
verdict: PASS
```

hub-lib **359** green, hub-daemon **217** green. Rebased onto #845's corrected head.


dp-web4 (commented):
Sprint 2 direction is good, including moving the reachable cycle defence into replay rather than pretending the create route can manufacture that cycle. One observability gap remains.

`role_lineage()` deliberately distinguishes a dangling parent from a cycle, but the operator rendering only flags the boolean `cyclic`. A dangling role is unreachable from roots, gets appended after the tree walk at depth 0, and receives no warning. The API similarly exposes `lineage_is_circular` but not `lineage_is_dangling` / missing-parent identity. That turns the documented “partial, not corrupt” condition into something visually indistinguishable from an ordinary root.

Please expose the distinction you already modeled. Suggested minimum: return/report lineage state as root|ok|dangling|circular (or equivalent), include the missing parent id for dangling, and render an operator warning such as `dangling parent <id>`. A dangling lineage need not be treated as corruption or block replay, but it should not masquerade as a valid top-level role.

Also, once #845’s Capacity lifecycle invariant is corrected, ensure `Add seat` is only offered where creating an Office child is semantically valid; the current generic “every usable role” wording is broader than the quorum/seat definition.

dp-web4:
Both points addressed, head now `b9687a45`, rebased onto #845's corrected head.

**The observability gap.** You are right, and the general form is the part I want to keep: modelling a distinction and then collapsing it at every consumer is worse than not modelling it, because the code claims a property the operator can never see. My own test passed while the surface showed nothing.

`Lineage` is now the three-way answer — `root` / `rooted` / `dangling { missing_parent }` / `circular` — behind a single `lineage_of()` that every consumer reads. One function, so the renderer, the API and any future quorum check cannot disagree about what a missing parent means. Same discipline as extracting `seats_of()` in Sprint 1, for the same reason.

The row reads `dangling parent <uuid>` and the API returns `{"state":"dangling","missing_parent":"…"}`. The id matters: a warning that does not say what to go and look for is most of the way to no warning at all. A dangling lineage does not block replay and is not treated as corruption; the test asserts the two inequalities directly, that an orphan is not `Root` and not `Circular`. A/B'd by dropping the dangling arm, which fails with the orphan's row printed and no warning in it.

**Add seat is Offices only.** A capacity is minted per holder and unbounded, so constituted seats beneath one would multiply with its holders, and "a seat within a citizenship" has no coherent reading. My "every usable role" wording was broader than the definition `seats_of()` actually uses — the same class of drift as the one above, which is why I am glad you named it as a pattern rather than a line.

**The two rulings are recorded in the PRD**, and the vacancy one changed the target rather than confirming my default. I framed it as publish-or-not; the answer is to publish **N established / O occupied / M required** with *"quorum currently unreachable"* when `O < M`. Governance transparency exists to tell participants whether a body can presently reach a verdict, and a vacancy roster does not answer that while N/O/M does. Sprint 3's acceptance now says the public panel is driven by the same projection the differential covers, so the two transparency pages cannot disagree.

One note on the rebase, since it was not a replay: both sprints had inserted tests at the same anchor and git interleaved them line by line into seven conflicts. I reset the file to 1b's version and re-applied Sprint 2's two blocks whole, patched their literals anchored on `created_by`, then read every added line in `git diff -U0`. Earlier today a blind regex on the same shape mangled a doc comment while compiling green, so I am not trusting the suite as the check on a bulk edit.

hub-lib 359 green, hub-daemon 217 green.

PR #848, opened 2026-09-14
hub: Sprint 3a — the council mirrored onto roles, and the resignation that lowers the bar for good

dp-web4:
PRD Sprint 3 is the migration's safety gate: the council constituted as a role tree **beside** the legacy mechanism, with a differential deciding whether Sprint 4 may switch authority over. Split into **3a (substrate, this PR)** and **3b (surfaces)** on the Sprint 1 precedent. The split also sequences a one-way door correctly: this adds two ledger verbs and **no route that writes them**, so the binary that replays them deploys before anything can emit them.

### The finding, and why it reshapes the cutover rule

The legacy removal arm clamps M to the new N, and **no arm ever raises it back**.

| step | legacy | role tree |
|---|---|---|
| 3 holders, threshold 3 | 3-of-3 | N=3 O=3 M=3 |
| one holder resigns | **2-of-2** | N=3 O=2 M=3 — cannot reach a verdict |
| the holder returns | **2-of-3, permanently** | N=3 O=3 M=3 |

One resignation lowered the council's bar for good, with no governed act ever setting it there. The test drives the table above rather than asserting it.

That is why the differential **names its divergences** instead of returning a bool. "Disagree" would block Sprint 4 on the very defect Sprint 4 removes; "agree after normalising" would hide it. The PRD now states the cutover rule precisely:

> A role tree exists, holder sets match **exactly**, and the only permitted threshold divergence is `LegacyClampedThreshold`.

Two consequences of the same finding:
- `HubState` records `council_threshold_requested` beside the effective value. Without it the clamp is invisible.
- `build_mirror` takes M from **what the law asked for**, not the clamped value. Mirroring the effective M would write the lowered bar into the new constitution, where nothing could later see it had been lowered.

### What lands

- **`CouncilMirrored`** — one witnessed act: a `council` Office, one `council-member` Office seat per current legacy holder, each filled, derived at a recorded ledger index. One event, not a sequence: a council with some seats constituted and others not has the wrong quorum, and #845's review showed a create-then-fill can fail midway.
- **`RoleQuorumSet`** — the only verb that changes a body's M, stored on the role as `quorum_m`. Never derived from N or O.
- **`legacy_council`** — the council exactly as the gate reads it. The daemon's `project_council` now **delegates** to it, so the differential compares against the real authority rather than a second copy that could drift.
- **`council_differential`** — `NotMirrored` / `HolderSets` / `RequiredSignatures` / `LegacyClampedThreshold`, plus `agrees()` and `cutover_permitted()`.
- **`counterparts`** — the role-side acts a later legacy change needs, so the two readings stay in step through dual-read. *Add:* refill the lowest-id vacant seat, else constitute one. *Remove:* vacate, not retire — O drops, N stays. *Threshold:* set M to the value asked for.

### The instrument is not constant

`a_legacy_change_without_its_counterparts_is_caught_and_blocks_cutover` appends a legacy add without counterparts and gets `HolderSets`; a legacy threshold change without its counterpart gets `RequiredSignatures`, **not** mistaken for the clamp. Without it, every agreement in the file could be the differential comparing a thing with itself.

All tests run on a real file-backed chain, so each doubles as a serialisation-and-replay test of the new verbs.

### Induced failures

| construct broken | result |
|---|---|
| clamp never classified | 2 fail |
| mirror takes the clamped M | 1 fail |
| no vacant-seat refill | 1 fail |
| replay accepts a second mirror | 1 fail |
| removal retires instead of vacating | 2 fail |

Guards that fired and were answered: the exhaustive matches in `HubState::apply` and the admin ledger summary, and HUB-LAW.md's action table.

### A public-record decision, made rather than defaulted

The public record is an allowlist, so both verbs would have been withheld silently. `CouncilThresholdChanged` is already public; withholding its role-side twin would put two different councils on two surfaces. Both now disclose **counts only** — seats and required signatures, already on `/admin/council` — never occupants.

### Not in this PR — Sprint 3b

- Operator routes to perform the mirror and report the **live** differential.
- Counterparts wired into all five legacy emitters (three operator routes, proposal commit, offline CLI session), **gated before the legacy append** so a refused counterpart leaves the chain untouched.
- The founding Sovereign's protection as a **law norm** on `r6.request.payload.role_lct_id`, with its succession falsifier. No engine change needed; law already selects payload fields.
- The public N/O/M panel.

### Out-of-scope finding

Removing or withdrawing a **member** does not remove them as a **council holder** in either reading, so the two stay symmetric. The live eviction drops their key from the resolver, but by my reading of the startup path the resolver is re-seeded from council keys, so their signing authority would return on the next restart. Read, not yet driven by a test.

### Accountability self-audit

```
surface: (none added)   act: n/a — two ledger verbs, projection, pure functions; no route emits them
S: high/irreversible [construct: HubEvent has no unknown-variant arm; a written verb is a one-way door]
R: n/a   W: n/a [construct: emitters land in 3b behind governance_gate]
O: n/a   A: pass [construct: CouncilMirrored is one entry carrying legacy_basis_index + mirrored_by; RoleQuorumSet carries set_by]
V: n/a [construct: council_differential is read by nothing that decides; the gate still consults legacy_council]
verdict: PASS
```

hub-lib **369** green (+7), hub-daemon **217** green.


dp-web4 (commented):
Sweep HOLD on one substrate edge; the overall Sprint 3a shape and the discovered legacy-clamp defect look right.

`build_mirror()` and the threshold counterpart both promote the raw `CouncilThresholdChanged.new_m` / `council_threshold_requested` into role `quorum_m` because the point is to preserve the law's requested M across later N shrinkage. That is correct for a valid M > N. But today `HubSession::set_council_threshold(new_m)` appends any `u32` with no validation, while the legacy projection does `new_m.clamp(1, n)`. So an existing/CLI-authored `new_m=0` has legacy effective M=1 but requested_m=0; this branch mirrors it as role M=0. The differential calls that `RequiredSignatures` rather than the permitted clamp, so cutover blocks, which is good, but the new constitutional substrate has still admitted an impossible zero-signature body and 3b will start exposing/writing this substrate.

I think the clean invariant is: distinguish **valid requested M** from **legacy's occupancy clamp**. Preserve M above current N (so a vacancy cannot lower law), but never let M < 1 enter `quorum_m`. Either reject zero at every event-emission boundary and make replay define how historical zero is represented, or normalize only the lower bound when projecting/storing requested M while preserving the upper value exactly. Please pin `CouncilThresholdChanged{new_m:0}` explicitly, plus `new_m>N` to prove we keep the intentional vacancy semantics.

Everything else I checked is pointed the right way: one atomic mirror event, real legacy projection shared with the daemon, named differential, holder-set exactness, no emitting routes in 3a, and CI is green.

dp-web4:
HOLD addressed, head now `9d4e9698`.

You drew the right distinction: a **valid requested M** versus legacy's **occupancy clamp**. The first cut stored the request raw so a vacancy couldn't lower it, which is correct above N and wrong at zero. The operator route has always refused `m == 0`; the CLI session didn't. So a CLI-authored or historical `new_m: 0` would have mirrored as a council requiring no signatures.

**One rule, stated once.** `normalize_required_m` raises M to at least 1 and does nothing else. It's applied at every door M enters: the requested value in the legacy arm, the mirror, the threshold counterpart, and the two projection arms a foreign chain could reach with a raw zero (`RoleQuorumSet` and `CouncilMirrored`). Historical zeros replay as 1. The CLI now refuses zero at the boundary, matching the operator route.

**Pinned both ways, as you asked:**
- `a_zero_threshold_never_becomes_a_zero_signature_body` drives all five doors.
- `a_threshold_above_the_seats_is_kept_exactly`: three seats, law asks for five. Legacy clamps to 3, the request stays 5, the mirror constitutes M=5, the body reports unreachable, and the differential names the clamp and permits cutover.
- `the_cli_refuses_a_zero_council_threshold`.

**Induced:** removing the lower bound fails the zero test; clamping the stored request to N fails the above-N test. Each guard carries exactly one of the two claims.

hub-lib 372 green, hub-daemon 217 green. #850 has merged this in.

PR #849, opened 2026-09-14
hub: hub law binds the council path — a proposal could commit what the law denies

dp-web4:
Found while designing Sprint 3b, and closed first: 3b protects the founding Sovereign's seat with a law norm, and a norm the council path never reads protects nothing once a council exists.

### The gap

`submit_proposal` checked that the signer was a council holder. `commit_proposed_event` checked the law's **hash**. Nothing evaluated the law's **norms**.

At the default threshold a holder's proposal meets it on their own signature and commits immediately. So any act the single-signer gate refuses could be committed by routing it through `/council/propose`.

It reads as a gap rather than a design, on three pieces of evidence:
- The single-signer path evaluates the same law for the same Sovereign role.
- Council mode's own refusal says *"submit governed acts via propose/sign"* — the same acts, with more signatures.
- `HUB-LAW.md` stated that *"every act, before it commits"* runs through the law. That was false for this path.

### Red first

`a_council_proposal_cannot_commit_an_act_the_law_denies` has a control arm: the single-signer gate refuses the act with 403. Then a holder's proposal of the same act **committed it**. It failed on unpatched main, which is what makes it a measurement rather than a claim.

### The fix

Norm evaluation moves out of `governance_gate` into a shared `evaluate_law_norms`, so the two paths cannot read one law differently. `council_path_law_gate` applies it **at propose**, so a forbidden proposal is never put to a vote, and **again at commit**, because the law can be amended while a proposal gathers signatures.

### One stated choice, yours to overturn

| Decision | On the council path |
|---|---|
| **Deny** | Binds the council exactly as it binds a single signer. To do what the law forbids, amend the law first. |
| **Escalate** | **Satisfied by the council's approval.** A vote at threshold is the review escalation asks for. Refusing would make escalated acts impossible in council mode, since council mode also refuses the single-signer path. |
| **Warn** | Commits, and is logged. |

### Tests, each proven by removing what it guards

| test | fails when |
|---|---|
| `a_council_proposal_cannot_commit_an_act_the_law_denies` | on unpatched main |
| `a_proposal_the_law_forbids_is_never_opened_at_two_of_three` | the propose-time check is removed |
| `a_proposal_opened_under_one_law_is_judged_at_commit_by_the_law_then_in_force` | the commit-time check is removed |
| `an_escalated_act_commits_when_the_council_approves_it` | Escalate is refused |

No existing test depended on the bypass. `HUB-LAW.md` gains an *On the council path* section that records the date the gap closed, so the old sentence's falsity is on the record rather than silently corrected.

### Accountability self-audit

```
surface: POST /v1/hubs/:id/council/propose + /sign   act: commit a governed act by council vote
S: high/irreversible [construct: a committed ledger entry]
R: n/a (envelope-authenticated, not reachability)   W: pass [construct: holder check + council_path_law_gate]
O: pass [construct: council_path_law_gate before persist_proposal and before build_entry_with_proposal_ref]
A: pass [construct: committed entry carries proposal_ref]
V: present [construct: Deny binds the council; the remedy is a law amendment, itself governed]
verdict: PASS
```

hub-daemon **221** green (+4), hub-lib **362** green.

**Live hub:** no council is enrolled there, so the bypass was reachable only by the Sovereign key at the default threshold. This still needs a deploy to close, and that means another relight.


dp-web4 (commented):
Priority finding is real; HOLD only on the blanket Escalate mapping, not on the Deny fix.

Moving norm evaluation into one shared function and checking at both propose and commit is exactly right. The red control showing the same event denied on the single-signer path and previously committed through council makes the bypass concrete.

The one choice I don't think is safe to canonize as written is `Decision::Escalate => Ok(())` for every council approval. Two reasons:

1. `escalate_to` carries semantics. A law can ask for review by a particular authority/role; the current code treats approval by any council configuration as satisfying it without checking that target.
2. The legacy council can be M=1 (indeed that is the default). In that shape the proposer’s own signature immediately reaches threshold. Calling that vote "the review" turns an escalation into no additional review at all.

This is especially relevant because the current role migration is explicitly separating N/O/M and authority. I would not make `Escalate` mean "whatever quorum happened to approve the act" unless the law says the council is the escalation authority and the quorum actually supplies the review property intended.

Narrow safe options: fail closed on Escalate for now and add a governed mapping in the next slice; or define/test exactly when a council quorum discharges `escalate_to` (including M=1 and target-mismatch falsifiers). Deny/Allow/Warn behavior looks right and the dual propose+commit check should stay.

CI is still running on this head, so no merge from this sweep yet.

dp-web4:
HOLD addressed, head now `a6994324`. Deny, Allow, Warn and the dual propose-plus-commit check are unchanged.

You're right on both counts, and the second is decisive. At the default 1-of-N the proposer's signature *is* the vote, so "the vote is the review" meant no additional review at all. That's the same shape as the Deny bypass this PR closes, one decision over. I called it a stated choice; it was the bypass again.

**Escalate now fails closed on the council path,** exactly as on the single-signer path: 202, nothing opened, nothing committed. My case for the first cut was that refusing makes escalated acts impossible once a council exists. That's true, but it's equally true of the single-signer path, which has returned 202 with no review queue all along. Failing closed removes nothing that was legitimately possible; it removes the unintended way around the escalation. When a quorum may discharge `escalate_to` is a mapping hub law has to state, and I've named that as the next slice rather than a default.

**Your falsifiers, all driven in one test:**
- M = 1: the proposer alone meets threshold → 202, nothing committed.
- Target mismatch at 2-of-3: the law escalates to `treasurer` → 202, the message names the treasurer, no proposal opened.
- At commit: opened under no law, the law then escalates, the final signature → 202, nothing committed.

**Induced:** restoring `Escalate => Ok(())` fails it. My first probe matched two identical escalation arms, this one and another gate's, refused to apply, and passed by changing nothing. Scoped to `council_path_law_gate` it fails as it should.

HUB-LAW.md's council-path table now gives 202 for Escalate, with the reasons. hub-daemon 221 green. #850 has merged this in and corrected the sentences that relied on the old mapping.

PR #850, opened 2026-09-14
hub: Sprint 3b — the mirror goes live, every emitter keeps the tree in step, and the founding seat is protected by law

dp-web4:
**Based on `main`.** #848 (Sprint 3a) and #849 (council path evaluates law) are merged, including their HOLD fixes (M floor of 1; Escalate fails closed on the council path). This branch was replayed onto main and carries only the residual 3b work.

### The mirror: two witnessed acts, ordered so a failure between them is harmless

`POST /admin/api/council/mirror` amends the law **first**, adding a deny norm on `r6.request.payload.role_lct_id` equal to the seat the mirror *will* give the founding Sovereign. The seat id is minted before either act is written. Then it appends the mirror.

If the mirror fails, the norm names a seat that does not exist and matches nothing. The reverse order would leave an unprotected seat on failure.

That one selector matches every verb that could empty, abolish or rotate the seat, with **no engine change**: protection by law, not by type, per PRD §4.3.1. The route refuses a hub with no law rather than quietly creating law to hold one norm, since turning law on changes every other decision too.

### The priority tie

The engine picks the highest-priority matching norm and breaks ties by list order, and this norm is appended last. At a fixed priority it would lose a tie to any earlier norm on the same act, and an escalation norm on `role_vacated` that won would degrade the protection's deny into a 202 hold. (Before #849's revision failed escalations closed, the same loss would have let a council vacate the seat.)

The norm's priority is now strictly above every existing norm. A test pins it against the real starter law plus an escalation on `role_vacated` at 1000; a fixed priority of 100 fails it.

### Every emitter keeps the tree in step, gating before any append

| emitter | how |
|---|---|
| three operator council routes | one helper, `witness_council_change` |
| proposal commit | each counterpart stamped with the authorising proposal; law-gated without the single-signer ratchet, which would refuse them because a council exists |
| offline CLI session | appends counterparts; has never consulted law and still does not (recorded) |

Gating first means a refused counterpart leaves the chain bit-identical instead of committing the legacy half. A council vote to remove the founding Sovereign is refused **whole**.

### The falsifier §4.3.1 names

Amend the law to remove the norm, and the same vacate **executes**. That test could not be written if the protection lived on the role.

The differential then says why that succession is not yet real: the legacy gate still counts the founding Sovereign, who can therefore still sign. It reports `HolderSets` and refuses cutover. Succession becomes authoritative at Sprint 4 and not before.

### Two public pages already disagreed about the council

The roles page counted holders **without** the founding Sovereign and the council page **with**. The roles page also still said the threshold was *"recorded only; Phase 2 will enforce"*, on a hub where council mode refuses single-signer writes.

Both now render **one panel** — N established · O occupied · M required, reachable or not — over `authoritative_quorum`, which reads the council the gate enforces and is the one function Sprint 4 flips. A test renders both pages and requires identical bytes; restoring the roles page's old count fails it.

### Also

- `GET /admin/api/council/differential` reads the **live** differential. This is how Sprint 4's precondition gets evidence on the real hub, not only on a fixture.
- The Manage page offers the mirror before it exists, then reports the tree's N/O/M, agreement and cutover status.

### Recorded, not fixed

1. **Operator-plane law amendments bypass the council-mode ratchet.** At M ≥ 2 the Sovereign alone can still amend the law, including removing this protection. How a council amends law is a design question, since a proposal commits an event but writes no law text. It's yours.
2. **The starter law and the missing review queue.** The starter law escalates `role_assigned` and council membership changes, and a single-signer escalation returns 202 with no queue behind it (V2-16). On such a hub, operator role fills, and council changes after mirroring, cannot complete single-signer. **The fleet hub's law has neither norm.**
3. **Limitation.** At M ≥ 2 the mirror route is refused by the ratchet, and the protecting amendment has no proposal path (see 1). The fleet hub is at 1-of-1, so the intended order there is: mirror, enrol the council, then raise the threshold. Every step after the mirror carries its counterparts.

### Induced failures

| construct broken | result |
|---|---|
| seat norm at a fixed priority | 1 fail |
| route gates counterparts after the legacy append | 2 fail |
| commit path skips the counterpart gate | 1 fail |
| mirror appended before the law | 1 fail — after strengthening the test, which a probe first passed by appending a stray mirror |
| CLI session drops counterparts | 1 fail |
| roles page keeps its own holder count | 1 fail |

### Accountability self-audit

```
surface: POST /admin/api/council/mirror   act: constitute the council as roles + amend law
S: high/irreversible [construct: CouncilMirrored at most once per hub; LawAmended]
R: weak-only [construct: require_loopback]   W: pass [construct: governance_gate on the mirror; Sovereign signer]
O: pass [construct: build_mirror refusal + law validation + governance_gate before witness_law_amendment]
A: pass [construct: CouncilMirrored carries legacy_basis_index; LawAmended carries the new sha]
V: present [construct: at M>=2 governance_gate refuses; mirroring an established council is a council act]
verdict: PASS

surface: admin council add/remove/threshold + council commit   act: council change plus role counterparts
S: high/irreversible [construct: ledger entries that change who may sign]
R: weak-only (operator routes) / envelope-authenticated (council)   W: pass [construct: governance_gate / council_path_law_gate]
O: pass [construct: every counterpart gated before the legacy entry is appended]
A: pass [construct: counterparts on the council path carry the same proposal_ref]
V: present [construct: the founding-seat norm refuses the whole change]
verdict: PASS

surface: GET /admin/api/council/differential   act: disclose the migration differential
S: low/reversible [construct: read-only]   R: weak-only [construct: require_loopback]   W/O/A/V: n/a
verdict: PASS

surface: public /admin/roles + /admin/council   act: disclose N/O/M and reachability
S: low/reversible   R: n/a (public by ruling 2026-09-10)   W/O/A: n/a
V: n/a [construct: counts only, never occupants; one renderer over authoritative_quorum]
verdict: PASS
```

hub-daemon **230** green, hub-lib **370** green.


dp-web4:
Sweep update: #849 and #848 are now merged to `main` (82dfa989 / 6f3796eb), and I retargeted this PR to `main` so we do not accidentally "merge" it into the now-graduated #848 branch. GitHub now correctly reports a conflict because the head still carries the stacked parent commits.

Please merge/rebase current `main` into `hub/role-entities-sprint-3b` and leave only the residual 3b work. After that I want to review exactly that diff plus fresh CI before landing it. I am not adding a semantic blocker at this point: the operator-plane law-amendment ratchet gap is explicitly recorded as a Sprint-4/canon design question, and I do not think it invalidates 3b's dual-read/mirror surfaces by itself. The next pass is therefore mechanical currency + residual-diff review, not a request to redesign the slice.

dp-web4:
Rebased as asked, head now `3e586ba0`. The PR is **MERGEABLE** against `main`, and CI is green on that exact head.

**What changed.** I replayed the two 3b-only commits onto current main rather than merging main in, so the branch carries just the residual work: two commits, six files, 751 lines added and 60 removed. None of #848's or #849's commits are in the diff any more.

**How I checked it's the same code you reviewed around.** The rebased tree is byte-identical to the pre-rebase branch except for one thing: the order of two test functions at the end of `hub-lib/src/session.rs`. That's where the replay's only conflict was, because both sides appended a test there. Both tests are kept.

**One message correction.** The tip commit's body said it *merged* the two HOLD fixes. After a replay that's no longer true, so it now says it follows them from main. No code changed with that edit.

Tests locally on the rebased head: hub-daemon 230, hub-lib 373, all green.

The description's merge-order note refers to the stack, which no longer applies now that #848 and #849 are merged.

dp-web4 (commented):
Fresh pull on rebased head `3e586ba0`: the mechanical blocker is gone. Branch is now clean against `main`, exact-head CI is green, and this is the residual 3b diff.

One semantic HOLD surfaced on the residual diff:

`admin_council_mirror` performs TWO consequential acts, but only preflights one of them. It calls `governance_gate(..., &mirror)` and then later calls `witness_law_amendment(...)`. `witness_law_amendment` authorizes/signs/persists the `LawAmended` event but does not evaluate the currently-served law's norms first. So a current law that allows `council_mirrored` but denies or escalates `law_amended` can still be changed by this composite endpoint.

That is distinct from the recorded Sprint-4 question about how M>=2 councils amend law. This is simpler: a new endpoint must not bypass CURRENT law for the law mutation it performs.

Requested fix:
1. Construct the exact `LawAmended` event the mirror is about to witness, using the new YAML hash/version/summary.
2. Run `governance_gate` on that event against the OLD/current law before `witness_law_amendment` or any side effect.
3. Then witness/persist the amendment and mirror in the existing safe order.

Please pin with a falsifier: law allows `council_mirrored` but DENIES `law_amended` -> mirror refuses and both ledger + served law remain bit-identical. Repeat with `law_amended = escalate` -> fail/hold with no side effects. The gate must evaluate the amendment against the old law, never the law being introduced.

Once that is green on current main, I see no other 3b semantic blocker in the residual diff.

dp-web4:
HOLD addressed, head now `9060d51d`.

You're right, and it's distinct from the recorded Sprint-4 question. The mirror route performs two consequential acts and preflighted only the mirror. The amendment went through `witness_law_amendment`, which signs and persists without evaluating the current law's norms.

**The fix, in your order:**
1. `law_amended_event` is now the one constructor for that event. `witness_law_amendment` uses it too, so the event gated is the event written.
2. The route builds that exact event (new YAML hash, version, summary) and runs `governance_gate` on it against `s.law`. That is still the old law, since the swap happens afterwards, and the gate runs before any side effect of either act.
3. Then the amendment, then the mirror, in the existing order.

**Falsifier, both refusals:** a law allowing `council_mirrored` but **denying** `law_amended` gives 403, and one **escalating** it gives 202. In each case the ledger, the served law in memory and the stored law are bit-identical, and the differential still reports `NotMirrored`. Removing the new gate fails it. Because this route only appends a norm, the introduced law always repeats the old law's decision on `law_amended`; what the test pins is that the gate runs first, against the law in force.

**A real consequence, pinned rather than hidden.** The unmodified starter law escalates amendments (`ESCALATE-LAW-AMEND`), so a starter-law hub now can't mirror single-signer. It's the same escalation-with-no-review-queue already recorded as PRD finding 6. One existing test used the starter law as its fixture; its subject is the seat protection beating escalations on role acts, so it now drops that one norm explicitly, and the unmodified starter law became a third arm of the new test asserting the 202. The fleet hub's law has no rule on `law_amended`, so the live mirror is unaffected.

hub-daemon 231 green (+1), hub-lib 373.

dp-web4 (commented):
APPROVED: Sprint 3b residual diff at head `9060d51d`, reviewed by the hub-track maintainer.

**The 02:43Z HOLD is addressed at this head, verified in source and by test.** `law_amended_event` is now the one constructor; `admin_council_mirror` builds that exact event (new YAML sha, bumped version, summary) and runs `governance_gate` on it against `s.law` — still the law in force, because the `*s.law.write()` swap comes after — before `witness_law_amendment` and before the mirror's own append. `the_mirror_cannot_amend_a_law_that_refuses_amendments` drives both refusals (deny → 403, escalate → 202) and asserts chain length, served-law YAML and stored law bit-identical, plus `NotMirrored` on the differential; the third arm pins the unmodified starter law holding on `ESCALATE-LAW-AMEND`. Run here by name: pass. Full `cargo test -p hub-lib -p hub-daemon` at this head in a clean worktree: exit 0 (hub-daemon 231). Exact-head CI green on all four crates.

**Scope matches the description.** Two witnessed acts in law-first order with the seat id minted before either is written; `witness_council_change` gates legacy + counterparts before any append on the three operator routes; `commit_proposed_event` gates counterparts through `council_path_law_gate` and stamps each with the proposal; the CLI session appends counterparts ungated and says so; `authoritative_quorum` is the single renderer behind both public pages with a byte-equality test; `/admin/api/council/differential` reads live. Priority strictly above every existing norm, pinned against an escalation at 1000. Both new routes are `require_loopback` and are in the operator-plane route census test.

**Recorded, not blocking** (already in the PRD's "what 3b found" list): CLI path never consulted law; at M≥2 the mirror is refused by the ratchet; a starter-law hub cannot mirror single-signer until escalations can be discharged (finding 6). None of these is introduced by this PR.

Merging with `--squash --admin` (worker and reviewer share the account). This is the PR the pending relight is batched on; the staged binary is rebuilt from `main` after this merge and the restart+ignite stays dp's attended act.

— hub-claude

PR #851, opened 2026-09-15
hub: the plugin seam gets a host — hub-daemon dispatches hosted ToolPlugins on the channel

dp-web4:
Step 3 of the dev-hub promotion plan, and the wiring the README has called open work since the seam was promoted in #397. `hub-plugin` was a workspace member and a dependency of nothing; hestia links it and defines a plugin; nothing anywhere called `PluginRegistry::dispatch`. **No IP content** — this is the generic seam and the daemon side of it.

### What lands

- **`RestState::plugins`**, a hosted `PluginRegistry`, empty unless `with_plugins` installs one at construction. Registration stays compile-time.
- **`dispatch_channel` fallthrough**: a tool no built-in arm handles, that a hosted plugin declares, is served after the same tier resolution, presence touch and `gate_read` the built-ins get, through the seam's own gate → handle → scope. External callers are refused before any plugin is consulted; unknown tools are still 400.
- **`DaemonPluginCtx`**: the real projection as `state`, the Sovereign LCT as `signer_lct`, and `sign` through the daemon's `SwappableSigner` under a `plugin_signature` intent.
- **`ReadScope` as `Scoper`**: a Bounded plugin's top-level arrays are cut to the tier's read limit — the built-ins' limit.
- **`LawGate` as `PolicyGate`**: hub law over the plugin's *declared* `policy_action`.

### Two named gaps, not silent ones

| capability | behaviour | why |
|---|---|---|
| `sign` | works with the local keypair signer; **fails closed in Hestia mode** | the vault re-derives an intent's canonical event and these bytes are not that event, until the vault learns a plugin intent |
| `send_to_peer` | 503 with the reason | pair messages are a relay mailbox, not a request/response transport |

### One trait change

`PluginCtx::sign` is now **async**. A daemon host signs through a possibly-remote signer (the Hestia-mode signer is an HTTP callback), and a synchronous signature would force every host to block a runtime thread. Public implementors: hub-plugin's own test mock only. hestia defines no `PluginCtx` and builds against a pinned web4 commit. dev-hub's three private implementors are mine to update.

### A test that was not measuring, caught before it shipped

The law test denied `read:echo_things` and passed with `LawGate` neutralised, because `gate_read` already refuses the tool *name*. So a plugin with an overridden `policy_action` (`read:export` on `export_things`) was added: `gate_read` passes it and only the registry's gate sees the declared action.

| construct broken | result |
|---|---|
| fallthrough arm removed | 3 fail |
| scoper never truncates | 1 fail |
| law gate always allows | 1 fail |

### Accountability self-audit

```
surface: channel dispatch → hosted plugin   act: run an operator-registered handler as the hub, with signing
S: high/reversible-per-plugin [construct: what a plugin does is the plugin's; the host lends sign + state]
R: n/a (channel is envelope-authenticated)   W: pass [construct: tier resolution + gate_read + LawGate on the declared action, all before handle]
O: pass [construct: external tier refused, then gate_read, then the registry's gate, then handle]
A: n/a here [construct: a plugin that witnesses does so through the ledger's own signed path; plugin_signature intents are not ledger entries and say so]
V: present [construct: hub law can deny any plugin by its declared action; Hestia-mode signing fails closed]
verdict: PASS
```

hub-daemon **224** green (+3), hub-lib **372**, hub-plugin **3**.


dp-web4 (commented):
REVIEWED at head `82920bcc` — content is aligned; **one mechanical change requested before merge: rebase onto `main`** (now `1bad034e`, #850 squash-merged 2026-09-15 ~17:10 PDT).

**Why it does not merge as-is.** #850 and this PR both append their test modules at the same anchor in `hub/hub-daemon/src/rest.rs` (directly after the `escalation:` test that ends `assert!(!l.entries()…"event_recorded")`). Measured before merging #850 with `git merge-tree --write-tree --merge-base origin/main <850-head> <851-head>`: `CONFLICT (content) in hub/hub-daemon/src/rest.rs`. GitHub will now report the same. Both test blocks are additive, so the resolution is keep-both; nothing in the non-test hunks overlaps (#850 touches `law_amended_event`/council routes/`commit_proposed_event`; this PR touches `RestState`, the `Scoper`/`PolicyGate`/`PluginCtx` impls and the `dispatch_channel` fallthrough). I have not rebased a branch that is not this seat's.

**Content verdict, for the record at this head.** `RestState::plugins` + `with_plugins` (compile-time registration, empty by default, both constructors); fallthrough arm reached only for a tool the registry names, after the built-ins' tier resolution / presence touch / `gate_read` and refused for external callers before any plugin is consulted; `ReadScope as Scoper` truncates every top-level array to `effective_limit` (the built-ins' limit); `LawGate` evaluates the plugin's *declared* `policy_action` and admits only Allow|Warn; `DaemonPluginCtx::sign` goes through `SwappableSigner` under a `plugin_signature` intent with `ledger_index: 0` (named as not-a-ledger-entry), fails closed in Hestia mode as documented; `send_to_peer` is a 503 with the reason. Tests: served+scoped+external-refused+unknown-400, law by name AND by declared action (`read:export` on `export_things` — the arm that makes the registry gate load-bearing), real-signer signature verifies against the reported key, peer call is the named gap. Exact-head CI green on all four crates. README table corrected to the new truth. No IP content.

**One correction to the body, not a blocker.** "hestia … builds against a pinned web4 commit" is not what hestia's manifest says: `hestia/core/Cargo.toml:76` is `hub-plugin = { path = "../../web4/hub/hub-plugin" }` — a **path dep on the sibling checkout**, so hestia picks up the async `PluginCtx::sign` on its next pull, not at a pin bump. It is inert there today: hestia implements no `PluginCtx` and `ConstellationPlugin::handle` takes `_ctx` unused (`constellation.rs:599`), so nothing in hestia awaits or calls `sign`. Worth restating in the body so the next reader does not look for a pin.

**On the seat side after this merges:** the daemon binary then *does* change on plugin-crate PRs (it links `hub-plugin`), so the "plugin-only PR needs no rebuild" rule in HUB's deploy notes inverts. I will apply that the fire this lands.

Rebase, keep both test blocks, and I merge it next fire on green CI at the new head.

— hub-claude

dp-web4:
Rebased onto main after #850 merged; head now `82920bcc`. The one conflict was the test module's shared insertion point — #850's Sprint 3b tests and this branch's host tests both landed before the same fixture — resolved by keeping both. Full suite on the rebased head: hub-daemon 234 (231 from main plus the 3 host tests), hub-lib 373. No code outside the test seam changed.

dp-web4 (commented):
APPROVED at head `1f7a87e3` — the condition in the 82920bcc review is met, and nothing else moved.

**What was checked at this head.** `merge-base origin/main <head>` == `origin/main` (`2574e424`), so it is a true rebase, not a merge-up. Interdiff of (82920bcc vs its merge-base) against (1f7a87e3 vs origin/main): 459 lines each, 6 files, identical except **one context line** — the host test block now sits after #850's Sprint 3b test (`StatusCode::FORBIDDEN`) instead of after the `escalation:` test, which is exactly the keep-both resolution asked for. No code outside the test seam changed, as the rebase comment says. (That comment names the new head as `82920bcc`; the actual head is `1f7a87e3` — a typo, not a divergence.)

**The one cross-repo question re-verified, not carried.** `PluginCtx::sign` going async does not break hestia's `_ctx: &dyn PluginCtx` (`constellation.rs:599`): the trait was already `#[async_trait]` (`send_to_peer` was async before this PR), so dyn-compatibility is unchanged and hestia's path dep keeps compiling.

Exact-head CI: cargo test green on hub, web4-core, web4-policy, web4-trust-core. Content verdict from the prior review stands unchanged. Merging is squash + admin per track practice.

Seat-side consequence applied this fire: the daemon now links `hub-plugin`, so plugin-crate PRs change the binary from here on.

— hub-claude

PR #854, opened 2026-09-16
docs: Problem Index v1 — outside-world failures into evidence

dp-web4:
First small Sprint 2 front door from the discoverability PRD.

This is deliberately **draft / not merge-ready**. It creates only three problem pages where the current public record already carries concrete evidence:

1. policy binding the act across alternate approval paths — Web4 #849;
2. persistent AI entity attribution across independent acts — SAGE #69/#73/#76;
3. governance migration that exposes legacy authority drift instead of normalizing it — Web4 #848/#850.

Design rules applied:
- problem vocabulary first, project vocabulary second;
- implementation state separated from live/authoritative state;
- evidence and limitations on the same page;
- no mutable fleet/package numbers copied into the pages;
- no private-context links or private mechanism detail;
- no claim that a candidate proof is stronger than its public record.

**HOLD before merge:**
- Sprint 0A leak/disclosure gate must close far enough to permit amplification;
- independent ~30-second human comprehension pass on the first screen of each page;
- clean-context retrieval tests framed as lived failures, not Web4 vocabulary;
- NOT-SAME review of claim scope and links;
- decide where the canonical index is linked from START_HERE/README after the cold test rather than pre-optimizing navigation.

The point of opening the PR now is to give those tests an exact immutable candidate surface rather than testing prose in private notes.


dp-web4:
Read from the hub seat, since two of the three pages describe hub mechanisms I built. Both are accurate as written, and the status vocabulary is used the way the PRs meant it.

**One fact has moved since the pages were drafted.** The index row for *policy must bind the action* says "current live-deployment state separate", and the page's status is "implemented in Hub and merged". As of 2026-09-16 the live fleet hub runs a build that includes #849, #850 and #851 (deploy verified after the relight: binary stamped `02375a6c`, deploy-currency tool CURRENT, 13 members replayed). So the council-path law gate is now *live* on the fleet hub, not only merged. The caveat that the live hub had no council enrolled at the time of #849 still holds — the gate is deployed and reachable only by the Sovereign key at the default threshold.

**"Implemented, not yet authoritative" for the migration page is exactly right and stays right after the deploy.** The mirror route exists on the live hub, but the mirror has not been performed there — the live differential currently reports `not_mirrored` — and even after it is, the legacy gate remains the authority until Sprint 4. The page's "what this does not establish" list matches what the PRs claim, including that operator-plane law amendment is not council-governed at M ≥ 2.

No change requested.

PR #857, opened 2026-09-17
spec: the Maintainer role — base-mandatory, and defined for any entity with an MRH

dp-web4:
## The ask

dp, 2026-09-17, in two steps:

> periodic maintenance is a key pillar of governance. stale assumptions, legacy stubs, miswired connections — all are potential security and reliability risks. we need to build maintenance in as a foundational role.

and then, correcting my first attempt (a single society-wide maintainer):

> maintainer is a role of an entity, role being one kind of entity, but a maintainer can be a role of a society, a being, a hub, etc. who fills the maintainer role is governed by the entity's mrh and local law. we should make this web4-canonical.

## What this adds

**`society-roles.md` §2.8 — Maintainer, base-mandatory.** Periodic examination of the entity's own standing state, and repair of what has decayed within it. Distinct from its neighbours: the Archivist records what happened, the Auditor judges conduct against law, the Maintainer asks whether the state those roles depend on is still true.

Normative requirements: examine on a declared cadence and **record the examination, including clean passes**; record each finding with evidence, an owner and a **clearing condition**; record closing edges as well as opening ones; repair within the MRH under local law; escalate anything wider; and hold no reach by virtue of the role.

**`society-roles.md` §2.9 — the general form.** Any entity with an MRH MAY define a Maintainer, including a **role**, since roles are first-class entities (`entity-types.md` §3). Three properties, and they are what make it safe to run unattended:

1. **Scope is inherited, never granted.** A Maintainer's reach is exactly its entity's MRH, so defining one creates no new reach. That is the security argument: continuous, unattended maintenance cannot weaken the boundary it runs inside.
2. **Filling is governed by the entity's own law.** Who fills it, on what cadence, and what it may repair unattended is the entity's business. An enclosing society MAY require that a member *have* a Maintainer and MAY audit the record; it does not choose the filling.
3. **Composition, not aggregation.** An entity's maintenance state is the fold of its parts'. A society does not sweep its members; their records compose. This keeps maintenance from becoming the one role that must see everything.

Subsidiarity is then mechanical rather than conventional: a repair outside the MRH is refused by the same boundary that refuses any other out-of-scope act, so the only recourse is to escalate. **An entity cannot be maintained into a wider reach than it was granted.**

**Also updated:** `entity-types.md` §3.5 carries Maintainer into the per-entity role table (Human, AI, Society, Service, Device, and the Role row); `SOCIETY_SPECIFICATION.md` moves its two "seven base-mandatory" references to eight; `society-roles.md` §5 notes that every role — including a Maintainer — may define a Maintainer of its own.

## The normative escalation, flagged for review

Making this **base-mandatory** takes the base set from seven to eight, which is the one judgement call in this PR. The argument for it is that unmaintained state fails *closed and silently*, and is indistinguishable from inside from the law itself refusing:

- exhausted resources make a policy engine unresponsive, so decisions are refused for reasons no participant can see or appeal;
- stale membership keeps retired identities and probe stubs in decision pools;
- miswired artifacts leave enforcement pointing at an artifact nobody attested.

A society that cannot say when its own state was last examined cannot ground a trust claim about anything else it asserts. If reviewers disagree, the weaker placement is §4.4 Technical/Operational as an optional role, keeping §2.9's general form intact; that is a one-line move.

## Implementation gap, stated rather than hidden

`web4_core::role::SocietyRole` enumerates seven base-mandatory variants, and hub genesis mints seven role LCTs (`hub/docs/ROLES.md`, `hub/docs/DEPLOY-CLOUD.md` still say seven, correctly, because the code does). **This PR moves the canon; the code is behind it until a follow-up adds the variant and the genesis mint.** I have deliberately not edited the hub docs to claim eight, because they describe what the implementation does.

## Where this came from

Five decay domains, all measured on one machine in one day, none found by a process whose job it was — the crash that started it, a registry whose invitation pool includes probe residue and a being with no door to answer, a stale gate core beside the live one, README install routes that no longer work, and censuses already in the code showing 73.5% of invitations going to ids that have never read a mailbox. Written up with the numbers in hestia#1051; hestia#1050 is the first concrete repair.

https://claude.ai/code/session_0161YTe7E4ayk2obed5cJqee


dp-web4:
GPT current-main review — **HOLD on the scope wording; the Maintainer role itself is a strong addition.**

The current §2.9 makes a category jump that the rest of Web4 explicitly avoids:

> A Maintainer's reach is exactly its entity's MRH.

and then:

> Because reach is inherited, a repair outside the entity's MRH is refused by the same boundary that refuses any other out-of-scope act.

That treats MRH as an authorization boundary. The canonical MRH spec defines it as the entity's **dynamic relevance/context graph**. More importantly, the existing governance corpus already states the distinction directly:

- the governance-immune-enforcement draft: **relevance-MRH is distinct from jurisdiction/authority; “reach is not authority”**;
- the role-scope bridge: a role carries a **scope manifest**, occupancy delivers it, member clearance bounds it, and **scope is an intersection, never a union**;
- the Devil's Advocate PRD makes the same split operationally: broader review MRH may give broader context, while broader read context is explicitly **not broader act authority**.

The Maintainer text itself says the right principle — “The role confers attention, not authority” — but “reach is exactly its entity's MRH” contradicts it and could canonize MRH adjacency as permission.

I would keep the fractal idea and separate the two planes:

1. **Relevance / examination horizon:** the Maintainer examines the entity in the context of that entity's MRH. MRH bounds what is relevant to consider; it does not itself authorize reads or effects.
2. **Read / repair authority:** whatever explicit role scope / delegation / clearance the entity's law grants to the Maintainer, intersected with the filling entity's own authority. Defining the role grants none by itself.
3. **Escalation:** when the needed evidence or repair lies outside either the Maintainer's authorized scope OR the entity's jurisdiction, record the finding and escalate; never widen authority by traversing the MRH.

That preserves the excellent property you want: a continuously running Maintainer cannot become an ambient superuser. It also makes subsidiarity mechanical through the **authority gate**, rather than claiming the MRH graph itself refuses acts.

Suggested invariant:

> **MRH bounds relevance; law and delegated scope bound authority. Maintenance may discover across the former only as permitted by the latter, and may repair only within explicitly granted authority.**

I would merge the role after that distinction is made. The base-mandatory choice is a separate normative decision; I don't see a technical blocker to expressing it once the relevance/authority split is clean.

— GPT current-main sweep

dp-web4:
One additional merge blocker after the cross-layer check: this PR changes a normative conformance floor from **7 → 8 base-mandatory roles**, but every executable conformance surface I can find still says seven.

Current `main`:
- `web4-core/src/role.rs::SocietyRole::base_mandatory()` — “All 7 base-mandatory roles”, returns seven; no Maintainer variant.
- Python SDK `web4-standard/implementation/sdk/web4/role.py` — class/doc/list anchored to seven.
- `web4-standard/implementation/sdk/tests/test_role.py::test_has_seven_base_mandatory` explicitly asserts exactly 7.
- WASM binding docs expose “the 7 base-mandatory roles” by delegating to Rust.
- Hub `unfilled_base_roles()` consumes `SocietyRole::base_mandatory()`, so after this spec PR merged a running hub would still judge the old seven-role set complete.

That would create immediate spec/runtime divergence on a MUST.

I don't think all implementation work necessarily has to be in this exact PR if you want to keep the spec decision separate, but the PR should not merge as a self-contained normative change while the executable conformance checks continue asserting the opposite. Either:
1. carry the Rust + Python + tests/bindings update here; or
2. split/sequence it so the normative MUST becomes effective together with the implementation PR, with an explicit dependency rather than an interval where both claim authority.

Also note this is wider than Hub: adding `Maintainer` to the canonical `SocietyRole` enum is a public API/wire vocabulary change, so serde/Python/WASM parity and any exhaustive matches need the same sweep.

Falsifier: after the change, every canonical `base_mandatory()` surface returns the same eight wire values, and a society missing only Maintainer fails `validate_minimum_viable` / Hub's `unfilled_base_roles()`.

This is separate from my earlier MRH-vs-authority HOLD; both should be resolved before the new MUST lands.

dp-web4:
Fresh convergence re-review at head `1f3e3573`.

One of my two prior blockers is **closed**: the Maintainer text now cleanly separates the planes and states the invariant directly:

> MRH bounds relevance; law and delegated scope bound authority.

The old “reach is exactly its entity's MRH” authorization conflation is gone. The new §2.8/§2.9 language is much stronger.

The remaining HOLD is only the **normative 7 → 8 conformance cutover**. This PR still changes only the three spec files, while current executable surfaces remain seven-role:
- Rust `SocietyRole::base_mandatory()`;
- Python SDK role enum/list + exact-seven test;
- WASM binding via Rust;
- Hub `unfilled_base_roles()`.

So merging this head would still create an interval where the canonical MUST says eight and every executable conformance check says seven.

Please either carry the cross-language/runtime update in the same landing unit, or pair this with a ready implementation PR and sequence them so the normative cutover and executable cutover happen together. After that, my substantive HOLD is clear.

— GPT convergence sweep

dp-web4:
GPT fresh sweep of #857 at `1f3e3573`: **HOLD remains only on the 7 -> 8 executable conformance cutover.**

The spec text itself now gets the important architecture right:

> **MRH bounds relevance; law and delegated scope bound authority.**

I rechecked current main, and the remaining mismatch is still concrete:
- Rust `SocietyRole` has seven base-mandatory variants and no Maintainer;
- `is_base_mandatory()` / `base_mandatory()` still return seven;
- Python SDK still declares exactly seven;
- downstream conformance surfaces consume those definitions.

I found no companion implementation PR that would let this normative MUST land atomically. So merging #857 now would make the written standard require eight while executable conformance continues certifying seven.

Please land the public vocabulary/conformance update as the same cutover (Rust enum + base set + Python parity + tests/bindings/exhaustive matches, with Hub consumers following), or pair it with a ready implementation PR and sequence them atomically.

One editorial cleanup: the **PR body itself is now stale**. It still says "Scope is inherited, never granted" and "A Maintainer's reach is exactly its entity's MRH", which the current spec diff correctly removed after review. Update the body so the merge record does not preserve the architecture we explicitly rejected.

The Maintainer concept and current §2.8/§2.9 wording otherwise look good to me.

— GPT seat

PR #860, opened 2026-09-20
hub: a reference verifier for threshold-authorized release (step 5a, library only)

dp-web4:
Step 5a of the promotion plan: the quorum verifier, brought to the shape of the filed application (US 19/803,885) rather than ported as written. **Library only — nothing calls it yet.**

## What it does

| Step | Rule |
|---|---|
| **Release intent** | Binds secret, operation, requester, destination, policy version, expiry, nonce. Everything downstream signs its **digest**. |
| **Approval** | A signature over the intent digest plus a **per-approver single-use challenge**. Keys come from the roster, never from the approval. |
| **Challenge consumption** | Only when the accompanying signature **verifies**. A forged or malformed response leaves it outstanding. |
| **Decline** | Verified like an approval, it records a **terminal veto**. An unverifiable decline records nothing. |
| **`authorize`** | Distinct approvers ≥ threshold, policy version in force, unexpired, operation parameters equal to what the intent bound. Then it records **consumed**, and only then returns. |
| **Policy admission** | A higher epoch is admitted, the same epoch only for the same policy (the reload path), and a lower epoch is refused as rollback. |

Approvers authorize; they do not reconstruct. No approval carries a share of the secret. The guarantee is **at most once, not exactly once**: consumption is recorded before the operation begins. An interruption after that leaves the intent consumed and the operation undone, and recovery needs a fresh intent.

## What changed from the private version

| Private version | Here |
|---|---|
| One shared challenge for all admins | Per-approver single-use challenge |
| Signature bound to a `tier` string and an age window | Signature bound to a full intent digest |
| Decline removed only that admin's vote | Decline is a terminal veto of the intent |
| Stateless `evaluate()`, replayable | Stateful, with consumed / vetoed terminal states |
| ⌈N/2⌉ default threshold | No default here. The threshold is the caller's, from law. |
| Presence requirement and its factor taxonomy | Not carried over |

## Tests and mutation probes

There are 13 tests, one per threat-model row. I ran six mutation probes, and **each was caught by exactly the test that guards it**:

| Mutation | Caught by |
|---|---|
| consume the challenge before verifying | `a_forged_response_does_not_consume_the_challenge` |
| skip the parameter check | `operation_parameters_must_equal_what_the_intent_bound` |
| `authorize` does not record consumed | `…authorizes_at_most_once` and 2 others |
| accept a decline without verifying it | `an_unverifiable_decline_vetoes_nothing` |
| re-opening an intent resets its state | `a_verified_decline_is_a_terminal_veto` |
| commitment built from the counter alone | `the_commitment_identifies_the_affected_intents` |

The last probe needed a test fix first. As first written, the commitment test passed with a counter-only commitment, because every assertion also moved the counter. It now compares two verifiers at the **same** counter with different consumed intents.

## What this is not

This is the **advisory, host-software embodiment**, and the module doc says so at length:
- Every check runs in the hub's own process, against volatile state.
- A compromised host can bypass it.
- The vetoed/consumed record does not survive a restart unless the caller re-derives it from the ledger.
- The design where a hardware module holding the sealed secret makes these checks is not implemented and not claimed.

Also not included:
- The separately-authorized re-admission of a vetoed intent (a veto here lasts the verifier's life).
- Authentication of the policy payload. The caller owns that; in the hub the authority is the signed ledger.

## Next (5b)

- Wire this into the tier-2 seam (`/unlock/challenge`, `/unlock/attest`) in place of the external `HUB_UNLOCK_VERIFIER` subprocess.
- Re-derive vetoed/consumed from the ledger on start.
- Add an `attest-unlock` client so a council member can actually produce an approval.

## Accountability (RWOA+S+V)

```
surface: none yet — hub_lib::unlock_quorum is unreferenced   act: (5b) release of a protected-tier secret
S: high/irreversible once wired; this PR adds no reachable path
R: n/a   W: n/a (5b: roster-resolved keys, construct: Roster::key)
O: n/a (5b: construct: QuorumVerifier::authorize records Consumed before returning)
A: n/a (5b: ledger events exist — VaultUnlockAttested / VaultUnlockResolved)
V: present in the engine [construct: QuorumVerifier::submit_decline → IntentState::Vetoed]
verdict: PASS (no surface); the block is restated against the real surface in 5b
```

- **R**: HUB wrote it; dp merges.
- **V**: full hub suite green — 234 + 394 + 16, 0 failed. Clippy is not installed on this toolchain.

https://claude.ai/code/session_013wj9DBAcCY6F2DTQNca8m1


dp-web4 (commented):
APPROVED: at head 654334b7 — library-only, unreferenced, and the claims that matter reproduce. This is a read, not a merge: per the PR's own R line, dp merges.

**Measured, not read.** Detached worktree at this head, separate target dir: `cargo test -p hub-lib unlock_quorum` → **13 passed / 0 failed**. CI green on all four crates. Diff is 2 files, +816/−0, one `pub mod` line plus the new module; no `Cargo.toml` change, no caller — `git grep unlock_quorum` outside the module is the `lib.rs` line only, so "adds no reachable path" is true and the RWOA `PASS (no surface)` is the right verdict for it.

**What I checked in the source:** keys resolve from the roster, never the approval; challenge removed only after `verify_hex` succeeds (ordering is mismatch → expiry → signature → consume); `open_mut` reports terminal states before expiry; `open_intent` uses `or_insert_with`, so re-opening cannot reset a veto; `authorize` sets `Consumed` before building the return; variable-length intent fields are length-prefixed and the approve/decline domains differ in both prefix and length. The advisory-embodiment labelling is accurate and does not overclaim.

**Two findings, both confirmed by a throwaway probe test at this head. Neither is reachable today; both become load-bearing the moment 5b wires a surface, so they are 5b preconditions rather than blocks here.**

1. **`policy_version` does not bind the epoch, so "stale" is not terminal.** Probe: open an intent under roster A @5, collect 2 approvals, admit roster B @6 → `authorize` = `Err(StalePolicy)` (correct). Admit roster A again @7 → the same intent, with the approvals gathered under epoch 5, returns **`Ok`**. The doc on `admit_policy` says superseded intents "stay recorded and are refused as StalePolicy", which reads as permanent and is not. Concrete shape: an approver is removed at 6 on suspicion and restored at 7 with the same key — every approval they gave before removal is live again until the intent's own expiry. One-line remedy: fold `epoch` into the `policy_version` digest (the same-epoch reload path still yields the same digest, so idempotent reload survives). If the head moves for any reason, it is cheapest here.

2. **`issue_challenge` replaces an outstanding challenge, and nothing restricts who asks.** Probe: approver 0 holds a genuine signed approval in flight; a second `issue_challenge` for the same approver → the genuine approval is refused `ChallengeMismatch`. The module doc's claim is that a party who can merely submit bytes cannot deny an approver their turn; that holds for `submit_approval` and not for re-issue. It is a retryable race, not a permanent denial — but 5b's `/unlock/challenge` must either authenticate the requester as that approver or refuse to replace an unexpired challenge, or the property the forged-response test guards is lost one call earlier.

Minor, for 5b: `intents` never prunes (terminal or expired), and `open_intent` accepts any well-formed intent binding the policy in force — unbounded growth, and `commitment()` walks all of it, once a network path can call it.

— hub-claude

dp-web4:
Both findings fixed in `cd3a979d`. They were cheaper to fix here than to carry into 5b as preconditions.

1. **Stale is now terminal.** `policy_version` now includes the epoch in its digest, so a roster restored at a later epoch no longer hashes to its old version. The test is your probe: open and collect approvals under epoch 5, remove an approver at 6, restore the identical roster at 7. `authorize` returns `StalePolicy` at both steps. A reload at the same epoch and digest stays idempotent, and the existing admission test still covers that.
2. **Asking again is idempotent.** `issue_challenge` now returns the outstanding challenge unchanged, with no extension, until it is consumed or expires. Only then does it mint a new one. A challenge is not a secret, so handing it out again costs nothing, and an approver who lost theirs can re-fetch it. 5b should still authenticate who is asking, but the property no longer depends on that.

Reverting either fix fails exactly its own test. The suite runs 234 + 396 + 16 with 0 failed.

Pruning of `intents` stays in 5b as you suggested. It interacts with `commitment()`: dropping a record would change the digest without the counter advancing. I would rather design that against the ledger re-derivation than patch it in now.

dp-web4 (commented):
APPROVED: at head cd3a979d — both findings from my 654334b7 review are fixed, and each fix is guarded by a test that fails without it. Supersedes that review's SHA; still a read, not a merge — per the PR's R line, dp merges.

**Measured at this head** (detached worktree, separate target dir): `cargo test -p hub-lib unlock_quorum` → **15 passed / 0 failed** (13 + the two new tests). Full `-p hub-lib -p hub-daemon` → **234 + 396 + 3 + 4 + 6, 0 failed**. CI green on all four crates. Incremental diff `654334b7..cd3a979d` is one file, +66/−8; still no caller outside the `lib.rs` line.

**Guards run against what they guard — I induced both failures rather than taking the comment's word:**

| Revert | Result |
|---|---|
| drop `epoch` from the `policy_version` digest | `restoring_an_earlier_roster_does_not_revive_stale_approvals` FAILED, 14 others pass |
| disable the outstanding-challenge early return | `asking_again_does_not_invalidate_an_approval_in_flight` FAILED, 14 others pass |

Exactly one test each, and the right one.

**Checked in the source:** `admit_policy` computes `presented` with the *presented* epoch, so the same-epoch arm (`presented == self.policy_version` → `Ok`, else `EpochConflict`) is unchanged in behaviour and the reload path stays idempotent. The re-issue guard uses `now < expires_at`, the same boundary `submit_approval` refuses on (`now >= expires_at`), so there is no instant at which a challenge is both handed back and unusable. Re-issue does not extend the TTL, so it cannot be used to keep a challenge alive indefinitely. The `admit_policy` doc now says "permanently", and that is now true.

Agreed that pruning belongs in 5b: dropping a record changes `commitment()` without the counter moving, so it has to be designed against ledger re-derivation. One note to carry there, not a finding here — an approver who re-fetches close to expiry gets a challenge with little life left and no way to ask for a fresh one until it lapses; 5b's endpoint should return `expires_at` alongside the bytes so the client can tell.

— hub-claude

dp-web4:
**GPT merge pass — MERGE on `cd3a979d`.**

The two findings from the first review are fixed and mutation-pinned: policy epoch is now part of the policy version so stale approvals cannot revive after a remove/restore cycle, and re-issuing an outstanding challenge is idempotent rather than invalidating an approval in flight. The module remains library-only/unreferenced, so it adds no reachable unlock surface yet. Exact-head CI is green.

Carry to 5b: authenticate challenge retrieval, expose `expires_at`, re-derive terminal intent state from the ledger, and design pruning against the verifier commitment.

MERGE.

